PR #11, opened 2020-12-28
Add support for device id 00

skundrik:
Add a simple support for device id `00:` They are TRVs I have on my fairly old Hometronic system.

I have added a new key `RAD1` into `ZONE_TABLE`. I can provide the packet and hass logs as you requested if you let me know where to upload them.


zxdavb:
This has been addressed in the latest merge into master

PR #24, opened 2021-04-02
Add colorama to requirements.txt

NotBobTheBuilder:
I noticed this was missing from requirements.txt so added the latest version


zxdavb:
From: https://pypi.org/project/colorlog/

>colorama is included as a required dependency and initialised when using colorlog on Windows.

zxdavb:
Sorry - the *library* doesn't use **colorama**, but the CLI does. Please, what is your use-case?

zxdavb:
Please use requirements_cli.txt

PR #49, opened 2022-07-28
Add 22F1 and 22F3 to HVAC FAN device class

tomkooij:
An Orcon RF15 display can send 22F1 and 22F3 commands to HRC Fan unit,
which can relay these commands to a SmartComfort valve.
Hence FAN devices can send 22F1 and 22F3.

### DIFF
--- a/ramses_rf/protocol/ramses.py
+++ b/ramses_rf/protocol/ramses.py
@@ -977,7 +977,7 @@ _DEV_KLASSES_HVAC: dict[str, dict] = {
         Code._313F: {I_: {}, RP: {}},
         Code._31D9: {I_: {}, RP: {}},
         Code._31DA: {I_: {}, RP: {}},
-        # Code._31E0: {I_: {}},
+        Code._31E0: {I_: {}},
     },
     DEV_TYPE.CO2: {
         Code._042F: {I_: {}},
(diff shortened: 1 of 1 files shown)


tomkooij:
@zxdavb I'm not sure this needs tests. If so, please point me in the right direction and I'll add it.

zxdavb:
I am not sure I want to do it this way. I think I'd rather create a new entity that is a FAN/REM hybrid.

So, in **device\hvac.py**:
```python
classHvacVentilatorDual(HvacVentilatorBase, HvacRemoteBase):  # DUA
```

Although, the above is a very simplistic hint of what I would do... Some tidy-up would be required, e.g. you wouldn't want faking.
```python
classHvacRemote(Fakeable, HvacRemoteBase):  # REM
classHvacVentilator(HvacVentilatorBase):  # FAN
```

What we're saying, above is that there's a device that is a FAN, _and_ a REM.
```yaml
known_devices:
  35:12345: {class: DUA}
```

In the upper layer, you can then do things like:
```python
if isintance (device, HvacRemoteBase)
```

This is OK, because I already have plans for extending FAN into HRU, CVE & PIV.

All the above is an architectural issue - so, we don't want to get it wrong.

tomkooij:
@zxdavb  It is certainly much more robust that way. It will probably be easier to identify FAN devices if we assume they never send fan speed commands.

I'll try to implement when time permits.

zxdavb:
Closing - please re-open if still needing to be progressed.

PR #64, opened 2022-11-20
Convert signed integers

tomkooij:
Today the outside temperature went below zero:
`2022-11-20T08:32:06.904058 063 RP --- 32:134446 37:171685 --:------ 31DA 030 00EF007FFF2F1B0226069A07EEFFE4F8000038988F0000EFEF1F2420FC00
 `

Temperature `FEE4` should be -2.8 degrees.

two's complement signed integers to negative values

EDIT: DO NOT MERGE. There is a bug (factor 10) this results in -0.28 degrees. Obviously this needs a unit test.

### DIFF
--- a/ramses_rf/protocol/helpers.py
+++ b/ramses_rf/protocol/helpers.py
@@ -104,13 +104,20 @@ def date_from_hex(value: str) -> Optional[str]:  # YY-MM-DD
 
 
 @typechecked
-def double_from_hex(value: str, factor: int = 1) -> Optional[float]:
+def double_from_hex(value: str, factor: int = 1, neg_factor: int = None) -> Optional[float]:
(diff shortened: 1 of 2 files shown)


ramses_rf/protocol/helpers.py
```diff
@@ -104,13 +104,20 @@ def date_from_hex(value: str) -> Optional[str]:  # YY-MM-DD
 
 
 @typechecked
-def double_from_hex(value: str, factor: int = 1) -> Optional[float]:
+def double_from_hex(value: str, factor: int = 1, neg_factor: int = None) -> Optional[float]:
```
tomkooij, line 107: This probably breaks type checking

tomkooij:
Disregard this.

PR #65, opened 2022-11-20
bugfix: 31DA not using two's complement for negative temperatures

tomkooij:
```
2022-11-20T08:32:06.904058 063 RP --- 32:134446 37:171685 --:------ 31DA 030 00EF007FFF2F1B0226069A07EEFFE4F8000038988F0000EFEF1F2420FC00
````

The above payload parses like so: `'outdoor_temperature': 655.08`. However, temperatures such as `FEE4` should be -0.28 °C.

The solution is to use `temp_from_hex('FEE4')` instead of `double_from_hex(...)`.

The Fan is an Orcon HRC 425

[edited by dev]

### DIFF
--- a/ramses_rf/protocol/parsers.py
+++ b/ramses_rf/protocol/parsers.py
@@ -2171,10 +2171,10 @@ def parser_31da(payload, msg) -> dict:
         SZ_CO2_LEVEL: double_from_hex(payload[6:10]),  # ppm, 1298[2:6]
         SZ_INDOOR_HUMIDITY: percent_from_hex(payload[10:12], high_res=False),  # 12A0?
         SZ_OUTDOOR_HUMIDITY: percent_from_hex(payload[12:14], high_res=False),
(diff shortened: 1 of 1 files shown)


zxdavb:
@tomkooij thanks for your PR - looks like an easy win.

For posterity, please indicate the make/model of FAN: Orcon HRU HRC400?

The reason we need to know make/model because some of the fields in `31DA` vary by make (manufacturer), although the temperature fields are not likely to be so affected.

I will add this to the test suite, too.

tomkooij:
> @tomkooij thanks for your PR - looks like an easy win.
> 
> For posterity, please indicate the make/model of FAN.
> 
> The reason we need to know make/model because some of the fields in `31DA` vary by make (manufacturer), although the temperature fields are not likely to be so affected.

iI added Orcon HRC 425

zxdavb:
And we're done. Thanks.

PR #89, opened 2023-12-01
Ignore battery levels of 0% as they are impossible

### LINKED ISSUE #88: DT4R always reports battery level as 0
I have a couple of Honeywell Home DT4R wireless thermostats that appear to always report their battery level as 0:
```
2023-12-01T12:35:26.260222 043  I --- 22:014503 --:------ 22:014503 1060 003 000001
2023-12-01T12:46:31.899891 048  I --- 22:014583 --:------ 22:014583 1060 003 000001
```

As this can never be the case, would it be possible to add a condition for this in `parser_1060` and return the level as None? I presume this would then map through to Home Assistant via ramses_cc as a battery level attribute of `Unknown` which is preferable to the erroneous data.

trvrnrth:
Resolves #88


zxdavb (changes requested):
This sort of behaviour should be encouraged - thanks!

However, adding the frames (as comments) to **parsers.py** is no longer best practice.

Instead, add the frames to **tests/parsers/code_1060.log**. This file doesn't exist, but should - have a look at **tests/parsers/code_10d0.log**, and you'll see exactly what is required.

Add it to this this commit.

zxdavb (approved):
That will do me (I'll make some minor chanes)

PR #109, opened 2024-03-27
Refactor protocol/transport

zxdavb:
Move elements of protocol and transport to more appropriate layers, for example:
 - duty cycle / min write gap moved down to transport layer
 - device id filtering moved up to protocol layer
 - Qos is now enabled by default (**QosProtocol** and **PortProtocol** merged)

Related changes:
 - tweak timers for above
 - refactor protocol FSM to be clearer / more robust
 - can now compare packets (as Rx'd) with commands (as Tx'd)

Add / refactor / bugfix tests for above.


zxdavb (commented):
Reviewed.

PR #137, opened 2024-10-14
Tweak `1298` for Vasco HVAC

silverailscolo:
found a few packets


tests/tests/parsers/code_1298.log
```diff
@@ -0,0 +1,3 @@
+# a Vasco CO2 remote sending a measurement:
+2024-10-14T20:14:21.028276 064  I --- 37:117647 --:------ 37:117647 1298 003 000322  # {'co2_level': 802}
+2024-10-10T22:59:41.307778 074  I --- 37:117647 --:------ 37:117647 1298 003 0001FF
```
zxdavb: Thanks for adding **code_1298.log** - a useful addition.

For the `I|1298|0001FF` - add: `# {'co2_level': 511}`

zxdavb (changes requested):

tests/tests/parsers/code_1298.log
silverailscolo: Also tried to RQ the CO2 using ramses_cc action. Works, added to log. 1298 done so far

src/ramses_tx/ramses.py
```diff
@@ -273,6 +273,8 @@
     Code._1298: {  # co2_level
         SZ_NAME: "co2_level",
         I_: r"^00[0-9A-F]{4}$",
+        RQ: r"^00$",  # tried in ramses CLI, got reply:
+        RP: r"^00[0-9A-F]{4}$",
```
zxdavb: When the `RP` payload REGEX is the same as the `I`, then you don't add the `RP`, but you do need to have the `RQ`.

That is what this code (later in **ramses.py** does:
```python
for code in CODES_SCHEMA.values():  # map any RPs to (missing) I_s
    if RQ in code and RP not in code and I_ in code:
        code[RP] = code[I_]
```

So, please remove the `RP: r"^00[0-9A-F]{4}$",`

tests/tests/parsers/code_1298.log
```diff
@@ -0,0 +1,7 @@
+# a Vasco CO2 remote sending a measurement:
+2024-10-14T20:14:21.028276 064  I --- 37:117647 --:------ 37:117647 1298 003 000322  # {'co2_level': 802}
+2024-10-10T22:59:41.307778 074  I --- 37:117647 --:------ 37:117647 1298 003 0001FF  # {'co2_level': 511}
+
+# requesting the current value
```
zxdavb: I suggest:
```
# Vasco HRU will RP to a contrived RQ
````

zxdavb (changes requested):

zxdavb:
Make these two changes, and we'll merge

zxdavb (approved):

zxdavb:
Thanks @silverailscolo

PR #140, opened 2024-10-14
Tweak `10D0` parser for Vasco HVAC

silverailscolo:
No new logs for `10D0`. Checks passed


zxdavb:
Do you have any `10D0` packets in a packet log to support your PR? I would expect to see them.

If so, they need to be added to **tests/tests/parsers/code_10d0.log**? They will, if they are different to the existing `W|10D0|00FF` packet:
```
2022-07-03T22:52:34.571579 045  W --- 37:171871 32:155617 --:------ 10D0 002 00FF          # {'reset_counter': True}
```

silverailscolo:
> They will, if they are different to the existing W|10D0|00FF packet:

They are identical, so I didn't add a copy

silverailscolo:
I received a new packet Filter Change Reply from the D60 after placing the ramses_esp nearby:
```
2024-10-15 12:33:55.595000 000 RQ --- 18:130140 32:022222 --:------ 10D0 001 00            # {}
2024-10-15 12:33:55.609000 --- RP --- 32:022222 18:130140 --:------ 10D0 006 00A9B6B90000  # {'days_remaining': 185}
```

Added to the 10D0 tests file.

src/ramses_tx/parsers.py
```diff
@@ -990,9 +990,19 @@ def parser_10d0(payload: str, msg: Message) -> dict[str, Any]:
...
+        if (
+            payload[2:4] == payload[4:6]
+        ):  # seen on ClimaRad VenturaV1x 2021, calculate days_remaining
+            result.update(
+                {
+                    "days_remaining": int(
+                        int(payload[2:4], 16) * (hex_to_percent(payload[6:8]) or 0)
+                    )
+                }
+            )
```
zxdavb: My experience tells me that `FE`, like `FF`, is a sentinel value.

You will need to look at tomorrow's payload, to confirm that it changes from: `00FEFE79FFFF` to `00FDFE79FFFF`.

Otherwise, the dict should be (similar to):
```
{
  'days_remaining': None, 
  'days_lifetime': None, 
  'percent_remaining': 0.605
}
```

src/ramses_tx/parsers.py
```diff
@@ -990,9 +990,19 @@ def parser_10d0(payload: str, msg: Message) -> dict[str, Any]:
...
     return result
```
zxdavb, line 999:
```suggestion
    result: dict[str, bool | float | None]

    if msg.verb == W_:
        return {"reset_counter": payload[2:4] != "00"}

    result = {}

    if payload[2:4] not in ("FF", "FE"):
        result["days_remaining"] = int(payload[2:4], 16)

    if payload[4:6] not in ("FF", "FE"):
        result["days_lifetime"] = int(payload[4:6], 16)

    result["percent_remaining"] = hex_to_percent(payload[6:8])

    return result
```

zxdavb: The above suggestion is for the entire block,

zxdavb (changes requested):

src/ramses_tx/parsers.py
zxdavb: I think I wasn't clear - that code block was for the entire function.

tests/tests/parsers/code_10d0.log
```diff
@@ -3,6 +3,12 @@
 2022-07-03T22:52:34.571579 045  W --- 37:171871 32:155617 --:------ 10D0 002 00FF          # {'reset_counter': True}
 2022-07-03T22:52:34.596526 066  I --- 32:155617 37:171871 --:------ 10D0 006 0047B44F0000  # {'days_remaining':  71, 'days_lifetime': 180, 'percent_remaining': 0.395}
 
+2024-10-15 12:33:55.595000 000 RQ --- 18:130140 32:022222 --:------ 10D0 001 00            # {}
+2024-10-15 12:33:55.609000 --- RP --- 32:022222 18:130140 --:------ 10D0 006 00A9B6B90000  # {'days_remaining': 169, 'days_lifetime': 182, 'percent_remaining': 0.925}
+
+2024-10-15T12:33:13.714200 000 RQ --- 18:130140 37:153226 --:------ 10D0 001 00            # {}
+2024-10-15T12:33:13.731547 060 RP --- 37:153226 18:130140 --:------ 10D0 006 00FEFE79FFFF  # {'days_remaining': 153, 'days_lifetime': 254, 'percent_remaining': 0.605}
```
zxdavb: Will become: `# {'percent_remaining': 0.605}`

zxdavb (changes requested):

src/ramses_tx/parsers.py
silverailscolo: Seems correct. Today, the payload is exactly the same `00FEFE79FFFF` so no days. Code suggestion adopted.

zxdavb (approved):

zxdavb:
Thanks @silverailscolo

PR #141, opened 2024-10-16
Tweak `3150` for ClimaRad Ventura V1X

silverailscolo:

PR #143, opened 2024-10-17
Tweak `3150` for ClimaRad HVAC

silverailscolo:
just 1 (static?) packet


silverailscolo:
Duplicate, already merged as #141

PR #146, opened 2024-10-18
Tweak `3222` for Vasco D60

silverailscolo:
parser not a percentage, looks like a sort of mapping


zxdavb:
Good work, but I am going to completely refactor this on your behalf:
 - I am going to omit the typed dict for nwo
 - you can probably drop this PR (but do submit another if you feel you've more to add)

PR #147, opened 2024-10-18
Tweak `22F4` for ClimaRad

silverailscolo:
Compare to `22F2` modes. These new modes could also be applied to the existing packets.


zxdavb:
This has opened a can of worms - I think easier if I do it, based upon your packet logs.

Please have a look at the `22F4` parser in the latest master, and submit another PR based upon that.

PR #151, opened 2024-10-21
Tweak `2210`

silverailscolo:
Still not clear what the 3 values in `2210` mean


silverailscolo:
@zxdavb can you review this?

zxdavb:
@silverailscolo I am sorry I have not reviewed your PRs - I am under the hammer with work at the moment, and another repo is consuming all my spare capacity  - please bear with me.

silverailscolo:
I support your prios. We'll get to it

src/ramses_tx/parsers.py
```diff
@@ -1363,15 +1363,11 @@ def parser_1fd4(payload: str, msg: Message) -> PayDictT._1FD4:
...
-    assert payload in (
-        "00FF" + "00FFFFFF0000000000FFFFFFFFFF" * 2 + "FFFFFF000000000000000800",
-    ), _INFORM_DEV_MSG
-
-    return {}
+    return {
+        "unknown_78": payload[78:80],
+        "unknown_80": payload[80:82],
+        "unknown_82": payload[82:],
+    }
```
zxdavb: don't cut out data if you don't know what it means

zxdavb (changes requested):

src/ramses_tx/parsers.py
zxdavb: other people need to see this data, so they can contribute to the corpus

silverailscolo: The parser in current head is returning nothing at all, so I think we're not cutting out anything, and the assert didn't fit, so I took that out as advised earlier

zxdavb (approved):

PR #134, opened 2024-10-13
Tweak `10E0` for Vasco/ClimaRad HVAC

silverailscolo:
This PR adds signature information for the following hardware:

## Vasco
- `32:022222` HRU D60 (VMD-17RPS01) legacy product, no docs on Vasco site
- `37:117647` CO2 Sensor / Remote (VMS-17C01) [docs](https://vasco.eu/files/downloads/technical-sheet-rf-co2-switch/90.01.06.80-TECHNISCHE-FICHE-CO2-RF-BEDIENING-ENG.pdf)
- `29:091138` REM 4-button RF Remote (VMN-17LMP01) [docs](https://vasco.eu/files/downloads/technical-sheet-rf-3-position-switch/90.01.06.76-TECHNISCHE-FICHE-RF-SCHAKELAAR-ENG.pdf)

[Vasco D60 brochure NL.pdf](https://github.com/user-attachments/files/17402182/Vasco.D60.brochure.NL.pdf)

## ClimaRad
- `29:099029` FAN MiniBox (VMC-07RP01) [docs](https://www.climarad.nl/en/products/climarad-minibox/) 
- `37:153226` HRU Ventura V1x (VMD-07RPS13) [docs](https://www.climarad.nl/en/products/climarad-ventura-v1x/)
- `29:123150` REM 4-button RF Remote (VMN-07LM01) [docs](https://www.climarad.nl/en/products/climarad-wireless-control-panel-rf/)

PR #152, opened 2024-10-21
Tweak `31D9` for ClimaRad HVAC

silverailscolo:
Uses `31DA` parsing already in place for Itho (only entity `fan_speed` is useful/varies)


src/ramses_tx/parsers.py
```diff
@@ -2115,7 +2115,7 @@ def parser_31d9(payload: str, msg: Message) -> dict[str, Any]:
 
     bitmap = int(payload[2:4], 16)
 
-    # NOTE: 31D9[4:6] is fan_rate (itho?) *or* fan_mode (orcon?)
+    # NOTE: 31D9[4:6] is fan_rate (minibox!, itho?) *or* fan_mode (orcon?)
     result = {
         **parse_exhaust_fan_speed(payload[4:6]),  # itho
```
silverailscolo, line 2144: Is it OK if we use `_22F1_MODE_ITHO` from parsers to map these values to something more meaningful?

zxdavb: > Is it OK if we use _22F1_MODE_ITHO from parsers to map these values to something more meaningful?

I am not exactly sure what you're asking - please provide more information.

src/ramses_tx/parsers.py
```diff
@@ -2128,6 +2128,18 @@ def parser_31d9(payload: str, msg: Message) -> dict[str, Any]:
     }
 
     if msg.len == 3:  # usu: I -->20: (no seq#)
+        if msg.src.type == "32" and payload[:4] == "0000":  # Vasco D60 HRU
+            # can't access 'vasco' scheme as this PR is separate from _22f1
```
zxdavb: ??? See above

zxdavb (changes requested):

src/ramses_tx/parsers.py
silverailscolo: At the moment, the `parse_exhaust_fan_speed` helper returns simply the value it was given. Using a lookup on the mode-list we could return a state name instead.

src/ramses_tx/parsers.py
silverailscolo: This is not directly connected to the above question. Once the 22F1 is merged I can point this parser to the `vasco` scheme.

src/ramses_tx/parsers.py
silverailscolo: We have that scheme/lookup list here in ramses_rf, but for i18n/localization it might be best to wait for ramses_cc to assign the strings used in the UI

sembeek:
I don't know if it helps you, but i have picked up values for this code that are not yet in this pull request.
If it does not help you anything at all, please let me know how i can help if you want.

```
I --- 29:098549 --:------ 29:098549 31D9 003 008001 < PacketInvalid( I --- 29:098549 --:------ 29:098549 31D9 003 008001 < Unexpected code for src (HUM) to Tx)
I --- 29:098549 37:255488 --:------ 31D9 003 008001 < PacketInvalid( I --- 29:098549 37:255488 --:------ 31D9 003 008001 < Unexpected code for src (HUM) to Tx)
I --- 29:098458 --:------ 29:098458 31D9 003 000001 < PacketInvalid( I --- 29:098458 --:------ 29:098458 31D9 003 000001 < Unexpected code for src (HUM) to Tx)
```

silverailscolo:
> I don't know if it helps you, but i have picked up values for this code that are not yet in this pull request.
> 
> If it does not help you anything at all, please let me know how i can help if you want.
> 
> 
> 
> ```
> 
> I --- 29:098549 --:------ 29:098549 31D9 003 008001 < PacketInvalid( I --- 29:098549 --:------ 29:098549 31D9 003 008001 < Unexpected code for src (HUM) to Tx)
> 
> I --- 29:098549 37:255488 --:------ 31D9 003 008001 < PacketInvalid( I --- 29:098549 37:255488 --:------ 31D9 003 008001 < Unexpected code for src (HUM) to Tx)
> 
> I --- 29:098458 --:------ 29:098458 31D9 003 000001 < PacketInvalid( I --- 29:098458 --:------ 29:098458 31D9 003 000001 < Unexpected code for src (HUM) to Tx)
> 
> ```

Thanks, added them in this PR. If I understand your message, this 29:x device is a HUM/humidity sensor? And it reports to a 37:x a fan?
Could you provide the device info messages for both? What brand/manufacturer?
This info could be added to `10E0`, see [here](https://github.com/zxdavb/ramses_rf/pull/134)

src/ramses_tx/parsers.py
```diff
@@ -2129,6 +2129,24 @@ def parser_31d9(payload: str, msg: Message) -> dict[str, Any]:
     }
 
     if msg.len == 3:  # usu: I -->20: (no seq#)
+        if (
+            payload[:4] == "0000"
+            and msg._addrs[0] == msg._addrs[2]
+            and msg._addrs[1] == NON_DEV_ADDR
+        ):
+            # Vasco D60 HRU and ClimaRad minibox REM
+            from .ramses import _31D9_FAN_INFO
```
zxdavb: This is unusual - please justify not doing the import at the top of teh module - add a comment, like: `# avoids circular import` if so

src/ramses_tx/ramses.py
```diff
@@ -1339,6 +1339,19 @@
     },
 }
 
+# ventilation speed description
+_31D9_FAN_INFO: dict[int, str] = {
```
zxdavb: is this for all 31d9, or only for some vendors?

zxdavb (changes requested):

src/ramses_tx/parsers.py
silverailscolo: Copied this pattern from parser_22f1, where it is used 3 times + a rename. We do not need a rename here, so have moved import to the top

src/ramses_tx/ramses.py
silverailscolo: As far as I can tell, is only for Vasco/Climarad fans reporting speed via 31d9 as it was set by (faked) 4-button remotes. Note that "auto" mode is also in this dict, so can't use existing `_31DA_FAN_INFO` dict

silverailscolo: Renamed item to _31D9_FAN_INFO_VASCO to make this evident

src/ramses_tx/parsers.py
silverailscolo: comment line 2132 has been removed since 22F1 was updated in the meantime

zxdavb (approved):

PR #155, opened 2024-10-23
Tweak `3200` for ClimaRad

silverailscolo:


src/ramses_tx/ramses.py
```diff
@@ -568,6 +568,7 @@
     },
     Code._3200: {  # boiler output temp
         SZ_NAME: "boiler_output",
+        I_: r"^00[0-9A-F]{4}$",
         RQ: r"^00$",
         RP: r"^00[0-9A-F]{4}$",
```
zxdavb, line 573: We can take out the `RP`, given we add the ` I` (and they're the same payload).

(I will do that)

zxdavb (approved):

PR #156, opened 2024-11-03
Tweak `22F4` ClimaRad payload

silverailscolo:
Merge after #169, which will fix linting and testing.


src/ramses_tx/parsers.py
```diff
@@ -1637,24 +1637,30 @@ def parser_22f3(payload: str, msg: Message) -> dict[str, Any]:
...
-        0x01: "speed 1",  # "low", or trickle?
-        0x02: "speed 2",  # "medium-low", or low?
-        0x03: "speed 3",  # "medium",
-        0x04: "speed 4",  # "medium-high", or high?
+    RATE_LOOKUP = {  # note no i18n, passed to UI as is
+        0x00: "0",  # "off"?,
+        0x01: "1",  # "low", or trickle?
+        0x02: "2",  # "medium-low", or low?
+        0x03: "3",  # "medium",
+        0x04: "4",  # "medium-high", or high?
```
zxdavb: i suggest keep the old values - please explain why you've chosen not to do that

zxdavb (changes requested):

src/ramses_tx/parsers.py
silverailscolo: Passing along the word 'speed' in a non-English UI is not advised from i18n (internationalization). Can be added in UI by user if desired, harder to take it out. The label should make clear what a value means, so perhaps one would want to replace 'fan_info' in the UI with 'Speed'.

silverailscolo: Rolled back the existing strings, as requested.

src/ramses_tx/parsers.py
```diff
@@ -1660,19 +1658,25 @@ def parser_22f3(payload: str, msg: Message) -> dict[str, Any]:
 
 # WIP: unknown, HVAC
 def parser_22f4(payload: str, msg: Message) -> dict[str, Any]:
-    # HACK: for dev/test: 37:153226 is ClimaRad Ventura fan/remote
-    payload = payload[8:14] if msg.src.id == "37:153226" else payload[:6]
+    # ClimaRad Ventura fan/remote
+    if msg.len == 13 and payload[14:] == "000000000000":
+        if payload[10:12] == "00":
+            payload = payload[:4] + payload[12:14]
```
zxdavb: The payload should be immutable, no?

src/ramses_tx/parsers.py
```diff
@@ -1660,19 +1658,25 @@ def parser_22f3(payload: str, msg: Message) -> dict[str, Any]:
...
+        0x00: "off",
         0x20: "paused",
         0x40: "auto",
         0x60: "manual",
     }
     mode = int(payload[2:4], 16) & 0x60
     assert mode in MODE_LOOKUP, mode
 
-    RATE_LOOKUP = {
+    RATE_LOOKUP = {  # note: no i18n here, localize in application
```
zxdavb: no comment required

zxdavb (dismissed):

src/ramses_tx/parsers.py
silverailscolo: Now using an extra var `_pl`

src/ramses_tx/parsers.py
silverailscolo: Removed this comment

PR #136, opened 2024-10-14
Tweak `22F1` for Vasco REM/CO2

silverailscolo:
For Vasco REM/CO2 sensor.
Split off from PR #133.

Merge after #169, which will fix linting and testing.

PR #171, opened 2025-03-18
Tweak `22F1`

silverailscolo:
Add Vasco REM/CO2

Replaces PR #136, same contents as reviewed (I made an error pulling to the original branch)


src/ramses_tx/parsers.py
```diff
@@ -1531,6 +1531,8 @@ def parser_22e9(payload: str, msg: Message) -> Mapping[str, float | str | None]:
 
 # fan_speed (switch_mode), HVAC
 def parser_22f1(payload: str, msg: Message) -> dict[str, Any]:
+    # ClimaRad VenturaV1x HRU does not send 22F1 for speed, uses 22F4 for mode + step
+
```
zxdavb, line 1535: Going forward - put these comments in the test log files

Otherwise, teh codebase is going to bloat

zxdavb (approved):

src/ramses_tx/parsers.py
silverailscolo: Done as part of PR #156

PR #174, opened 2025-03-21
Tweak `22F2`

silverailscolo:
add DevType.FAN code 22F2


silverailscolo:
`mypy` passes locally

zxdavb (changes requested):
needs to pass all test, then we can merge - same for other PRs.

zxdavb (approved):

PR #176, opened 2025-04-10
hvac indoor temp

silverailscolo:
Fetch device/hvac indoor temp from `12A0` (ClimaRad Ventura only) and from `31DA`.


src/ramses_rf/device/hvac.py
```diff
@@ -383,6 +383,19 @@ def indoor_humidity(self) -> float | None:
...
+            ):  # FAN Ventura sends RH/temps as a list, use element [0] for indoor_temp
+                for k, v in self._msgs[Code._12A0].payload[0].items():
+                    if k == SZ_TEMPERATURE:
+                        return float(v)
+            for k, v in self._msgs[Code._12A0].payload.items():
+                if (
+                    k == SZ_TEMPERATURE
+                ):  # ClimaRad minibox FAN sends (indoor) temp in 12A0
+                    return float(v)
+
```
zxdavb: This doesn't look right - at least use a guard clause:
```python
if Code._12A0 not in self._msgs:
    return None
```

zxdavb: This part of the code is a mess... 

In future, you can look at switching over to `MessageIndex` (see **ramses_rf/database.py**).

zxdavb (changes requested):

src/ramses_rf/device/hvac.py
silverailscolo: added a guard clause 2 times

PR #177, opened 2025-04-10
hvac indoor humidity

silverailscolo:
Fetch device/hvac indoor humidity from 12A0 (Ventura only) and from 31DA


src/ramses_rf/device/hvac.py
```diff
@@ -379,6 +379,21 @@ def fan_info(self) -> str | None:
 
     @property
     def indoor_humidity(self) -> float | None:
+        """
+        Extract humidity value from _12A0 or _31DA JSON message payload
+
+        :return: percentage <= 1.0
+        """
+        if Code._12A0 in self._msgs:
```
zxdavb: The way I did this was to look at both `12A0` and `31DA`, and return the value in the most recent of the two - you'll see examples of this elsewhere.

zxdavb (changes requested):

src/ramses_rf/device/hvac.py
silverailscolo: added a guard clause, see comment in PR #178 for motivation why I don't think I can use the example

PR #178, opened 2025-04-10
hvac exhaust temp

silverailscolo:
Fetch device/hvac exhaust temp from `12A0` (ClimaRad Ventura only) and from `31DA`


src/ramses_rf/device/hvac.py
```diff
@@ -371,6 +371,13 @@ def exhaust_flow(self) -> float | None:
 
     @property
     def exhaust_temp(self) -> float | None:
+        if Code._12A0 in self._msgs:
```
zxdavb: The way I did this was to look at both `12A0` and `31DA`, and return the value in the most recent of the two - you'll see examples of this elsewhere.

zxdavb (changes requested):

src/ramses_rf/device/hvac.py
silverailscolo: For example line 438 uses the same key for both codes. That is different here, where the key in 12A0 is just 'TEMPERATURE'. I chose to leave the 31DA return as is (and couldn't refactor/understand to use the shorthand style used in line 438 to my case).

src/ramses_rf/device/hvac.py
```diff
@@ -371,6 +371,13 @@ def exhaust_flow(self) -> float | None:
...
+            if isinstance(
+                self._msgs[Code._12A0].payload, list
+            ):  # FAN Ventura sends RH/temps as a list, use element [2] for exhaust temp
+                for k, v in self._msgs[Code._12A0].payload[2].items():
+                    if k == SZ_TEMPERATURE:
+                        return float(v)
```
zxdavb: there are helper functions you should be using instead of `float()`.

For example, there are some magic values for temperature.

zxdavb (changes requested):

src/ramses_rf/device/hvac.py
silverailscolo: The HEX value was already parsed, including the magic. Here, float() just casts the return value.

silverailscolo: Now replaced by an assert, including a possible None

PR #179, opened 2025-04-10
hvac filter-change docs

silverailscolo:
Text edit to documentation

### DIFF
--- a/src/ramses_rf/device/hvac.py
+++ b/src/ramses_rf/device/hvac.py
@@ -292,7 +292,7 @@ class HvacDisplayRemote(HvacRemote):  # DIS
 class HvacVentilator(FilterChange):  # FAN: RP/31DA, I/31D[9A]
     """The FAN (ventilation) class.
 
-    The cardinal codes are 31D9, 31DA, 12A0.  Signature is RP/31DA.
+    The cardinal codes are 31D9, 31DA.  Signature is RP/31DA.
     """
 
     # Itho Daalderop (NL)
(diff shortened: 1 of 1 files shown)


src/ramses_rf/device/hvac.py
```diff
@@ -289,16 +289,18 @@ class HvacDisplayRemote(HvacRemote):  # DIS
     #     )
 
 
-class HvacVentilator(FilterChange):  # FAN: RP/31DA, I/31D[9A]
+class HvacVentilator(FilterChange):  # FAN: RP/31DA, I/31D[9A], I/12A0
```
zxdavb: No.

Anything that sends `RP/31DA, I/31D[9A]` is - by definition - a FAN.

Devices that send `I/12A0` may not be FANs.

zxdavb (changes requested):

src/ramses_rf/device/hvac.py
```diff
@@ -292,13 +292,15 @@ class HvacDisplayRemote(HvacRemote):  # DIS
 class HvacVentilator(FilterChange):  # FAN: RP/31DA, I/31D[9A]
     """The FAN (ventilation) class.
 
-    The cardinal code are 31D9, 31DA.  Signature is RP/31DA.
+    The cardinal codes are 31D9, 31DA, 12A0.  Signature is RP/31DA.
```
zxdavb: 12A0 is not a cardinal code of **HvacVentilator** (although it may be of **FilterChange**).

The definition of cardinal is something like: if that device issues those codes, then it much be a device of this class - TBH, a risky premise, in any case.

zxdavb (changes requested):

zxdavb (approved):

PR #180, opened 2025-04-10
hvac filter-remaining

silverailscolo:
Add hvac filter-remaining-percentage
This sensor is infrequently updated so anyway useful to store last seen value in a HA measurement.


silverailscolo:
@zxdavb All my small hvac @param PR's are ready for review, all review requests addressed

PR #182, opened 2025-04-10
hvac exhaust fan speed

silverailscolo:
Fetch hvac exhaust fan speed from 31D9 for ventura, others use key in 31DA


src/ramses_rf/device/hvac.py
```diff
@@ -362,8 +362,15 @@ def co2_level(self) -> int | None:
...
-    def exhaust_fan_speed(self) -> float | None:  # was from: (Code._31D9, Code._31DA)
-        return self._msg_value(Code._31DA, key=SZ_EXHAUST_FAN_SPEED)
+    def exhaust_fan_speed(
+        self,
+    ) -> float | None:  # some fans use Code._31D9 for speed + mode
+        for c in (Code._31DA, Code._31D9):
+            if c in self._msgs:
+                for k, v in self._msgs[c].payload.items():
+                    if k == SZ_EXHAUST_FAN_SPEED:
+                        return float(v)  # pick either code
```
zxdavb:
```
if v := self._msgs[c].payload.get(SZ_EXHAUST_FAN_SPEED):
    return f(v)
```

src/ramses_rf/device/hvac.py
```diff
@@ -362,8 +362,15 @@ def co2_level(self) -> int | None:
         return self._msg_value(Code._31DA, key=SZ_CO2_LEVEL)
 
     @property
-    def exhaust_fan_speed(self) -> float | None:  # was from: (Code._31D9, Code._31DA)
-        return self._msg_value(Code._31DA, key=SZ_EXHAUST_FAN_SPEED)
+    def exhaust_fan_speed(
+        self,
+    ) -> float | None:  # some fans use Code._31D9 for speed + mode
+        for c in (Code._31DA, Code._31D9):
```
zxdavb, line 372: The idea is this:
 - if both packets exist, and both have the key you want - use the most recent of the two (one could be 5 minutes ago, the other 1 hour ago)
 - if only one packet exists, use that
 - if no packets exist - return None

zxdavb (changes requested):

silverailscolo:
Error after removing float(): 

```
src/ramses_rf/device/hvac.py:374: error: Returning Any from function declared to return "float | None"  [no-any-return]
```

but value can be None.
Added an assert to safely check for either

PR #196, opened 2025-06-13
SQLite MessageIndex

silverailscolo:
Migrating from storing and retrieving messages using as index a set of nested Dicts to a SQLite central messages index (See our "[Untangling the Organic Knot](https://github.com/ramses-rf/ramses_rf/discussions/191)" discussion).

Stores ~the _parsed_ message payload as JSON in column `pl`, plus~ a summary of all enclosed keys in column `plk` for checking before loading the payload itself. Eventually replaces all uses of `_msgz_`.

TODO:

- [x] @zxdavb Confirm this is the intended approach
- [x] Test fetching payload FAN_INFO values for my Vasco
- [x] Add a test for this process (current tests keep passing, so... they don't fetch?)
- [x] Use this table instead of the ~`_msgs_` and~ `_msgsz_` Dict
- [ ] Fix any issues
- [x] Remove unused `_msgz_` Dict from `ramses_rf/entity_base.py` and `ramses_rf/database.py`


src/ramses_rf/const.py
```diff
@@ -33,6 +33,7 @@
     SZ_EXHAUST_TEMP as SZ_EXHAUST_TEMP,
     SZ_FAN_INFO as SZ_FAN_INFO,
     SZ_FAN_MODE as SZ_FAN_MODE,
+    SZ_FAN_RATE as SZ_FAN_RATE,
```
zxdavb, line 38: What does this have to do with this PR? Please submit atomic PRs - as small as possible.

Doing so will help greatly to get them merged.

Please see: https://developers.home-assistant.io/docs/review-process/#creating-the-perfect-pr

src/ramses_rf/database.py
```diff
@@ -76,23 +118,21 @@ def msgs(self) -> MsgDdT:
...
+
+        Fields:
+
+        - dtm  message timestamp
+        - verb _I, RQ etc.
+        - src  message origin address
+        - dst  message destination address
+        - code packet code aka command class e.g. _0005, _31DA
+        - ctx  message context, created from payload as index + extra markers (Heat)
+        - hdr  packet header e.g. 000C|RP|01:223036|0208 (see: src/ramses_tx/frame.py)
```
zxdavb, line 144: The packets `dtm` is its unique key. I am not sure if we need/want `ctx` - I will need a think. Maybe a query will be enough.

src/ramses_rf/database.py
```diff
@@ -76,23 +118,21 @@ def msgs(self) -> MsgDdT:
...
+        - pl   the parsed message payload, stored as JSON string
+        - plk the keys stored in the parsed payload, separated by the | char
```
zxdavb: I was not thinking to store anything in the SQL DB, other than keys. I think you have to remove this, unless you can justify otherwise - what are you trying to achieve?

The packet objects can/should be stored in a Python dict - the key is the dtm (no two packets arrive at the same time).

src/ramses_rf/database.py
```diff
@@ -114,20 +156,22 @@ def _setup_db_schema(self) -> None:
...
+        # no index on pl
+        self._cu.execute("CREATE INDEX idx_plk ON messages (plk)")
 
         self._cx.commit()
 
     async def _housekeeping_loop(self) -> None:
         """Periodically remove stale messages from the index."""
 
-        def housekeeping(dt_now: dt, _cutoff: td = td(days=1)) -> None:
+        async def housekeeping(dt_now: dt, _cutoff: td = td(days=1)) -> None:
```
zxdavb, line 176: Really needs a docstr to describe it.

src/ramses_rf/database.py
```diff
@@ -114,20 +156,22 @@ def _setup_db_schema(self) -> None:
         self._cu.execute("CREATE INDEX idx_code ON messages (code)")
         self._cu.execute("CREATE INDEX idx_ctx ON messages (ctx)")
         self._cu.execute("CREATE INDEX idx_hdr ON messages (hdr)")
```
zxdavb, line 168: Some of these need to be unique keys - e.g. hdr.

src/ramses_rf/database.py
```diff
@@ -199,12 +243,16 @@ def _insert_into(self, msg: Message) -> Message | None:
                 msg.code,
                 msg._pkt._ctx,
                 msg._pkt._hdr,
+                json.dumps(msg.payload, indent=4),
+                payload_keys(msg.payload),
```
zxdavb: No! No objects in the SQL DB.

The processing overhead is a disaster - use the record's `dtm` field to pull the packet out of the dict I mentioned above: `{pkt_dtm: pkt}`

src/ramses_rf/database.py
```diff
@@ -215,6 +263,7 @@ def rem(self, msg: Message | None = None, **kwargs: str) -> tuple[Message, ...]:
         if msg:
             kwargs["dtm"] = msg.dtm.isoformat(timespec="microseconds")
 
+        msgs = None
```
zxdavb, line 340: Maybe put these fixes in a precursor PR (to keep you main PR smaller).

src/ramses_rf/device/base.py
```diff
@@ -95,7 +95,7 @@ def _update_traits(self, **traits: Any) -> None:
 
         if traits.get(SZ_FAKED):  # class & alias are done elsewhere
             if not isinstance(self, Fakeable):
-                raise TypeError(f"Device is not fakable: {self} (traits={traits})")
+                raise TypeError(f"Device is not fakeable: {self} (traits={traits})")
```
zxdavb, line 98: separate PR.

src/ramses_rf/device/hvac.py
```diff
@@ -375,7 +377,51 @@ def exhaust_temp(self) -> float | None:
...
+                    k == SZ_FAN_MODE and v != "FF"
+                ):  # Prevent ClimaRad Ventura constant "FF" to pass
+                    info = str(v)
+        _LOGGER.info(f"{info} # FAN_INFO FETCHED from MessageIndex _31D9")  # DEBUG
+
+        sql = """
+            SELECT pl from messages WHERE verb in (' I', 'RP')
+            AND (src = ? OR dst = ?)
+            AND (code = _Code._22F4)
+            AND (plk like %SZ_FAN_MODE% OR plk like %SZ_FAN_RATE%)
```
zxdavb: See my previous comment. Leave this magic for a subsequent PR.

src/ramses_rf/dispatcher.py
```diff
@@ -268,8 +268,8 @@ def logger_xxxx(msg: Message) -> None:
 
     else:
         logger_xxxx(msg)
-        if gwy._zzz:
-            gwy._zzz.add(msg)
+        if gwy.msg_db:
+            gwy.msg_db.add(msg)
```
zxdavb, line 272: pre-cursor PR?

src/ramses_rf/device/hvac.py
```diff
@@ -18,6 +18,8 @@
     SZ_EXHAUST_FLOW,
     SZ_EXHAUST_TEMP,
     SZ_FAN_INFO,
+    SZ_FAN_MODE,
+    SZ_FAN_RATE,
```
zxdavb, line 24: I haven't read below (yet) - but I wonder if you're being over ambitious?

In this PR, do not make any changes to the devices classes. That can come later.

For now, simply wrap the new methods (those that invoke SQL queries) with the old (e.g. `_MessageDB._msg_value()`).

In a later PR - have devices query the SQL DB directly rather than go though those abominations (another e.g. `_MessageDB._get_msg_by_hdr()`).

zxdavb (changes requested):
You're taking on too much - there are at least 3 PRs here:
1. clean up typos/typing/formatting errors of existing code (and harden up ruff config in **pyproject.toml**)

2. abstract the old methods that query the nested dicts with ones that query the SQL DB instead (and get rid of those nested dicts).

3. Switch devices over to using their own queries, rather than using the old methods. You don't have to do them all at once - a good start would be the OT (opentherm) device.

src/ramses_rf/const.py
silverailscolo: import has since been added

src/ramses_rf/database.py
silverailscolo: OK, dropped the payload. Store the keys to filter only the messages that actually contain the keys queried

src/ramses_rf/database.py
silverailscolo: that is defined in CREATE TABLE

src/ramses_rf/device/hvac.py
silverailscolo: took it out

src/ramses_rf/database.py
silverailscolo: included it for now, there seems to be an issue with _HW and we might use cox for that

src/ramses_rf/device/base.py
silverailscolo: In PR #264

src/ramses_rf/dispatcher.py
silverailscolo: in PR #264

src/ramses_rf/device/hvac.py
silverailscolo: SQL queries now used in _msg_value_code, must fine tune but we now follow this 3 step approach, thanks

src/ramses_rf/database.py
silverailscolo: added docstr

src/ramses_rf/database.py
silverailscolo: Took out the JSON, pretty clear

wimpie70:
I'm getting errors like:
```
2025-10-03 12:18:39.905 ERROR (MainThread) [homeassistant.helpers.dispatcher] Exception in add_devices when dispatching 'ramses_cc_new_devices_sensor': ([37:126776 (CO2), 37:168270 (REM), 37:154519 (HVC), 32:153289 (FAN), 29:176861 (REM), 37:169161 (DIS)],)
Traceback (most recent call last):
  File "/config/custom_components/ramses_cc/sensor.py", line 121, in add_devices
    and hasattr(device, description.ramses_rf_attr)
        ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/ramses_rf/device/hvac.py", line 958, in fan_info
    res_mode: list = self._msg_qry(sql)
                     ~~~~~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.13/site-packages/ramses_rf/entity_base.py", line 568, in _msg_qry
    for rec in self._gwy.msg_db.qry_field(sql, (self.id[:9], self.id[:9])):
               ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/ramses_rf/database.py", line 419, in qry_field
    self._cu.execute(sql, parameters)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
sqlite3.OperationalError: near "%": syntax error
```

silverailscolo:
That, and 
`src/ramses_rf/entity_base.py:294: error: Missing return statement  [return]`
Working on it, in steps

By now (5 Oct) it works as library for HVAC. Not for Heat!! as no schema is discovered in tests either.

silverailscolo:
@wimpie70 Your `2411` tests pose a new problem with the SQLite gwy.MessageIndex:

At the end of `mock_gateway()`, we need to async start the SQLite MessageIndex. It takes a little bit of time, including housekeeping to throw out "old" messages. Could be AsyncGenerator or the like, as in: tests/tests/helpers.py:
```
@pytest.fixture
async def gwy() -> AsyncGenerator[Gateway, None]:
```
Please have a look and tweak that part.

```
@pytest.fixture
def mock_gateway() -> Generator[MagicMock, None, None]:
    """Create a mock Gateway instance for testing."""
    gateway = MagicMock(spec=Gateway)
    gateway.send_cmd = AsyncMock()
    gateway.dispatcher = MagicMock()
    gateway.dispatcher.send = MagicMock()

    # Add required attributes
    gateway.config = MagicMock()
    gateway.config.disable_discovery = False
    gateway.config.enable_eavesdrop = False
    gateway._loop = MagicMock()
    gateway._loop.call_soon = MagicMock()
    gateway._loop.call_later = MagicMock()
    gateway._loop.time = MagicMock(return_value=0.0)
    gateway._include = {}
    # Add msg_db attribute that's accessed by the message store
    # gateway.msg_db = MagicMock()
    # gateway.msg_db.get.return_value = {}. < this is now in master
    # activate the SQLite MessageIndex
    gateway.msg_db = MessageIndex(). # < we need this to get _msgs()

    yield gateway
```
Also, the device's self._msgs() dict[code, msg] is rebuilt from the MessageIndex every time it is called, in entity_base.py#_msgs(self) . So adding stuff to it in `handle_message()` won't work.

Also, the MessageIndex has a fixed size of 4 chars for the code (as str), breaking on your extended `2411_dev` keys.
Using SQLite queries we can perhaps get to the same stuff. For example, there is a `plk` field that holds all (non-None) payload keywords, separated by `|`.
Or perhaps in the tests you can use only the _msgb_ dict.

wimpie70:
There's the same problem I had storing the different parameters with the old storage:
- 2411 messages overwrite each other
- the 4 digits are not enough to distinguish the params
- we don't know all params yet. I found someone mentioning params 1-32...
- 2411 are config parameters. They should be kept, not removed and shown unavailable after some time. It may take a long time before a value is refreshed...
I suggest to create a separate table:
```
CREATE TABLE device_parameters (
    device_id TEXT NOT NULL,
    param_code TEXT NOT NULL,     -- "3F", "40", "41", "2411_dev_01"
    param_value TEXT,             -- The actual value as string
    last_updated TIMESTAMP,
    full_payload TEXT,            -- Complete original payload
    PRIMARY KEY (device_id, param_code)
);
```

I'll have a look at the testfile

wimpie70:
Did you see my prev. msg. ? What do you think about a special table for 2411 or maybe better, a config table for values that should not be removed (2411 or other)

silverailscolo:
> a special table for 2411 or maybe better

Could a good ol' dict work for that? If you use the code as key, and the value as such, you'd have all values available.
What would happen after reboot? The cached logs will be interpreted, but that would mean only a single 2411 msg.

wimpie70:
Thought you wanted to do as much as possible in the database... but yes sure, a dict will work, and store it in hvac ? Or at another lvl. Where should it be available ?
(Builder Pattern)
After restart or empty cache, hvac will send a discovery RQ for 1 param. If we get any 2411 I or RP from a fan, it will get_all_parameters ( or update_? Not at my compu), execute a callback and create or update entities in _cc


Op 6 okt 2025, 13:13, om 13:13, Egbert Broerse ***@***.***> schreef:

silverailscolo:
@wimpie70 The way I understand your need to keep multiple 2411 statuses, a plain dict[code, value] can do a lot in python. No setup required.
The power of the SQLite MessageIndex will be of use to gain custom access to payload key:values across Codes. See the "co2_level" demo in hvac.py. Current handling of 4-5 nested msg dicts, especially for a wider range of applications, has reached the limits of dicts (David told me).
Just fixed your `2411` mocked tests by adding a `maintain` parameter to the MessageIndex init. Default is True for daily cleanup of stale msg, not important in tests unless that is what you test ;-)
Also added a unites for MessageIndex. Found that really useful to understand the inner workings. Still a couple of tests to fix.

wimpie70:
Yes, i agree to use a dict. Was working on that, but got a bit distracted when I noticed all the new errors showing up on my _cc PR. Something strange is going on there with the new requirements. Maybe PYTHONPATH is off. Will give it a better look later.
Hope to find some time to fix _rf part first.

Op 6 okt 2025, 22:59, om 22:59, Egbert Broerse ***@***.***> schreef:

PR #210, opened 2025-08-04
Improve MQTT behaviour

IMMRMKW:
My MQTT broker tends to be unstable and occasionally restarts. When this happens, I’ve noticed that the Ramses gateway switches its state from 'ok' to 'problem' and it never recovers back to 'ok', even though the system resumes normal operation.

Currently, the only way to reset the state to 'ok' is by restarting the Ramses CC integration in Home Assistant, which is far from ideal.

I’ve made improvements to the MQTT transport layer in transport.py aiming to make the Ramses-RF integration more robust. The key changes include:

- Error Handling: Instead of closing the transport on MQTT failures, errors are logged and the system continues running.

- Connection State Awareness: Write operations are dropped (with debug logging) when disconnected, preventing exceptions and unnecessary crashes.

- State Tracking: The connection state is now actively monitored.

- Auto-Reconnect: When the connection is lost, the system attempts to reconnect automatically at varying intervals.

I’ve tested these changes over several days with repeated broker restarts, and the gateway has consistently recovered without manual intervention. Even though I did not find other people complaining about this, I think my version is worth merging into the main repository because it improves the robustness.

Let me know what you think or if you'd like me to adjust anything etc.


silverailscolo:
@IMMRMKW thanks for your PR. Please install a venv and the pre-commit hooks (see the wiki) and rerun ruff and mypy. Both these tests currently fail.

silverailscolo:
Please roll back the changes to anything but the file `_tx.transport.py` you initially edited.
By running `ruff format .` only edited files should be checked and updated. 
For mypy there seems to be a version issue between the check running on GitHub and local checks. Please just inspect the errors flagged in the above check.

IMMRMKW:
Ok, I rolled back to my original commit.

First, I tried solving `src/ramses_tx/transport.py:1039: error: Incompatible types in assignment (expression has type "Callable[[Client, Any | None, ReasonCode, Any | None], None]", variable has type "Callable[[Client, Any], None] | None")  [assignment]`

Next, I performed ruff only on *transport.py*. Using: `ruff format .\src\ramses_tx\transport.py`

I find the directions on a PR as found [here](https://github-wiki-see.page/m/ramses-rf/ramses_rf/wiki/1.1-How-to-submit-a-PR) a bit confusing. Running `mypy` results in a big list of errors but the one mentioned above is not in there. Also, I think it is probably wise to first run `mypy`, and then `pre-commit run -a`, since during error fixing, the formatting may be changed again. Also, `pre-commit run -a` changes more files than just the files I changed.

silverailscolo:
Thanks.

> Also, pre-commit run -a changes more files than just the files I changed.

It certainly does. You only need to run `pre-commit run` to focus on files changed in the PR.
I have the same problem on my system, where the changes made by pre-commit hook aren't accepted by GitHub pre-commit. Trying to sort that out...

silverailscolo:
Typing test (failure report) suggests that 3 annotations are no longer required (see Checks > Typing).
You should update your local branch by pulling in the remote changes.

IMMRMKW:
Instead of the ignore statements, I now changed the function handlers. I hope this solves the issue.

silverailscolo:
> I find the directions on a PR as found on the Wiki a bit confusing. Running mypy results in a big list of errors but the one mentioned above is not in there. Also, I think it is probably wise to first run `mypy`, and then `pre-commit run -a`

I agree, especially as I get different results locally. In the end, the GitHub checks are leading, so a PR has to remove the annotations mypy flags.

Be aware that pre-commit on our GitHub repo only looks at changed files (lucky us). So do the same before committing by leaving out -a. From the docs:

> Some example useful invocations:
>
> pre-commit run: this is what pre-commit runs by default when committing. This will run all hooks against _currently staged_ files.
>
> pre-commit run --all-files (or `-a`): run all the hooks against _all_ the files.

The tests again show some small things, hope you have time to fix them 👍

silverailscolo (approved):
although IMHO MQTT only knows about `topic`s, it might be OK for users to word it as `device`.

silverailscolo:
Thanks @IMMRMKW , I plan to soon make a new release with your contribution included

Issue #170, opened 2025-03-16
ruff vs pre-commit


silverailscolo:
I see contradicting behaviour between running ``ruff format .`` (V0.11.0 on macOS)
```
(venv) mac$ ruff format .
5 files reformatted, 83 files left unchanged

(venv) mac$ ruff format --check .
88 files already formatted
```
and then run pre-commit:

```
(venv) mac$ pre-commit run -a
ruff.....................................................................Passed
ruff-format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

5 files reformatted, 83 files left unchanged
```
which I expect to run ``ruff format --check .`` looking at the .github/workflow configuration.
Only the parentheses in these ``assert`` lines seem to be affected (left = by ruff format P3.13; right = by IDE pre-commit hook P3.12?):
![Image](https://github.com/user-attachments/assets/7576e0a7-bf98-4cce-b455-9898ef074a83)

silverailscolo:
Somehow Python 3.12 and Python 3.13 have a subtly different handling of ``ruff``. 

Fixed the GitHub workflow lint test by running ``ruff format .`` in a Python 3.13 virtual environment Terminal session outside of my IDE, and once ready to commit from IntelliJ IDE, unchecked the Commit Options checkbox 'run pre-commit hooks' to be able to push the 5 reformatted files.

The f-string assert message was moved to its own line without parentheses, with the comma separating the assert conditions and message at the end of the previous line.

wimpie70:
I commented out 'ruff format .' in .pre-commit-config.yaml, and run it now myself just before pushing, so I have more control over it.
Before changing this file, I did add it to .gitignore. (and had to untrack .gitignore as well)

silverailscolo:
ruff 0.13.0 was released some time ago. ramses_cc requirements_dev is still at >=0.9.1.
OK to make the jump?

wimpie70:
Sure, i'll update to what's needed. Tnx for pointing at this.

wimpie70:
i noticed ruff 0.13.1 is available

silverailscolo:
HomeAssistant core is now at [ruff 0.13.0](https://github.com/home-assistant/core/blob/dev/requirements_test_pre_commit.txt) so will stay with that. Merged via PR #283 

Every dev should update by running `pip install -r requirements_dev.txt`

silverailscolo:
[...]

PR #258, opened 2025-09-01
Add bypass flag to `31D9` parser

Kars-de-Jong:
Based on packets from my Orcon HRU.


src/ramses_tx/parsers.py
```diff
@@ -2167,6 +2172,7 @@ def parser_31d9(payload: str, msg: Message) -> dict[str, Any]:
         SZ_FAN_MODE: payload[4:6],  # orcon, vasco/climarad
         "passive": bool(bitmap & 0x02),
         "damper_only": bool(bitmap & 0x04),  # i.e. valve only
+        "bypass_mode": "manual" if bitmap & 0x10 else "auto",
```
silverailscolo: To mark it: this is your new item in the 31d9 parser.

The github checks are leading, and I know pre-commit hooks in ramses_rf can ruff to a result that the github linting check rejects.
Manually run `ruff format .` and in your commit disable the hook to prevent that.
I reopened issue #170 as a reminder to fix this.

tests/tests/parsers/code_31d9.log
```diff
@@ -25,36 +25,40 @@
...
-2024-10-15T10:10:35.574076 077  I --- 29:099029 --:------ 29:099029 31D9 003 000001  # {'hvac_id': '00', 'exhaust_fan_speed': 0.005, 'fan_mode': '1 (trickle)', 'passive': False, 'damper_only': False, 'filter_dirty': False, 'frost_cycle': False, 'has_fault': False, '_flags': [0, 0, 0, 0, 0, 0, 0, 0]}  # minimum (Auto mode base speed)
+2024-10-23T12:21:40.179175 074  I --- 29:099029 --:------ 29:099029 31D9 003 000000  # {'hvac_id': '00', 'exhaust_fan_speed': 0.0, 'fan_mode': 'off', 'passive': False, 'damper_only': False, 'bypass_mode': 'auto', 'filter_dirty': False, 'frost_cycle': False, 'has_fault': False, '_flags': [0, 0, 0, 0, 0, 0, 0, 0]}  # off
+2024-10-15T10:10:35.374076 077  I --- 29:099029 --:------ 29:099029 31D9 003 0000C8  # {'hvac_id': '00', 'exhaust_fan_speed': 1.0, 'fan_mode': 'III (boost)', 'passive': False, 'damper_only': False, 'bypass_mode': 'auto', 'filter_dirty': False, 'frost_cycle': False, 'has_fault': False, '_flags': [0, 0, 0, 0, 0, 0, 0, 0]}  # speed 2 = not 100% manuf. error?
```
silverailscolo, line 45: This is not unique in the code, but for this fan msg without a bypass (it's actually in my bathroom;-) the fan_mode flag bit being off is only meaning it's not in 'manual' mode. So the generic value is "inactive". When we pass this bit to a property/sensor named `bypass_state`, assigning the "manual" vs "auto" label is then better handled in the front end/HA, where you can customize how to display inactive/active for your fan.

This is to prevent a situation where a fan that only uses the `22F7` code and not this flag, starts reporting 'auto' for this flag in `31D9` being not set.

src/ramses_tx/parsers.py
silverailscolo: Better use the key `bypass_state` for this flag, same key as the `22F7` parser uses:

`2024-01-01T01:00:00.000008 ...  I --- 32:155617 37:171871 --:------ 22F7 003 00FFC8  # {'bypass_mode': 'auto', 'bypass_position': 1.0, 'bypass_state': 'on'}`

Then it would be 'off' in any manual bypass_mode, wether you set the bypass 'open' or 'closed', see tests/tests/parsers/_22f7.log

Kars-de-Jong:
I disabled the pre-commit hook, I have no idea why it doesn't work properly.
The mypy tests still fail, they pass when I run them locally, no idea why.
Especially since I didn't touch any of the files it complains about...

silverailscolo (changes requested):
Rename the key to `bypass_state` - same as already in use - and pass only the state active/inactive, True/False, not "auto". Bit could be always 0 for other fans, but that is not "manual". See the earlier comments.
As an aside, simply push, don't force push your commits after you created your PR. Our discussion loses context when you force push.

src/ramses_tx/parsers.py
Kars-de-Jong: Sorry, but I don't agree.  This flag does not indicate the state of the bypass valve at all. It purely indicates whether control of the valve is done manually or automatically.

For my Orcon HRU, bypass position is in the `31D9` packet, expressed as a percentage. It is exposed as `bypass_position` in HomeAssistant (as a boolean).

tests/tests/parsers/code_31d9.log
Kars-de-Jong: If the flag is 0 it doesn't mean "inactive", it says nothing about the state of the bypass valve.

I think you really need to have the option of handling packets differently for different fan models / brands.

This is already needed for Orcon as the fan mode for Orcon fans is not in the `31D9` packet but in the `31DA` packet.

silverailscolo: We do combine some packets where it does no harm, but never overwrite values loaded under a different name (I once tried to, but you might have hated that because your information could have been lost).
HVAC manufacturers all do what they like with those packets, as David has explained before. See [Untangling the Organic Knot](https://github.com/ramses-rf/ramses_rf/discussions/191) for details.

Kars-de-Jong: So, like I said, different handling for different models/brands.
This change in `ramses_rf` doesn't overwrite anything, that would only happen if you use it incorrectly in `ramses_cc`.

Kars-de-Jong:
Apparently I got it wrong, this flag is just a boolean for the bypass position, where 0 means closed and 1 means open. So I'm closing this PR for now.

PR #282, opened 2025-09-22
Improve `22F3` parser

### LINKED ISSUE #276: Add support for Orcon specific content of `22F3` code
The `22F3` code is not fully understood yet, but I think I understand it for Orcon fans.
I have two different devices that send these codes. One is a battery powered remote (EAN 08718403862997), the other one a CO2 sensor with button controls for fan speeds (EAN 08718403863376).

When using the remote, `22F3` is sent with a specified mode and duration, and the "return mode" is specified as "use previous". So the fan returns to the previously set mode after the timer has expired.

When using the CO2 sensor, `22F3` is sent with a specified mode and duration, and the "return mode" is specified as "use specified", where the mode is specified in the packet (it always returns to `auto`).

Using this info we can determine the flag bits for Orcon.

The packet from the remote to request speed 3 for 15 minutes:
```
2025-09-06T09:11:31.977068 ...  I --- 29:123456 32:123456 --:------ 22F3 007 00020F03040000  # {'minutes': 15, 'flags': [0, 0, 0, 0, 0, 0, 1, 0], '_new_speed_mode': 'per_vent_speed', '_fallback_speed_mode': None, 'rate': None, '_unknown_5': '0000'}
```

So for Orcon this should actually be: `'_new_speed_mode': 'per_request', '_fallback_speed_mode: 'per_vent_speed', 'rate': 'high'`.

I think it would be better to use `new_rate` and `fallback_rate` instead, because there can be 2 rates in the packet. In this case that would lead to `'_new_speed_mode': 'per_request', '_fallback_speed_mode: 'per_vent_speed', 'new_rate': 'high', 'fallback_rate': None`

The packet from the CO2 sensor to request speed 3 for 60 minutes:
```
2025-09-06T22:31:19.792895 ...  I --- 37:123456 32:123456 --:------ 22F3 007 00123C03040404  # {'minutes': 60, 'flags': [0, 0, 0, 1, 0, 0, 1, 0], '_new_speed_mode': 'per_vent_speed', '_fallback_speed_mode': 'per_request', 'rate': None, '_unknown_5': '0404'}
```

For Orcon this should be (using the suggest names for the rates): `'_new_speed_mode': 'per_request', '_fallback_speed_mode: 'per_request', 'new_rate': 'high', 'fallback_rate': 'auto'`.

This also shows that `_unknown_5` is actually the requested fallback rate, where `'0000'` is sent when the mode is set to `'per_vent_speed'`. It is set to `'0404'` in the other packet, which indeed means mode 'auto' for Orcon fans.

By the way,, this logging also shows that the current attempt to decode the mode by calling the `22f1` parser doesn't work (it always shows `None` in my case).

`10E0` packets moved to separate issue #278

Kars-de-Jong:
Improve 22F3 parser:

 - Parse fan modes correctly using 22F1 parser.
 - Determine new speed mode and fallback speed mode.
 - Rename ~Remove~ _unknown_5 field, it is now known to be the fallback speed mode.

Note that there are still some speed modes in the code of which there are no packets. A question mark was added to their values in case of ambiguity.

Closes #276.


silverailscolo:
Thanks. I pulled in master before running the tests, so you should pull in these upstream changes too

silverailscolo (approved):

silverailscolo:
Is this ready to go?
Is there anything new that you wish to pass on to HA as a `@property` in `ramses_rf/device/hvac.py`?

Kars-de-Jong:
@silverailscolo No, nothing to add to `ramses_rf/device/hvac.py` from this. The fan should change state because of the remote commands, so the new state will come in via other packets.

However, I do see some more different `22F3` (and also `22F1`) payloads near the end of that file, that currently aren't being handled.
Current version of that file is at https://github.com/arjenhiemstra/ithowifi/blob/master/software/NRG_itho_wifi/main/cc1101/IthoPacket.h.
Apparently for Itho remotes / sensors there are different mode sets with different meanings of the mode.
Currently only mode set `04` seems to be handled.

So there is still some room for improvement, but the scope was (mostly) limited to Orcon fans.

PR #295, opened 2025-10-07
Feature/refactor 2411 params storage

wimpie70:
Added a simple _params_2411 dict to hvac devices.
Removed references in param methods to the old message storage (_msgs).
On reset or clear cache, the dict is initialized again. A discovery msg is sent (param 3F) to hvac devices. On a valid response, all values for known parameters are requested and the dict and entities are updated.
No dateTime is registered since these are config values.

### DIFF
--- a/src/ramses_rf/device/hvac.py
+++ b/src/ramses_rf/device/hvac.py
@@ -627,6 +627,8 @@ class HvacVentilator(FilterChange):  # FAN: RP/31DA, I/31D[9A], 2411
         # Normalize the value if needed
         if param_id == "75" and isinstance(param_value, (int, float)):
             param_value = round(float(param_value), 1)
+        elif param_id in ("52", "95"):  # Percentage parameters
+            param_value = round(float(param_value), 3)  # Keep precision for percentages
 
         # Store in params
         old_value = self.get_2411_param(param_id)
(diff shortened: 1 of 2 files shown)


wimpie70:
Seems the mypy-dev version 19 has a problem itself

wimpie70:
I noticed the params class in database. What's the plan with that ?

silverailscolo:
> I noticed the params class in database. What's the plan with that ?

Looks like it applies to message attributes stored in msg_db index, but the SQLite table is defined without them.

silverailscolo:
> Seems the mypy-dev version 19 has a problem itself

Well spotted. Wait till it's fixed and passes or tweak requirements_dev?

silverailscolo:
> Seems the mypy-dev version 19 has a problem itself

Well spotted. Wait till it's fixed and passes or tweak requirements_dev?

wimpie70:
The req_dev om my local installed a lower version.
They will probably fix it soon. But atm it fails on any commit.

>> Seems the mypy-dev version 19 has a problem itself
>
>Well spotted. Wait till it's fixed and passes or tweak
>requirements_dev?

silverailscolo:
See [mypy issue](https://github.com/python/mypy/issues/20006), closed since

wimpie70:
and there's another mypy-dev problem. Maybe only use stable versions instead ?

src/ramses_rf/device/hvac.py
```diff
@@ -446,6 +446,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         """
         super().__init__(*args, **kwargs)
         self._supports_2411 = False  # Flag for 2411 parameter support
+        self._params_2411: dict[str, Any] = {}  # Store 2411 parameters here
```
silverailscolo, line 449: Can you type this more strict in next round?

silverailscolo (approved):

Issue #290, opened 2025-09-30
`2411` param asserts failing


silverailscolo:
Edit: 2411 Params was released in 0.51.9, see ramses_cc [issue 317](https://github.com/ramses-rf/ramses_cc/issues/317)
Regex adapted in PR #308

**Describe the bug**
Out of the blue, I see this in the log after reboot:

```
Logger: ramses_tx.message
Bron: /usr/local/lib/python3.13/site-packages/ramses_tx/message.py:270
Eerst voorgekomen: 00:13:28 (6 gebeurtenissen)
Laatst gelogd: 00:13:29

RQ --- 18:154951 37:153226 --:------ 2411 003 0000DA < AssertionError(param DA is unknown)
RP --- 37:153226 18:154951 --:------ 2411 023 0000DA7F00000000000000000000000003000000018A00 < AssertionError(param DA is unknown)
RQ --- 18:154951 37:153226 --:------ 2411 003 000007 < AssertionError(param 07 is unknown)
RQ --- 18:154951 37:153226 --:------ 2411 003 000088 < AssertionError(param 88 is unknown)
RP --- 37:153226 18:154951 --:------ 2411 023 0000881510000002BC000001900000076C000000018A33 < AssertionError(param 88 is unknown)
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/ramses_tx/message.py", line 255, in _validate
    result = parse_payload(self)
  File "/usr/local/lib/python3.13/site-packages/ramses_tx/parsers.py", line 2973, in parse_payload
    result = _PAYLOAD_PARSERS.get(msg.code, parser_unknown)(msg._pkt.payload, msg)
  File "/usr/local/lib/python3.13/site-packages/ramses_tx/parsers.py", line 1912, in parser_2411
    assert payload[4:6] in _2411_TABLE, (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: param 87 is unknown
```

**Please include the following information:**
 - your **RF Gateway**: ramses_esp (over MQTT)
 - output from the **Serial Console**: see above
 - contemporaneous portion of the **packet.log**:
 ```
2025-10-01T00:01:04.750714 055 RQ --- 18:154951 37:153226 --:------ 2411 003 000087
2025-10-01T00:01:04.774463 079 RP --- 37:153226 18:154951 --:------ 2411 023 0000871400000000000000000000000002000000018A00

[...]

PR #300, opened 2025-10-10
Feature/2411 unknown params fix

wimpie70:
- stricter _params_2411_dict definition in hvac.py as requested
- fix for isue #290: assertion error on unknown parameter id's.
  logs a warning, and returns partial result.

PR for _cc to catch this and log in the UI is on it's way...


wimpie70:
Keep #290 open plz, I still want to decode these params. Maybe change the title..

wimpie70:
Don't merge yet. I need to fix a race condition

silverailscolo (approved):

wimpie70:
yeah, finished. We now get a warning msg in HA log with a request to report it.
On a RP or I from an unknown 2411 param, we do trigger 2411 support and create entities for known params.
Race condition happened when receiving an unknown param right after restart. With value = None support would no longer be detected.
No more need to create a _cc PR for this.

PR #306, opened 2025-10-14
Add `2411` param `01`

silverailscolo:
fix issue #290 

- tweak 2411 regex for len=22
- add param 01 (min_value = max_value= 0xFF)
- add MiniBox 2411 reply to parsers test log

### DIFF
--- a/tests/tests/parsers/code_2411_wip.log
+++ b/tests/tests/parsers/code_2411_wip.log
@@ -1,10 +1,8 @@
+# Orcon RF15 (with display)
 
 # From: https://gathering.tweakers.net/forum/list_messages/2135834/3
 # Set time to replace filter in menu option 10 from 180 to 150
 # Set after-run time moisture sensor (min) menu option 13 from 15 to 20
-
-
-# Orcon RF15 (with display)
 
 ### (31) Config settings - Time to change filter (days)
 2022-06-05T10:25:01.228218 067 RP --- 32:155617 37:171871 --:------ 0001 008 0020000900310A00                                # {'param_num': '31', 'slot_num': '09', 'next_slot_num': '0A', 'boolean_14': False, 'payload': '0020000900310A00'}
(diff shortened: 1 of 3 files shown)


silverailscolo:
@wimpie70 would this change be okay?

wimpie70:
Adding it to _2411_PARAMS_SCHEMA will create an entity for it. Not sure what it does when changing a value for it. Did you test this ?

I'm creating a terminal util: 2411_parser.py. It shows parsed values from a msg in blocks of 4, 6 and 8 bytes in different value formats. It might be handy for decoding.
I was planning to save it to ramses_rf/misc/2411_parser.py But maybe its better to put it along with ramses_rf/tests/tests/parsers/code_2411_wip.log ?
btw, i knew these logs were somewhere, forgot where to look :D

We might even use it in the codebase to help users decode it themselves ? Since they know what values change or might be related to their current states ? But it's quite a bit output... Tell me where to put it and i'll create a PR so you can have look.

silverailscolo:
In my (recent) experience, tests - especially unittests - are very useful to play with a certain feature. That's why I added a dispatcher.py test to go with the MessageIndex.
The "real" test takes care the whole system keeps working, but they don't give a lot of insight into what a function does.
Does your parsing tool require a running ramses setup with gateway, live dongle and all, or does it work with just the 2411_parser?

wimpie70:
only a parser, see docs in the file and #308

tests/tests/parsers/code_2411_wip.log
```diff
@@ -261,3 +259,9 @@
 2022-06-05T10:03:59.569545 067  I --- 32:155617 37:171871 --:------ 2411 023 0000953F0F000000C800000000000000C800000001A632
 2022-06-05T10:04:01.517182 075 RQ --- 37:171871 32:155617 --:------ 2411 003 000095
 2022-06-05T10:04:01.555434 071 RP --- 32:155617 37:171871 --:------ 2411 023 0000953F0F000000C800000000000000C800000001A632
+
+# ClimaRad MiniBox extraction FAN
+2025-10-14T11:04:50.340691 000 RQ --- 18:130140 29:099029 --:------ 2411 003 00003F
+# response:
+2025-10-14T00:00:00.000001 ... RP --- 29:099029 18:130140 --:------ 2411 022 00000100000000003200000000000000FF0000000120
+# meaning "not supported"? param_id 01 = ALL?, FF = None?
```
silverailscolo: I'd expect that the pre-commit hook would require an empty line at the end

PR #311, opened 2025-10-17
SQlite MessageIndex option

silverailscolo:
Big refactor as part of route to more flexible _Ramses RF_ type support, see [Discussion](https://github.com/ramses-rf/ramses_rf/discussions)
Stores and retrieves all Messages in/from an SQLite database table (for data structure see `database.py` class `MessageIndex` a.k.a. `msg_db`).
Inits during `gateway/start()`
For now, use of msg_db must be activated by user in ramses_cc config_flow (PR #[324](https://github.com/ramses-rf/ramses_cc/pull/324))
Will only default to use `msg_db` when user reports are good enough.
Replaces all uses of _msgz nested dicts per entity with custom sql query methods to lookup from the central Messages index by `dtm` timestamp, see database and entity_base.
Deprecates legacy nested dict `_msgz_`, to be removed per Q1 2026.
Bumps ramses_rf version to 0.52.0. Requires `ramses_cc` UI of same tag to access the switch.

Includes 2 unittests for `_rf.database` and `_rf.dispatcher`.
Existing tests passing, with 2 test messages corrected, but our tests don't cover that much.
Tested locally with HVAC and runs without error, including turning option on/off in UI + restart.

### Squash Merge!
Too many merge conflicts solved during development
New code based on #196, closed since


silverailscolo:
Merged as 5 sub-PR's:
- [x] src/ramses_cli/ #312 
- [x] src/ramses_rf/ (excluding gateway.py) #313 
- [x] tests/tests/ #315 
- [x] tests/tests_rf/ #314 
- [x] src/ramses_rf/gateway.py (activates msg_db) #316

silverailscolo:
Merged as 5 sub-PRs, closing this

PR #343, opened 2025-11-12
Bump pytest, close `sqlite3` dB in tests

silverailscolo:
- Fix: When running with pytest-cov, tests leaving behind unclosed `sqlite3` MessageIndex connections were reported
- Fix some docstr typos
- Bumps `pytest` and `pytest-asyncio` to latest/HA

Conclusion after trying to activate cov-reports in several configurations: this lib can't be tested with with the pytest-cov plugin, probably because of the protocol timing (see notes in code). Will create a separate GitHub Action for coverage.


silverailscolo:
3.11 `tests/tests_rf/test_virt_network.py::test_virtual_rf_dev_disc` failed 1x (timeout)

PR #347, opened 2025-11-15
add 2411_DATATYPES: 20 unknown

wimpie70:
This would fix the warning on 2411, param datatype 20 (param _01), https://github.com/ramses-rf/ramses_cc/issues/342#event-20962303184

But...it would also stop users from reporting these messages...


wimpie70:
i closed this...changed more than i needed for this PR...

silverailscolo:
> But...it would also stop users from reporting these messages...

Could you make the filter/warning more specific?
That follows zxdavb's strategy with ramses parsers and regex: only allow to pass what is seen at least once in real use.

PS You might also want to collect packets from users. We can me more strict to require those to accept an issue.

PR #361, opened 2025-12-14
Fix/mqtt startup stability

PWhite-Eng:
## Description
Improves the stability of the MQTT transport layer.

1. **Fix Zombie Connections:** If `wait_for_connection_made` times out (default increased to 60s), the transport is now explicitly closed. This prevents the transport from connecting successfully *after* the timeout and triggering callbacks on a cancelled future, which previously caused `AttributeError: 'NoneType'` crashes in `PortProtocol`.
2. **Robustness:** `PortProtocol.connection_made` now safely checks if `transport` is set before accessing it, handling edge cases where initialization was interrupted.
3. **Read-Only Support:** The `disable_sending` flag is now correctly passed to `MqttTransport`, ensuring read-only configuration is respected.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/ramses_tx/protocol.py
+++ b/src/ramses_tx/protocol.py
@@ -66,7 +66,7 @@ class _BaseProtocol(asyncio.Protocol):
         self._msg_handler = msg_handler
         self._msg_handlers: list[MsgHandlerT] = []
 
-        self._transport: RamsesTransportT = None  # type: ignore[assignment]
+        self._transport: RamsesTransportT | None = None
         self._loop = asyncio.get_running_loop()
 
         self._pause_writing = False  # FIXME: Start in R/O mode as no connection yet?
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
I have pushed an update to this PR to fix a critical crash I encountered during testing.

**The Issue:** When the integration timed out during setup (the 60s "Transport did not bind" warning), the transport state changed to "closing." However, the MQTT client thread was still active and received a message. This triggered a `TransportError("Transport is closing or has closed")`, which bubbled up and caused an `Uncaught thread exception` in `paho.mqtt.client`, effectively crashing the listener thread.

**The Fix:** I added a `try...except TransportError` block in `MqttTransport._on_message`. If the error occurs while the transport is closing, it is now safely ignored instead of crashing the thread.

**Verification:** I have tested this on my Home Assistant system. The logs confirm that when the timeout occurs, the crash is gone. The system now logs a graceful `connection_lost` message, successfully recovers, and connects fully on the subsequent retry.

silverailscolo:
Thanks. mypy flags a statement in `protocol.py` as "unreachable".

PWhite-Eng:
I have resolved the mypy static analysis error in tests/tests_rf/test_protocol_fsm.py.

The error `Item "None" of "FileTransport | MqttTransport | PortTransport | None" has no attribute "_extra"` occurred because the `_transport` attribute is optional in the source code. I added an `assert protocol._transport is not None` check to `_test_flow_30x`, which verifies the test fixture setup and correctly narrows the type for mypy.

I have run the full test suite locally (on Linux/WSL), and all 546 tests passed successfully.

silverailscolo (approved):

PR #362, opened 2025-12-14
Fix: Ensure commands use active Gateway ID instead of default 18:000730

PWhite-Eng:
## Description
This PR modifies `PortProtocol.send_cmd` in `src/ramses_tx/protocol.py` to fix an issue where commands were being sent with the wrong Source ID.

## The Issue
Previously, commands created without an explicit source address defaulted to the hardcoded HGI address `18:000730`. When the actual physical gateway had a different ID (e.g., `18:005960`), this caused:
1. "Foreign gateway" warnings in the logs.
2. Potential packet rejection or routing issues as the source address did not match the physical hardware.

## The Fix
This change intercepts commands at the `send_cmd` layer. It checks if the command is using the default placeholder (`18:000730`) and patches it to use the discovered Active Gateway ID (`self.hgi_id`) immediately before transmission.

## Verification
Tested locally with a physical Evohome system. 
- **Before:** Logs showed mismatched source IDs and foreign gateway warnings.
- **After:** The system correctly infers the gateway ID (`18:005960`). While the internal frame logger still shows the default ID (pre-patch), the physical transmission is correct, evidenced by the absence of "Foreign gateway" warnings and successful bi-directional communication with the controller.

### DIFF
--- a/src/ramses_tx/protocol.py
+++ b/src/ramses_tx/protocol.py
@@ -80,6 +80,9 @@ class _BaseProtocol(asyncio.Protocol):
 
         self._is_evofw3: bool | None = None
 
+        self._active_hgi: DeviceIdT | None = None
+        self._context: ProtocolContext | None = None
+
     @property
     def hgi_id(self) -> DeviceIdT:
         return HGI_DEV_ADDR.id
@@ -198,6 +201,10 @@ class _BaseProtocol(asyncio.Protocol):
         """
 
         self._pause_writing = False
+
+    async def _send_impersonation_alert(self, cmd: Command) -> None:
+        """Allow the Protocol to send an impersonation alert (stub)."""
+        pass
 
     async def send_cmd(
         self,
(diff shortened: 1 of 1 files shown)


silverailscolo:
Thanks. As is to be expected, some test asserts need to be adjusted as a result.
mypy has a few points too.

silverailscolo:
Tried to add an inline review  comment for "pass" but the app refused to store it, so:

This is not the preferred, I think. Consider a mypy hint to fix 'unreachable'.

In the connected tests, a different id is returned:

silverailscolo (changes requested):
Tried to add an inline review comment for "pass" but the Android app refused to store it, so:

In the connected tests, a different id is returned then before. Can you update that?

PWhite-Eng:
@silverailscolo Thank you for the review and the guidance!

I have updated the code to address your feedback:

Tests: All local tests (including tests_tx) are passing.

Changes have been pushed. Please let me know if you spot anything else!

silverailscolo:
One more hint:

If you write docstr exception descriptions as
`:raises yourErrorType:`
the Spinx docs will be created cleaner (see fan_param for examples).

Same for
`@param paramName:`
which are generated when you create a new docstr by entering the opening """ + return

silverailscolo (approved):

PR #370, opened 2025-12-18
Add `cli` typing, tests

silverailscolo:
to improve `cli` module's lagging coverage


silverailscolo:
Failing Coverage test is same failure as seen on current master:
```
=========================== short test summary info ============================
FAILED tests/tests_rf/test_hgi_behaviors.py::test_fake_evofw3[60] - ramses_tx.exceptions.ProtocolSendFailed: <ProtocolContext state=WantEcho cmd_=0008| I|18:222222|00, tx_count=1/4>: Expired global timer after 0.006 sec
```
@PWhite-Eng You seem to have fixed this in your PR #378 Any hints?

PWhite-Eng:
Hi @silverailscolo,

I have investigated the `ProtocolSendFailed` errors appearing in the CI checks on `master` since the merge of PR #375. I’ve verified the behavior locally and analyzed why my draft PR (#378) appears to fix it.

Here is the breakdown of the cause and the recommended solution.

### 1. The Root Cause: CI Race Condition

The failure occurs in `tests/tests_rf/test_hgi_behaviors.py`. The test enforces a strict **6ms** timeout for the Virtual RF stack to initialize:

Python

```
# tests/tests_rf/test_hgi_behaviors.py (approx line 60)
elif gwy._transport.get_extra_info("virtual_rf"):
    timeout = 0.003 * 2  # 6ms total

```

**Why it fails now:** PR #375 (now in `master`) introduced the necessary abstraction layers for the Transport Factory. While this architecture is correct, the additional layering added a tiny amount of initialization overhead (approx. 1-2ms). On shared GitHub runners, this slight increase pushes the total execution time just over the **6ms** limit, causing the `Expired global timer` error.

### 2. Verification (Local Testing)

I checked out `master` (including #375) to my local development environment and ran the test suite:

Bash

```
pytest tests/tests_rf/test_hgi_behaviors.py

```

**Result:** **PASS** (24 passed, 36 skipped). This confirms that the **code logic in `master` is correct**. The failure is strictly a performance race condition specific to the variable latency of the GitHub CI runners.

### 3. Why Draft PR #378 Fixes It

You noticed that my draft PR (#378) passes this test. This is a positive side effect of the architectural completion in that PR.
- **On `master`:** The `Gateway` still relies on some legacy internal logic to resolve and create the packet stack. This setup runs synchronously during `__init__` and counts toward the timeout.
- **On PR #378:** I updated `client.py` and `gateway.py` to fully implement the IoC pattern, allowing us to **inject** the `transport_factory` directly. This bypasses the legacy setup overhead, shaving off the 1-2ms needed to reliably beat the 6ms timer.

### 4. Recommendation

**Immediate Fix for Master:** Since PR #378 is still in draft and requires proper review, we should not rush it to resolve these timing issues. Instead, we can stabilize `master` by acknowledging that the current 6ms limit leaves insufficient margin for the new architecture on the shared GitHub runners

I recommend we relax the timeout in `tests/tests_rf/test_hgi_behaviors.py`:

Python

```
# Increase from 0.003 (6ms) to 0.010 (20ms) to account for CI jitter
elif gwy._transport.get_extra_info("virtual_rf"):
    timeout = 0.010 * 2 

```

**Long Term:** Once PR #378 is merged, the codebase will naturally return to the higher performance characteristics, but the relaxed timeout will ensure we don't see these false negatives in the future.

Would you like me to push a quick PR to `master` to adjust this timeout?

silverailscolo:
> Would you like me to push a quick PR to master to adjust this timeout?

That would be nice, thanks for the analysis report. Agree on the cause and fix

PWhite-Eng:
Done. I have raised #380 to apply this fix.

It increases the virtual RF timeout to 0.020s (20ms), which should provide plenty of headroom for the CI runners to complete the initialization without triggering the race condition.

silverailscolo:
> Done. I have raised #380 to apply this fix.

Worked, thanks

PR #372, opened 2025-12-22
fix(const): resolve incompatible type assignments in AttrDict

PWhite-Eng:
## Description
Refactors the `AttrDict` class in `const.py` to resolve Mypy errors regarding incompatible type assignments.

## Problem
Previously, `AttrDict` made methods read-only by assigning a generic `__readonly` class method to `__setitem__`, `pop`, etc. Mypy flagged this as an error because the signature of the generic method (taking `*args`) did not strictly match the signatures defined in the base `dict` / `MutableMapping` classes (e.g., `pop(key, default)`).

## Solution
Replaced the method assignments with explicit method definitions that call a `_readonly` helper. This satisfies the Liskov Substitution Principle required by strict type checkers while maintaining the read-only behavior.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/ramses_tx/const.py
+++ b/src/ramses_tx/const.py
@@ -157,19 +157,29 @@ class AttrDict(dict):  # type: ignore[type-arg]
     _SZ_DEFAULT: Final = "_default"
     _SZ_SLUGS: Final = "SLUGS"
 
-    @classmethod
-    def __readonly(cls, *args: Any, **kwargs: Any) -> NoReturn:
-        raise TypeError(f"'{cls.__class__.__name__}' object is read only")
+    def _readonly(self, *args: Any, **kwargs: Any) -> NoReturn:
+        raise TypeError(f"'{self.__class__.__name__}' object is read only")
 
-    __delitem__ = __readonly
-    __setitem__ = __readonly
-    clear = __readonly
-    pop = __readonly
-    popitem = __readonly
-    setdefault = __readonly
-    update = __readonly
+    def __setitem__(self, key: Any, value: Any) -> NoReturn:
+        self._readonly()
 
-    del __readonly
+    def __delitem__(self, key: Any) -> NoReturn:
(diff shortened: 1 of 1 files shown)


PWhite-Eng:
I wanted to note a finding regarding the test suite.

While validating my changes locally, I encountered a failure in `tests/tests/test_vol_schemas.py::test_known_list_bad[5]`. I verified that this failure also occurs on the `master` branch and is not caused by the changes in this PR.

The root cause appears to be the recent release of `voluptuous==0.16.0`, which seems to be more permissive with extra keys than previous versions, causing the "bad config" test to pass validation unexpectedly.

I confirmed that downgrading to `voluptuous<0.16.0` resolves the test failure on both `master` and this branch. Since `requirements_dev.txt` specifies `>=0.15.2`, new environments (like CI runners) might pick up the breaking version automatically.

PWhite-Eng:
Blocked: waiting for PR #373 to merge to fix CI failures.

silverailscolo:
I think we should better remove the temporary fixes instead. See note in PR #373

PWhite-Eng (commented):
Hi @silverailscolo,

Understood. I have just reverted the commits that pinned voluptuous to keep this PR aligned with the project's direction.

Aside from the dependency issue, this PR contains the refactor for AttrDict to resolve the Mypy incompatible type assignment errors. Since the voluptuous issue is still causing the CI to fail on Python 3.14, how would you like me to proceed with this PR?

Should I keep this in Draft until the global CI issues are resolved?

Or would you prefer I move it to Ready for Review so you can check the Mypy fixes (and potentially bypass the failing coverage test to merge)?

I'm happy to do whichever is easier for your workflow.

silverailscolo:
Merging this PR now that the `vol` issue has been fixed.

silverailscolo (approved):

PR #373, opened 2025-12-22
fix(deps): pin voluptuous < 0.16.0 to resolve CI test failures

PWhite-Eng:
## Description
Pins the `voluptuous` dependency to `< 0.16.0` in `requirements/requirements_dev.txt`.

## Problem
The recent release of `voluptuous` 0.16.0 introduced breaking changes regarding how extra keys are handled in schema validation. This caused `tests/tests/test_vol_schemas.py::test_known_list_bad` to fail because invalid configurations were no longer being rejected by the validator.

## Solution
Restricting the version to `< 0.16.0` ensures the test suite runs against a compatible version of the library, resolving the build failures on CI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@
     "colorlog>=6.9.0",              # latest
     "paho-mqtt>=2.1.0",             # latest
     "pyserial-asyncio-fast>=0.16",  # latest
-    "voluptuous>=0.15.2",           # latest
+    "voluptuous>=0.15.2,<0.16.0",   # pinned to fix CI failures
   ]
 
 #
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
@silverailscolo 
### Request for Assistance regarding CI

I have been attempting to resolve the CI failures caused by the breaking changes in `voluptuous==0.16.0`.

1. I verified locally that `voluptuous < 0.16.0` resolves the `test_known_list_bad` failures.
2. I pinned the version to `< 0.16.0` in `requirements/requirements_dev.txt`.
3. I also pinned the version in `pyproject.toml` dependencies.

Despite these changes, the CI runner persists in installing version 0.16.0 during the build step, causing the tests to fail.

As the maintainer, do you have the ability to update the project without having to run the automated tests?

silverailscolo:
Although unrelated to your PR, it seems better to align the test with latest Voluptuous instead of pinning it. 
There's actually a [PR from our project](https://github.com/alecthomas/voluptuous/pull/524) included in the latest Vol release to fix support for `RemoveExtras`. So we should better remove the temporary fixes instead.

(The Coverage test currently failing runs on Python 3.14)

PWhite-Eng (commented):
Hi @silverailscolo,

Thank you for the clarification, that makes perfect sense
I'll close this PR

Thanks for the guidance

PR #375, opened 2025-12-26
feat: implement IoC transport factory for MQTT support

PWhite-Eng:
### Summary 

This PR introduces an **Inversion of Control (IoC)** pattern to the transport layer. It allows external integrations (like `ramses_cc`) to inject a generic "Connection Builder," enabling native MQTT support without creating a hard dependency on Home Assistant or `paho-mqtt` within this core library.

### Proposed Changes

- **Refactor:** Updated `Gateway` and `transport_factory` to accept a `transport_constructor` argument.
- **New Feature:** Implemented `CallbackTransport` in `src/ramses_tx/transport.py`. This acts as a "Virtual Serial Port" that accepts an `io_writer` callback and exposes a `receive_frame` method.
- **Backward Compatibility:** If no external transport is provided, the library defaults to the existing Serial/USB logic. No existing behavior is changed.

### Design Context

The goal is to keep `ramses_rf` platform-agnostic. Instead of deciding how to connect to MQTT internally, the library now passively accepts a transport factory. This delegates all HA-specific logic (JSON unpacking, topic monitoring) to the integration layer.

### Testing

- Added unit tests using a Mock Transport Constructor to verify the factory logic without running a full HA instance.
- Verified "Virtual Serial Port" behavior (Write Path and Read Path).


PWhite-Eng:
Hi @silverailscolo (and @zxdavb),

Thanks for pointing me toward Issue #359 and the HA blog post. I’ve read through it, and these two Pull Requests (this one and the companion in ramses_cc) are the result.

I have set these as Drafts because this is my first time working on an architecture of this size, and it spans two projects. I've been running this on my local standalone Raspberry Pi, and it seems to be working great, but I want to respect the project's standards.

**My main questions for you:**

1. Is this "Inversion of Control" approach (injecting the transport rather than hardcoding MQTT logic) acceptable to you?
2. Do you want to pull this feature into the project?

I am happy to put in more work to get this to a point where you are happy to accept it, but I would love some guidance on what additional work (if any) is needed before I mark it as ready for review.

Thanks for the guidance so far

silverailscolo:
Open to your idea, but in my opinion you tend using too many CAPS (logging and method names). If your goal is just to spot specific messages, use a find command on the log.

PWhite-Eng:
Thanks for the feedback @silverailscolo , you were right about the styling - I had left a lot of my personal debugging "scaffolding" (CAPS and visual tags) in the code.

I have pushed a new commit that cleans this up:

- **Refactored Logging:** Removed all [TRACE] tags and uppercase messages. Logs are now standard sentence-case and use f-strings for readability.
- **Linting:** Ran ruff across the project to catch and fix formatting violations.

**Regarding Naming Conventions:** I performed a manual review of the method (snake_case) and class (PascalCase) names, and I did not find any that required changing to meet the standard conventions. However, I might be too deep in the code to spot every instance. If you do see any specific violations I missed, could you please point them out?

The core architecture (Inversion of Control) remains unchanged. Ready for your review and further comment

silverailscolo (approved):

silverailscolo:
Coverage (running on python 3.14!) repeatedly fails with `src/ramses_tx/protocol_fsm.py:363: ProtocolSendFailed`

The new pattern sounds interesting. Perhaps, it requires a different test setup?

Also, can the `client.py` use the same mechanism, or would it revert to a legacy "built-in" connection?

PWhite-Eng:
Hi @silverailscolo ,

I have updated the branch `feat/ha-mqtt-ioc-support` by merging the latest `upstream/master`.

GitHub Actions now reports that all checks (including the Coverage workflow) are passing.

Regarding the `ProtocolSendFailed` errors and the Python 3.14 coverage issues you observed: I believe these were side effects of the persistent CI failure caused by the `voluptuous` dependency conflict (specifically the `test_known_list_bad` case). As you noted, the `check-cov` workflow triggers against master, so the broken test case was causing the pipeline to fail or behave unpredictably regardless of the changes in my feature branch.

Now that PR #376 (which removed the outdated test case) has been merged and I have synced my branch with `master`, the CI environment is consistent and the tests are passing cleanly.

The PR is ready for your review.

Phil

PWhite-Eng:
### To answer your questions about the IoC pattern:

**1. Does it require a different test setup?** Yes, it enables a much more flexible test setup. Instead of relying on the `Gateway` to implicitly create a transport (which is hard to mock), we can now inject a factory function. You can see this in `ramses_rf/tests/test_HA_MQTT/test_transport_callback.py`. The test defines a simple `mock_factory` and passes it to the `Gateway` via the new `transport_constructor` argument:

```Python
gwy = Gateway("/dev/null", transport_constructor=mock_factory)
```
This allows us to inject a fully mocked transport before the Gateway even starts, making unit testing of the protocol stack much cleaner.

**2. Can `client.py` use the same mechanism?** Currently, `client.py` reverts to the legacy "built-in" connection logic. While the `Gateway` class now supports IoC, the CLI tool (`client.py`) is designed to take a simple string for the port (e.g., `/dev/ttyUSB0`). To support the IoC pattern from the CLI, we would need to add a way to pass a Python callable (factory) via a command-line argument, which `client.py` does not currently do. For now, it continues to function exactly as before using the standard transport factory.

src/ramses_tx/protocol.py
silverailscolo, line 1: @PWhite-Eng Since merging PR #375 into `master`, the Python3.14 Coverage check always fails, specifically `tests/tests_rf/test_hgi_behaviors.py`.
Could you help find the cause of the time out?
Prefer not to release as is.

PWhite-Eng: Agreed, we definitely shouldn't release while the tests are failing.

I have just posted a detailed response to your comment on #370 with the root cause analysis (it is a race condition) and a proposed fix.

PR #376, opened 2026-01-01
test: resolve CI failure by removing outdated known_list_bad test case

PWhite-Eng:
## Description
This PR resolves the persistent failure in the `check-cov` workflow (specifically `test_known_list_bad[5]`).

## The Problem
The workflow `check-cov.yml` triggers on `pull_request_target`, running tests against `master` while installing the latest dependencies.
* **The Dependency:** `voluptuous` released version `0.16.0+` which includes a fix for `RemoveExtras` (stripping extra fields).
* **The Conflict:** The existing test `test_known_list_bad` passes a configuration with a `_note` field, expecting it to **fail** validation.
* **The Result:** Because `voluptuous` now correctly accepts and strips the `_note` field (as defined in the schema via `vol.Remove`), the validation **succeeds**, causing the negative test to fail.

## The Solution
I have removed the specific test case containing `_note` from `KNOWN_LIST_BAD`.
* The schema correctly handles `_note` (stripping it), so it is no longer "bad" data.
* This fix unblocks the CI pipeline for all other Pull Requests.

### DIFF
--- a/tests/tests/test_vol_schemas.py
+++ b/tests/tests/test_vol_schemas.py
@@ -211,18 +211,18 @@ KNOWN_LIST_BAD = (
     # known_list:
     #   05:111111: {class: REM, scheme: xxxxxx}
     # """,
-    """
-    known_list:
-      01:111111: {class: FAN}
-      02:111111: {class: RFS}
-      03:111111: {class: CO2, faked: true}
-      04:111111: {class: HUM, faked: true}
-      05:111111: {class: REM, faked: true, scheme: nuaire}
-      06:111111:
-        class: DIS
-        scheme: orcon
-        _note: this is a note
-    """,
+    # """
+    # known_list:
+    #   01:111111: {class: FAN}
+    #   02:111111: {class: RFS}
+    #   03:111111: {class: CO2, faked: true}
+    #   04:111111: {class: HUM, faked: true}
+    #   05:111111: {class: REM, faked: true, scheme: nuaire}
+    #   06:111111:
+    #     class: DIS
+    #     scheme: orcon
+    #     _note: this is a note
+    # """,
 )
 KNOWN_LIST_GOOD = (
     """


PWhite-Eng:
Hi @silverailscolo,

I have identified why this check continues to fail despite the fix being present in this Pull Request.

**Executive Summary:**
The failure is a **false positive** caused by the CI configuration. The "Coverage" workflow is running the **old code from `master`** (which still has the broken test) against the **new `voluptuous` library**, effectively ignoring the fix in this PR.

**Technical Detail:**
1. **The Workflow Trigger:** The `.github/workflows/check-cov.yml` workflow is configured to run on `pull_request_target`.
   ```yaml
   on:
     pull_request_target:
       branches: [ "master" ]
   ```
2. **The Checkout Behavior:** By default, when `actions/checkout` runs on `pull_request_target`, it checks out the base branch (`master`), not the code in this Pull Request.
3. **The Dependency Conflict:**

    - **Old Voluptuous (<0.16.0)**: Had a bug where `vol.Remove` keys were often incorrectly flagged as illegal when `extra=vol.PREVENT_EXTRA` was set. The test `test_known_list_bad` relied on this bug to pass (expecting `_note` to fail validation).
    - **New Voluptuous (>=0.16.0)**: Includes a fix (mentioned in previous discussions) that correctly allows `vol.Remove` keys to pass `PREVENT_EXTRA` checks.

4. **The Failure:** - The runner checks out `master`, so it loads the old version of `tests/tests/test_vol_schemas.py` that still contains the outdated test case `test_known_list_bad[5]`.

    - The runner then installs the latest dependencies (including `voluptuous >= 0.16.0`).
    - The new `voluptuous` correctly handles `RemoveExtras`, causing the old negative test to fail (it now passes validation, but the test expects it to fail).

**Conclusion:** This PR **does** fix the issue (by removing the outdated test case), but the CI runner cannot see the fix because it is testing the `master` branch.

Please **ignore the failing Coverage check and merge this PR**. Once merged, `master` will be updated, and future checks (including my other PRs) will pass.

silverailscolo:
@PWhite-Eng Thanks for putting in the time to look at this closely. Superficially checking the result didn't explain what exactly was happening. Running this test om master instead of the new branch has to do with permissions.

Let's merge and wait for confirmation.

silverailscolo (approved):

silverailscolo:
ramses_rf@master Coverage Action now passing! Thanks again

PR #380, opened 2026-01-05
fix: increase virtual RF timeout to 20ms for CI stability

PWhite-Eng:
## Description
Increases the timeout for virtual RF device discovery in `test_hgi_behaviors.py` from 6ms to 20ms.

## Motivation and Context
As discussed in #370 and #375, the recent architectural refactoring of the Transport Factory introduced a minor initialization overhead (1-2ms). While this does not affect functionality, it causes a race condition on shared GitHub CI runners, leading to intermittent `ProtocolSendFailed: Expired global timer` errors when the execution time exceeds the previous strict 6ms limit.

## How Has This Been Tested?
- Verified locally on WSL environment (tests pass consistently).
- This change purely adjusts the test harness tolerance to account for CI environment latency.

## Closing Issues
Resolves the CI failures reported in #375 regarding `test_fake_evofw3`.

### DIFF
--- a/tests/tests_rf/test_hgi_behaviors.py
+++ b/tests/tests_rf/test_hgi_behaviors.py
@@ -108,7 +108,7 @@ async def _test_gwy_device(gwy: Gateway, test_idx: int) -> None:
     if isinstance(gwy._transport, MqttTransport):  # MQTT
         timeout = 0.375 * 2  # intesting, fail: 0.370 work: 0.375: 0.75 margin of safety
     elif gwy._transport.get_extra_info("virtual_rf"):  #   # fake
-        timeout = 0.003 * 2  # in testing, fail: 0.002 work: 0.003: 0.006 margin of ...
+        timeout = 0.010 * 2  # was 0.003 * 2: increased to 20ms for CI stability
     else:  #                                        # real
         timeout = 0.355 * 2  # intesting, fail: 0.350 work: 0.355: 0.71 margin of safety
 
--- a/tests/tests_rf/virtual_rf/__init__.py
+++ b/tests/tests_rf/virtual_rf/__init__.py
@@ -90,12 +90,12 @@ async def rf_factory(
 
     for idx, schema in enumerate(schemas):
         if schema is None:  # assume no gateway device
-            rf._create_port(idx)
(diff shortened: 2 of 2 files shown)


PWhite-Eng:
## Fix: Race Condition in `virtual_rf` Factory Initialization

This change resolves the intermittent `SerialException` failures observed in `test_binding_fsm.py` (and potentially other tests using `virtual_rf`) on Python 3.11/3.12 CI runners.

### The Root Cause
The issue is a **race condition** caused by concurrent modification of the port list during test initialization.

1.  **Setup Start:** `rf_factory` initializes the test environment by calling `VirtualRf(len(schemas))`.
2.  **Background Task:** The `VirtualRf` constructor immediately creates **all** requested ports and starts a background `asyncio` task (`_poll_ports_for_data`) to handle traffic. This task iterates over the internal `self._port_to_object` dictionary.
3.  **Concurrent Modification:** While this background task is running, the `rf_factory` loop proceeds to configure the gateways. However, it was incorrectly calling `rf._create_port(idx)` **again** for each schema.
4.  **The Crash:** Calling `_create_port` modifies the `self._port_to_object` dictionary. If the background task attempts to iterate over this dictionary at the exact moment it is being modified, it raises a `RuntimeError: dictionary changed size during iteration`.
5.  **The Fallout:** This unhandled exception crashes the polling task and/or closes the master PTY file descriptors. When the test runner (e.g., `Gateway.start()`) subsequently tries to connect to the slave PTY, it encounters a broken pipe or immediate EOF, resulting in the `SerialException: device reports readiness to read but returned no data` error seen in the CI logs.

This race is timing-dependent, which explains why it passed on faster runners (Python 3.13) where the factory loop likely completed before the background task hit the contention point.

### The Fix
I have removed the redundant `rf._create_port(idx)` calls from the loop in `ramses_rf/tests/tests_rf/virtual_rf/__init__.py`.

### Why It Works
The `VirtualRf` constructor already creates the exact number of ports required based on the `len(schemas)` argument. The subsequent calls to `rf.set_gateway(...)` are sufficient to configure these existing ports. By removing the redundant creation step, we eliminate the concurrent modification of the port list, ensuring the background polling task remains stable throughout the test lifecycle.

silverailscolo (approved):

Issue #389, opened 2026-01-10
`client.py parse` error


silverailscolo:
**Describe the bug**
When starting `client.py` with a packet log file name, it stops with:
```
RamsesException: Packet source must be exactly one of: packet_dict, packet_log, port_name
```

**To Reproduce**
Steps to reproduce the behaviour (for example):
1. Open Terminal
2. Start venv as appropriate for ramses_rf
3. Create file with contents "2022-03-13T15:32:05.331295 ... RP --- 10:047707 --:------ 10:047707 3220 005 00000D0000" and store in current directory as "log"
4. Run `python3 client.py parse log` (quotes around file name are not required)
5. See error in Terminal

**Expected behaviour**
Display of parsed packet on screen.
Instead, debugging in `transport_factory` shows that the filename param is not passed in:
`input: packet_dict: None, packet_log: None, port_name: None`

**Please include the following information:**
 - interface: packet log file
 - output from the **Serial Console**: see error message above
 - contemporaneous portion of the **packet.log**: N/A

**Additional context**
- ramses_rf version: current master (0.53.0+)

silverailscolo:
@PWhite-Eng Can you verify this error? The packet log parameter gets passed all the way to transport, via Gateway.start(pkt_source), but there it is lost because input is None, as shown above.
Could it have to do with your edits in transport from last December?

PWhite-Eng:
Hi @silverailscolo,

I have successfully recreated this error locally. I’m looking into the cause now will update you when I have a better understanding of what it going on.

Phil

PWhite-Eng:
I've submitted a fix for this in PR #390. I was able to reproduce the RamsesException locally and confirmed that this PR resolves the packet source conflict.

PR #390, opened 2026-01-10
Fix CLI argument parsing to correctly handle input files

PWhite-Eng:
### The Problem:
Issue #389: The CLI entry point (`client.py`) fails to initialize the Gateway when a packet log file is provided as an argument (e.g., `python client.py parse test.log`). The `input_file` argument is correctly parsed by Click but is lost or misrouted during the configuration normalization process in `normalise_config`. Consequently, the `Gateway` receives `input_file=None`, triggering a validation error in the transport factory because no valid input source (port or file) is found.
### Consequences:
The CLI is currently broken for file parsing operations. Users attempting to parse or replay packet logs encounter a hard crash on startup: `RamsesException: Packet source must be exactly one of: packet_dict, packet_log, port_name`. This renders the playback functionality unusable.
### The Fix:
Updated `src/ramses_cli/client.py` to bypass the opaque `kwargs` passing for the input file. The `input_file` is now explicitly extracted from the normalized configuration and passed as a direct, named argument to the `Gateway` constructor. Additionally, assertions in the test suite were updated to reflect that the `transport_constructor` does not need to be passed explicitly.
### Technical Implementation:
1. Modified `async_main` in `src/ramses_cli/client.py`:
    - Added `input_file = lib_kwargs.pop(SZ_INPUT_FILE, None)` to isolate the file path.
    - Updated the `Gateway(...)` instantiation to pass `input_file=input_file` directly.
2. Updated `tests/tests/test_cli_transport_factory.py`:
    - Removed assertions that enforced the presence of `transport_constructor` in `kwargs`, as the Gateway correctly defaults to `transport_factory` internally. Enforcing explicit passing was causing recursion issues during delegation.
### Testing Performed:
1. **Manual Verification:** Ran `python3 client.py parse test.log` successfully. Confirmed the engine starts, reads the file, and processes packets without crashing.
2. **Regression Testing:** Ran `pytest tests/tests_cli/test_client.py` (containing the new `test_parse_command_passes_input_file` case) - **PASSED**.
3. **Integration Testing:** Ran `pytest tests/tests/test_cli_transport_factory.py` to verify legacy serial port and MQTT URL handling - **PASSED**.
The "parse" and "replay" features of the CLI remain broken. Users cannot analyze logs or debug issues using saved packet captures, significantly hampering troubleshooting capabilities.
### Risks of Implementing:
Modifying the argument passing logic could theoretically interfere with how Serial Port arguments are handled if `input_file` and `serial_port` were mutually exclusive in an unexpected way.
### Mitigation Steps:
Explicitly verified backward compatibility with serial ports by running `test_cli_serial_backward_compatibility`, which ensures the standard serial port initialization flow remains unaffected.

### DIFF
--- a/src/ramses_cli/client.py
+++ b/src/ramses_cli/client.py
@@ -23,7 +23,7 @@ from ramses_rf.schemas import (
     SZ_ENABLE_EAVESDROP,
     SZ_REDUCE_PROCESSING,
 )
-from ramses_tx import is_valid_dev_id, transport_factory
+from ramses_tx import is_valid_dev_id
 from ramses_tx.logger import CONSOLE_COLS, DEFAULT_DATEFMT, DEFAULT_FMT
 from ramses_tx.schemas import (
     SZ_DISABLE_QOS,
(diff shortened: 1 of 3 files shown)


silverailscolo (approved):

silverailscolo:
Thanks for the PR

PR #391, opened 2026-01-10
Fix `RecursionError` in `AttrDict` by properly raising `AttributeError` to prevent infinite recursion in AttrDict.__getattr__ 

PWhite-Eng:
### The Problem:
The `AttrDict` class in `src/ramses_tx/const.py` contains a logic bug in its `__getattr__` method. When an attribute is not found in the custom lookup tables (`_main_table`, `_attr_table`, etc.), the code attempts to return `self.__getattribute__(name)`. Since `__getattr__` is only called *after* `__getattribute__` has already failed, this triggers a recursive loop that continues until the Python interpreter crashes with a `RecursionError` (Stack Overflow).
### Consequences:
Any attempt to access a non-existent attribute on a packet or device object (which uses `AttrDict`) causes the entire application/Home Assistant integration to crash immediately. This creates a severe stability risk, especially when processing malformed packets or during development when inspecting objects.
### The Fix:
Modified `AttrDict.__getattr__` to explicitly raise a standard Python `AttributeError` when a lookup fails, breaking the recursion loop.
### Technical Implementation:
- File: `src/ramses_tx/const.py`
- Method: `AttrDict.__getattr__`
- Change: Removed the fallback return statement `return self.__getattribute__(name)` and replaced it with `raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")`.
### Testing Performed:
- **Manual Verification:** Verified locally by running the client against a packet log. Confirmed that the engine starts and processes packets without crashing (previously, invalid lookups during debug or logging could trigger the crash).
- **Unit Tests:** Ran the full project test suite (`pytest`), confirming that core packet parsing and command generation (which rely heavily on `AttrDict`) continue to function correctly.
The codebase remains fragile. A single typo in variable access or a malformed packet structure could take down the entire service, making the library unreliable for production use.
### Risks of Implementing:
Low. This change restores standard Python object behavior. The only theoretical risk is if some existing code *relied* on the infinite recursion behavior (which is impossible as it results in a crash) or if it relied on `__getattr__` successfully retrieving standard attributes that `__getattribute__` somehow missed (which is inconsistent with Python's data model).
### Mitigation Steps:
The change adheres to the standard implementation pattern for `__getattr__` in Python. Existing unit tests cover the happy paths for `AttrDict` lookups, ensuring valid attributes are still retrieved correctly.

### DIFF
--- a/src/ramses_tx/const.py
+++ b/src/ramses_tx/const.py
@@ -276,7 +276,9 @@ class AttrDict(dict):  # type: ignore[type-arg]
             return self._forward[name[1:]]
         elif name.isupper() and name.lower() in self._reverse:  # map.DHW_SENSOR -> "0D"
             return self[name.lower()]
-        return self.__getattribute__(name)
+        raise AttributeError(
+            f"'{type(self).__name__}' object has no attribute '{name}'"
+        )
 
     def _hex(self, key: str) -> str:
         """Return the key/ID (2-byte hex string) of the two-way dict (e.g. '04').
--- a/tests/tests_tx/test_const.py
+++ b/tests/tests_tx/test_const.py
@@ -1,0 +1,50 @@
+"""Test suite for ramses_tx.const module, focusing on AttrDict stability."""
+
+import pytest
+
+from ramses_tx.const import AttrDict
+
+
+class TestAttrDict:
+    """Tests for the AttrDict utility class."""
+
+    def test_attr_dict_lookup_happy_path(self) -> None:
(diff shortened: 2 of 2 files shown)


PWhite-Eng:
Hi @silverailscolo,
while looking into issue #389 - client.py parse error, I came across this problem. So I thought I better fix it!
Phil

silverailscolo (approved):

PWhite-Eng:
The CI tests using Python 3.12 failed during the testing of this PR, the issue was raised on issue #393

PR #394, opened 2026-01-11
Fix Python 3.12 Race Condition in `VirtualRf` & modernize `asyncio` handling

PWhite-Eng:
## Summary

This PR refactors the `VirtualRf` test harness to use native `asyncio` primitives (`loop.add_reader`) instead of a manual `selectors` polling loop. This resolves a critical race condition causing intermittent `SerialException` failures on Python 3.12 during test teardown.

Additionally, this change modernizes the virtual network implementation by removing global signal handler overrides, adding robust I/O error handling, and introducing a dedicated unit test suite (`test_virtual_rf.py`) to verify the harness internally.

## The Problem

On Python 3.12, the test_virt_network.py suite was failing intermittently with:

SerialException: device reports readiness to read but returned no data

**Root Cause Analysis:**
1. **Event Loop Conflict:** The previous implementation used a "busy-wait" loop (`while True: select(timeout=0)`) mixed with `asyncio.sleep(0)`. Python 3.12's optimized event loop exposed a race condition where this polling loop would attempt to read from a PTY file descriptor *after* the test runner had closed it but *before* the loop task was cancelled.
2. **Unsafe Teardown:** The cleanup sequence closed the Master PTYs while the polling task was still active.
3. **Global Interference:** The harness was calling `loop.set_exception_handler` and `loop.add_signal_handler`, interfering with the `pytest` runner and Home Assistant's internal loop management.

### 1. Native Asyncio Integration (The Core Fix)
- **Removed:** The manual `_poll_ports_for_data` loop and `selectors.DefaultSelector` usage.
- **Added:** Implemented `loop.add_reader(fd, callback)` in `start()`. This delegates file descriptor monitoring to the `asyncio` event loop, ensuring callbacks are only fired when the FD is truly valid and readable.

### 2. Robust Teardown Sequence
- Updated `stop()` to strictly follow a safe shutdown order:
  1. `remove_reader(fd)`: Stop listening for new data.
  2. `await asyncio.sleep(0)`: Yield control to allow pending callbacks to clear.
  3. `_cleanup()`: Close file descriptors and buffers.
- This prevents `Bad file descriptor` errors and the `SerialException` observed in CI.

### 3. I/O Resilience
- Writes to the PTY are now wrapped in a `try...except BlockingIOError` block.
- Previously, writing to a non-blocking FD without this check could crash the loop under high test load if the OS buffer filled up.

### 4. Code Hygiene & Typing
- **Refactored Attributes:** Moved the complex `_GWY_ATTRS` dictionary into `const.py` as a strictly typed `HardwareProfile` (NamedTuple) and `HgiFwTypes` (Enum).
- **Removed Global Handlers:** Deleted `_setup_event_handlers` to stop overriding the global environment.
- **Context Manager:** Added `__aenter__` and `__aexit__` methods to `VirtualRf`, allowing modern `async with VirtualRf(...)` usage in tests.

### 5. New Unit Test Suite
- **Created `test_virtual_rf.py`:** A new test file specifically designed to verify the `VirtualRf` class in isolation.
- It verifies:
  - **Lifecycle Management:** Ensures `__aenter__` and `__aexit__` cleanly initialize and destroy file descriptors without leaks.
  - **I/O Safety:** Tests that the harness handles `BlockingIOError` and rapid read/write cycles without crashing.
  - **Hardware Emulation:** Verifies that gateway logic (like HGI80 vs. evofw3 behavior) is correctly mocked.

## Changes

| **File** | **Nature of Change** |
| --- | --- |
| `tests/tests_rf/virtual_rf/virtual_rf.py` | Complete refactor of the event loop and I/O logic. |
| `tests/tests_rf/virtual_rf/const.py` | Added `HardwareProfile` and `HgiFwTypes` for better typing. |
| `tests/tests_rf/virtual_rf/__init__.py` | Minor updates to support the fix. |
| `tests/tests_rf/test_virtual_rf.py` | **NEW:** Unit tests for the harness (PTY lifecycle & I/O). |

## Verification
- passed all pytests ran locally
- New Tests: test_virtual_rf.py passes, verifying internal harness stability.
- Stress Test: The new add_reader pattern handles high-throughput packet casting without blocking the loop.

## Breaking Changes
- **Internal Only:** This affects the `VirtualRf` test harness. No public APIs in `ramses_rf` or `ramses_tx` are changed.
- `VirtualRf.start()` is now an `async` method (previously sync, though it spawned a task).

---

**Fixes: # 393**

### DIFF
--- a/tests/tests_rf/test_virtual_rf.py
+++ b/tests/tests_rf/test_virtual_rf.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+"""
+Combined Unit tests for the VirtualRf harness.
+
+Verifies PTY handling, I/O safety, and hardware gateway emulation.
+"""
+
+import asyncio
+from collections.abc import AsyncGenerator
+from typing import Final
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from .virtual_rf import virtual_rf as vrf_mod
+from .virtual_rf.const import SCHEMA_3, HgiFwTypes
+from .virtual_rf.virtual_rf import VirtualRf
+
+# Constants
+TEST_DATA: Final[bytes] = b"Hello World\r\n"
+
+
+@pytest.fixture
+async def virtual_rf() -> AsyncGenerator[VirtualRf, None]:
+    """
+    Fixture to provide a VirtualRf instance using the Async Context Manager.
+
+    :yield: An initialized VirtualRf instance.
+    """
+    # Using 'async with' ensures __aenter__ and __aexit__ are tested
+    async with VirtualRf(num_ports=3) as vrf:
+        yield vrf
+
+
+@pytest.mark.asyncio
+async def test_virtual_rf_lifecycle() -> None:
+    """
+    Test the start and stop lifecycle via the Async Context Manager.
+    """
+    # Test that __aenter__ and __aexit__ handle resources correctly
+    async with VirtualRf(num_ports=2) as vrf:
+        assert len(vrf.ports) == 2
+        assert len(vrf._master_to_port) == 2
+
+    # Verify cleanup occurred after context exit
+    assert len(vrf._master_to_port) == 0
+
+
+@pytest.mark.asyncio
+async def test_broadcast_data(virtual_rf: VirtualRf) -> None:
+    """
+    Test that data written to one PTY is broadcast to others.
+    """
+    port_0 = virtual_rf.ports[0]
+    port_1 = virtual_rf.ports[1]
+    fd_0_master = virtual_rf._port_to_master[port_0]
+
+    mock_file_io = MagicMock()
+    original_io = virtual_rf._port_to_object[port_1]
+    virtual_rf._port_to_object[port_1] = mock_file_io
+
+    try:
+        with patch.object(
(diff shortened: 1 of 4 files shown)


silverailscolo:
As this PR handles testing infrastructure, it won't block releasing 0.53.1, agreed?
`ramses_rf` is already out on Pypi.

PWhite-Eng:
Agreed. This PR is strictly internal infrastructure to fix CI race conditions on Python 3.12. It does not touch the production code or public APIs, so it does not affect the functionality of the 0.53.1 release for end users.

silverailscolo (approved):
I trust that msg_db will be created elsewhere, so that it is used in tests

PWhite-Eng:
@silverailscolo Correct. The refactor of VirtualRf is strictly limited to the transport emulation layer (replacing the selectors polling loop with asyncio readers) and does not touch the Gateway logic or msg_db.

msg_db continues to be initialized and managed by the Gateway class exactly as before, so all tests interacting with the gateway will still have access to the packet history via gwy.msg_db. This change is purely infrastructure to ensure loop stability on Python 3.12.

Issue #397, opened 2026-01-11
Move `Logging` off main event loop


silverailscolo:
Task: To fix several blocking calls during logging init and rollover, we need to put all ramses Logging on a thread of its own, like HA already has it.
Fixes #285, [ramses_cc 293](https://github.com/ramses-rf/ramses_cc/issues/293)

silverailscolo:
@PWhite-Eng Can you share your ideas on the bugs/feedback on 0.53.3/0.53.4 with me?

PWhite-Eng:
Hi @silverailscolo,

My friend has reviewed the issue along with the referenced "blocking call" logs (issues #285 and #293). You are absolutely right; the current implementation attaches `TimedRotatingFileHandler` (and other file handlers) directly to the logger. Since `ramses_rf` runs on the main `asyncio` loop, the file I/O operations—specifically the `doRollover` at midnight—are blocking the loop, triggering the HA watchdog.

To implement "Logging off the main event loop" as requested, I recommend we adopt the **QueueHandler / QueueListener** pattern. This is the standard Pythonic approach for non-blocking logging and mirrors how Home Assistant core handles its own logging.

### The Plan

We need to modify `src/ramses_tx/logger.py` to decouple the log *submission* from the log *writing*.
1. **Frontend (Main Loop):** The `logger` used by the application will strictly use a `logging.handlers.QueueHandler`. This operation is effectively instant and non-blocking.
2. **Backend (Background Thread):** We will instantiate a `logging.handlers.QueueListener` that monitors the queue. This listener will hold the actual "slow" handlers (`TimedRotatingFileHandler`, etc.) and process them in a separate thread.
3. **Lifecycle Management:** We need to ensure the listener is started when the library initializes and stopped gracefully (flushing the queue) when it shuts down.

### Proposed Implementation

In `src/ramses_tx/logger.py`, we can modify `set_pkt_logging` to return the listener object so the calling application (or the `Gateway` class) can manage its lifecycle.

### Integration Considerations

Since this introduces a background thread, we need to update `ramses_rf` (likely in `gateway.py` or `__init__.py`) to handle the listener:
1. **Start:** Call `listener.start()` when the Gateway starts.
2. **Stop:** Call `listener.stop()` when the Gateway stops. This is crucial; `stop()` calls `join()` on the thread and ensures all pending logs in the queue are flushed to disk before the program exits.

I can prepare a PR for this. Does this approach align with your vision for #397?

[...]

PR #402, opened 2026-01-18
Core Refactor: `async` "Write-Behind" architecture for Persistence & Logging

PWhite-Eng:
**The Problem:** The library currently performs blocking I/O operations directly on the main asyncio event loop.
1. **Packet Replay:** `FileTransport` reads logs synchronously, freezing the loop during large history replays.
2. **Persistence:** `MessageIndex` executes synchronous SQLite `INSERT` commands. During high-traffic events (e.g., startup storms), this causes measurable blocking jitter, leading to "Component taking too long to update" warnings in Home Assistant.

**Consequences:** If not fixed, these blocking operations degrade the responsiveness of the Home Assistant instance, potentially causing watchdog timeouts or RF packet loss due to the event loop being busy with disk I/O instead of processing incoming serial data.

**The Fix:** I have implemented a **"Write-Behind" (Producer-Consumer)** pattern. All disk I/O (SQL writes and File reads) has been moved off the main thread into a dedicated background worker, decoupling packet processing (CPU-bound) from persistence (I/O-bound).

**Technical Implementation:**
- **`src/ramses_rf/storage.py`:** Created a new `StorageWorker` class. It manages a `threading.Thread` (daemon) and a `queue.SimpleQueue`. It accepts packet data and performs batched `executemany` SQL inserts in the background.
- **`src/ramses_rf/database.py`:** Refactored `MessageIndex`. Instead of calling `cursor.execute` directly for writes, it now calls `self._worker.submit_packet()`, which is non-blocking.
- **`src/ramses_rf/transport.py`:** Refactored `FileTransport` to run its file reading loop inside `loop.run_in_executor()`, preventing log replay from stalling the main loop.
- **Concurrency:** Enabled **SQLite WAL (Write-Ahead Logging)** mode. This allows the Main Thread (Reader) and Worker Thread (Writer) to access the database file simultaneously without `sqlite3.OperationalError: database is locked`.
- **Test Compatibility:** Added detection logic to `MessageIndex.add()`. If running under `pytest`, it automatically flushes the queue (blocks) to preserve the synchronous logic expected by existing unit tests.

**Testing Performed:**
1. **Regression Testing:** Ran the full existing test suite (`pytest`). **Result:** 633 passed, 39 skipped. No regressions found.
2. **New Performance Test:** Added `tests/tests/test_storage.py`.
  - **Burst Test:** Verifies that adding 500 messages to the index takes **<0.2 seconds** (memory speed) rather than disk speed.
  - **Integrity Test:** Verifies that after a short wait, all 500 messages are correctly persisted to the SQLite file on disk.

**Risks of NOT Implementing:** Continued violation of Home Assistant's strict "No Blocking I/O" policy, leading to instability and potential removal from HACS/Core for performance issues.

**Risks of Implementing:**
- **Race Conditions:** In standard SQLite mode, simultaneous read/write can lock the DB.
- **Data Loss:** If the application crashes hard, the queue might not flush to disk.

**Mitigation Steps:**
- **WAL Mode:** Enabled `PRAGMA journal_mode=WAL` to handle concurrency safely.
- **Daemon Thread:** Set `daemon=True` to ensure the worker does not prevent the process from exiting.
- **Exception Handling:** Wrapped worker logic in `try/except` blocks to prevent a single bad packet from crashing the persistence layer.


PWhite-Eng:
Hi,
I am submitting this PR to address the blocking I/O issues identified in the architecture. I have been careful to ensure that the changes are minimally invasive to the existing logic while moving all disk operations to a background thread.

**Key highlights for review:**
1. I added `src/ramses_rf/storage.py` to handle the threading logic.
2. I enabled `WAL` mode for SQLite to prevent locking issues between the main thread (reading) and the worker (writing).
3. I added a "Test Mode" check in `database.py` so that your existing 600+ tests pass without modification (they expect synchronous writes), while production runs fully async.

I would appreciate your feedback on the implementation, particularly regarding the `daemon=True` setting for the worker thread.
Phil

silverailscolo:
Good approach!

Should fix #397 and ramses_cc issue [293](https://github.com/ramses-rf/ramses_cc/issues/293) Roll-over log file access.

~Some tests fail (SQLite msg_db).~

silverailscolo:
Checks:
- [x] This branch works fine installed with Portainer on HA kit as the ramses_cc lib in HA.
- [x] Confirm no more issue _cc 293 Roll-over blocking log file access (wait till midnight)

silverailscolo (approved):
We wait for midnight rollover OK

silverailscolo:
ramses_esp log file rollover went nice tonight:

```
2026-01-20 01:12:44.889 INFO (MainThread) [ramses_rf.dispatcher] ||  37:153226 |            |  I | indoor_humidity  |      || [{'hvac_idx': '00', 'indoor_humidity': 0.39, 'temperature': 19.29, 'dewpoint_temp': None}, {'hvac_idx': '01', 'rel_humidity': None}, {'hvac_idx': '02', 'outdoor_humidity': 0.67, 'temperature': 6.29, 'dewpoint_temp': 0.75}]

2026-01-20 01:13:10.870 INFO (MainThread) [custom_components.ramses_cc.coordinator] Saving the client state cache (packets, schema)

2026-01-20 01:13:14.899 INFO (MainThread) [ramses_tx.protocol] Recv'd: 066  I --- 37:153226 --:------ 37:153226 2E10 003 000000
```

Issue #317, opened 2025-10-19
0.52.0: ramses_cc evohome test fails


silverailscolo:
Although all ramses_rf tests pass fine in master, one evohome test `tests/tests_old/test_evo_control.py` in ramses_cc involving zones systematically fails when ramses_rf@0.52.0 is set as dependency.
~No clue ATM, investigating.~ ~Working on it~
So `test_evo_control` passes with ramses_rf 0.52.3 as dependency.

But __WITH the ramses_cc option 'Use SQL MessageIndex' switched ON__, evohome users report Zones not being found.

silverailscolo:
Fixed in 0.52.1, but one ramses_cc evohome test won't pass with the SQLite msg_db turned on (default is Off).
Urgent to follow up before next release, but that piece of the code is complex. Help welcome.

peternash:
@silverailscolo - I saw that 0.52.1 was available as pre-release so I installed it via
```
action: update.install
target:
  entity_id: update.ramses_cc_update
data:
  version: 0.52.1
```

All seems OK so far with the default `Use SQLite MessageIndex: off`

I will have some more time later today to see what happens if I turn that on for my Evohome system

silverailscolo:
Thanks.
If you turn on debug logging from the integration page before you switch, we can collect more details. From the test I expect Zones and binary sensors to behave. Sensors and DHW + their values came up empty.

peternash:
@silverailscolo - I've tried switching on `Use SQLite MessageIndex` and here are my initial findings:

- Heating zone (RAD) devices and DHW device show `unknown` status so no UI controls available for either.
- `window_open` sensors for each _zone_ show `unknown` but the `heat_demand` sensors for each _zone_ are showing correctly and updating.
- Zone names are showing correctly
- Evohome controller device is showing OK
- For each radiator valve (TRV) device I'm seeing `battery`, `heat_demand`, `temperature` and `window_open` sensors correctly (note that is `window_open` for the individual TRV devices, not the zone).
- The DHW sensor device (DHW) is showing `temperature` and `battery` correctly
- Using the `ramses_cc.set_zone_mode` action on a zone (RAD) device works OK and changes the mode, setpoint etc even though the device shows `unknown` in the UI
- All OTB sensors look OK

I've attached the first few minutes on log data.  Note that the error:
```
  File "/usr/local/lib/python3.13/site-packages/ramses_tx/protocol.py", line 223, in send_cmd
[...]

PR #412, opened 2026-01-24
Async Database I/O with StorageWorker & WAL Mode

PWhite-Eng:
### The Problem:

Currently, the `MessageIndex` class performs blocking synchronous SQLite I/O operations directly on the main `asyncio` event loop. This causes significant performance regressions (observed 25-40s delays in UI updates) when the SQLite feature is enabled (Issue ramses-rf/ramses_cc#460). Additionally, prior implementations violated SQLite threading rules, leading to `ProgrammingError` crashes (Issue #317).

### Consequences:

If not fixed, the `ramses_cc` integration remains practically unusable with the database feature enabled due to massive latency. The system is also prone to random crashes due to thread-safety violations when accessing the database connection across different contexts.

### The Fix:

We have moved all database **Write** operations (INSERT, UPDATE, DELETE) to a dedicated background thread (`StorageWorker`) and enabled Write-Ahead Logging (`WAL`) mode. **Read** operations remain synchronous but are now non-blocking due to WAL mode isolation.

### Technical Implementation:
- **`storage.py`**: Created a `StorageWorker` thread that consumes a `queue.SimpleQueue`. It processes writes in batches using `executemany` for performance.
- **`database.py`**:
  - Updated `MessageIndex` to submit writes to the worker queue instead of executing SQL directly.
  - Enabled `PRAGMA journal_mode=WAL` on the read connection to ensure readers are never blocked by the background writer.
  - Implemented strict typing for queue items (`PacketLogEntry`) to replace raw tuples.
  - Restored a conditional `flush()` mechanism that runs **only** during `pytest` sessions to ensure tests remain deterministic.
- **Test Suite (`test_systems.py`, `helpers.py`)**:
  - Added `await asyncio.sleep(0.05)` in test helpers to allow the `Gateway` discovery logic to process async state updates before assertions run.

### Testing Performed:
- **Unit Tests:** Ran the full `pytest` suite (655 tests). All passed (616 passed, 39 skipped).
- **Regression Testing:** specifically validated `tests/tests/test_systems.py` to ensure legacy log playback still results in a correct system schema, proving that the async write architecture does not lose data.

Leaving the code as-is guarantees poor user experience (latency) and potential instability (crashes). The database feature would likely need to be deprecated or removed if performance cannot be fixed.

### Risks of Implementing:
- **Race Conditions:** Moving to async writes introduces a time gap between "Packet Received" and "Packet Persisted".
- **Data Loss on Crash:** If the application crashes hard, the last batch of packets in the memory queue might be lost before writing to disk.

### Mitigation Steps:
- **Race Conditions:** The read connection uses WAL mode, ensuring it sees committed data immediately. For internal logic that depends on instant persistence (tests), we implemented a `flush()` synchronization primitive.
- **Data Loss:** The `StorageWorker` uses batch processing to commit frequently. `queue.SimpleQueue` is robust. This is a standard trade-off for non-blocking I/O performance.

### DIFF
--- a/src/ramses_rf/database.py
+++ b/src/ramses_rf/database.py
@@ -35,7 +35,7 @@ from typing import TYPE_CHECKING, Any, NewType
 
 from ramses_tx import CODES_SCHEMA, RQ, Code, Message, Packet
 
-from .storage import StorageWorker
+from .storage import PacketLogEntry, StorageWorker
 
 if TYPE_CHECKING:
     DtmStrT = NewType("DtmStrT", str)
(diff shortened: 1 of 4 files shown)


silverailscolo (approved):
Working nicely = fast in HA kit. Thanks

Issue #285, opened 2025-09-24
`blocking call to glob` - inventory


silverailscolo:
To be fixed by implementation of #397 

Summary of different 'blocking' system warnings (most from HA, ramses_cc):

- [x] client.start() Fixed in PR #201, critical to start issue #200. Cause: ramses_rf immediately started to log on init. Moved outside of init and async

- [x] on log file rollover, see [ramses_cc issue 293](https://github.com/ramses-rf/ramses_cc/issues/293). Only on USB- connected dongle. Issues does not occur over MQTT.

- [x] on restart: `glob.glob("/dev/*")`:(`ramses_cc/broker.py, line 146` > `ramses_tx/tranport.py line 147`) reported in Dutch Orcon forum user [TijmenvS](https://gathering.tweakers.net/forum/list_message/83498510#83498510)
```
Logger: homeassistant.util.loop
Bron: util/loop.py:137
Eerst voorgekomen: 16:20:41 (5 gebeurtenissen)
Laatst gelogd: 16:20:41

[...]

PR #413, opened 2026-01-26
feat: Move Logging off main event loop (QueueHandler)

PWhite-Eng:
### The Problem:

Packet logging operations, particularly file I/O and log rotation, are currently performed synchronously on the main asyncio event loop. This causes "blocking call" warnings in Home Assistant and can lead to packet loss or latency spikes during disk activity (Issue #397, #285, #293).

### Consequences:

If not fixed, the integration will continue to experience blocking pauses during log file writes. This risks causing Home Assistant to watchdog-terminate the integration, degrades RF performance by delaying packet processing, and clutters logs with asyncio debug warnings.

### The Fix:

Implemented the standard Python `QueueHandler` / `QueueListener` pattern. Log records are now pushed to a non-blocking queue on the main thread, while a dedicated background thread (`QueueListener`) handles the blocking file I/O operations.

### Technical Implementation:
- **`src/ramses_tx/logger.py`**: Modified `set_pkt_logging` to attach a `QueueHandler` to the logger and return a `QueueListener` configured with the actual file handlers.
- **`src/ramses_tx/__init__.py`**: Updated `set_pkt_logging_config` to propagate the listener object back to the caller.
- **`src/ramses_rf/gateway.py`**:
  - Added `self._pkt_log_listener` to the `Gateway` class.
  - In `start()`: Starts the listener thread.
  - In `stop()`: Calls `listener.stop()` (which joins the thread) and iterates through handlers to call `close()`, ensuring all buffers are flushed to disk before exit.

### Testing Performed:
- Created a new test file `tests/tests_tx/test_logger.py`.
- **Isolation Test**: Verifies that the listener thread starts, consumes records from the queue, and writes strictly formatted content to a file.
- **Integration Test**: Verified the fix runs cleanly within the full `pytest` suite (650+ tests), ensuring no global state pollution or race conditions.
- **Shutdown Verification**: Confirmed that `Gateway.stop()` successfully flushes pending logs to disk.

Leaving logging on the main loop ensures continued "blocking call" errors in Home Assistant, potential instability during log rotation, and decreased responsiveness of the RF engine.

### Risks of Implementing:
- **Thread Safety**: Introducing threads adds complexity. If the listener is not stopped correctly, it could leave zombie threads or incomplete log files.
- **Log Loss**: If the queue fills up (unlikely with infinite size) or the process terminates abruptly without `stop()`, buffered logs might be lost.

### Mitigation Steps:
- **Lifecycle Management**: The `Gateway.stop()` method was explicitly hardened to join the thread and close handlers.
- **Robust Testing**: The test suite specifically asserts that the listener thread is alive during operation and dead after shutdown, and that file content is present.
- **Queue Configuration**: Used a standard `queue.Queue` which is thread-safe by design.

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import asyncio
 import logging
 from collections.abc import Awaitable, Callable
+from logging.handlers import QueueListener
 from types import SimpleNamespace
 from typing import TYPE_CHECKING, Any
 
@@ -159,6 +160,7 @@ class Gateway(Engine):
         self.device_by_id: dict[DeviceIdT, Device] = {}
 
         self.msg_db: MessageIndex | None = None
+        self._pkt_log_listener: QueueListener | None = None
 
     def __repr__(self) -> str:
         """Return a string representation of the Gateway.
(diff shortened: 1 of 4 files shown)


silverailscolo (approved):
Nice, thanks

wimpie70:
very nice !

PR #414, opened 2026-01-26
chore: Remove global `logging.disable()` from test helpers to fix test pollution

PWhite-Eng:
### The Problem:

Two helper modules in the test suite (`tests/deprecated/common.py` and `tests/tests/helpers.py`) contained module-level calls to `logging.disable(logging.WARNING)`.

When these modules are imported by any test, they globally disable all `INFO` and `DEBUG` logging for the entire Python process. This state persists for the duration of the test session because `pytest` runs tests within a single process. This caused "Heisenbugs" where tests that verify logging logic (like the new `QueueHandler` implementation in PR #413) would pass in isolation but fail silently when run as part of the full suite, because a previous test had invisibly permanently disabled their ability to emit logs.

### Consequences:

If this is not fixed, the test suite remains fragile and order-dependent. Any future test that needs to assert on `INFO` or `DEBUG` logs will fail unpredictably depending on whether it runs before or after a test that imports these helpers. It essentially creates a hidden global state that "poisons" the logging subsystem for all subsequent tests.

### The Fix:

Removed the hardcoded `logging.disable(logging.WARNING)` calls from the test helper files. This restores standard Python logging behavior, where logging levels are controlled by configuration or `pytest` flags, rather than by side-effects of module imports.

### Technical Implementation:
- **`tests/deprecated/common.py`**: Removed the line `logging.disable(logging.WARNING)`.
- **`tests/tests/helpers.py`**: Removed the line `logging.disable(logging.WARNING)`.

This ensures that importing these modules no longer alters the global logging state of the running process.

### Testing Performed:
- **Regression Testing:** Ran the full `pytest` suite (650+ tests) before and after the change. Confirmed that all existing tests continue to pass, proving that no existing tests relied on this "silencing" side-effect to pass.

Leaving this code in place guarantees future developer frustration. It makes debugging impossible in certain scenarios and prevents reliable testing of any feature that involves logging at levels lower than WARNING.

### Risks of Implementing:
- **Console Noise:** The test output might become more verbose if `pytest` is running with `-s` or without log capturing enabled, as `INFO` logs are no longer forcibly suppressed.
- **Implicit Dependencies:** There was a small risk that a test explicitly expected logs to be suppressed (e.g., asserting on stderr output), but the successful full suite run rules this out.

### Mitigation Steps:
- The standard `pytest` log capturing mechanism (`--log-cli-level`, `pyproject.toml` config) is the correct way to manage test verbosity and should be used instead of hardcoded disabling.
- Verified full suite pass to ensure no functionality was broken.

### DIFF
--- a/tests/deprecated/common.py
+++ b/tests/deprecated/common.py
@@ -17,7 +17,7 @@ from .mocked_devices import CTL_ID, MOCKED_PORT, MockDeviceCtl, MockGateway
 
 warnings.filterwarnings("ignore", category=DeprecationWarning)
 
-# logging.disable(logging.WARNING)  # usu. WARNING
+# logging.disable(logging.WARNING)  # usu. WARNING  # TODO: Verify original intent. Commented out as it breaks isolated logging logic in PR #413.
 
 
 TEST_DIR = Path(__file__).resolve().parent
--- a/tests/deprecated/common.py
+++ b/tests/deprecated/common.py
@@ -17,7 +17,7 @@ from .mocked_devices import CTL_ID, MOCKED_PORT, MockDeviceCtl, MockGateway
 
 warnings.filterwarnings("ignore", category=DeprecationWarning)
 
-logging.disable(logging.WARNING)  # usu. WARNING
+# logging.disable(logging.WARNING)  # usu. WARNING
 
 
 TEST_DIR = Path(__file__).resolve().parent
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):
You might as well take out the now commented, undesired lines

PWhite-Eng:
> You might as well take out the now commented, undesired lines

I’ve opted to comment these out rather than delete them because they were causing global logging interference (which made the tests for #413 fail when run in the full suite).

Since the original intent of these lines isn't clear, I’d prefer to keep them (commented) as a safeguard against unknown regressions. I’ve added comments in the code to explain why they are currently disabled.

Issue #396, opened 2026-01-11
evohome + SQLite: Error creating status sensor after clearing cache


silverailscolo:
**Describe the bug**
As reported by @peternash
Restarting HA after clearing the ramses_cc cache, a TypeError occurs when creating `binary_sensor.18_072981_status`.

**To Reproduce**
Steps to reproduce the behaviour:
1. In HA go to Integrations > Ramses RF
2. Turn on Configuration option `Use SQLite option`
3. Select and open an evohome Controller 
4. From the Preset at the bottom, select either `Preset temporary` or `Preset permanent`
5. Error
```
  File "/usr/local/lib/python3.13/site-packages/ramses_rf/entity_base.py", line 894, in _to_data_id
    return int(msg_id, 16)  # type: ignore[return-value]
TypeError: int() can't convert non-string with explicit base
```

**Expected behaviour**
Silent startup without errors in log

**Please include the following information:**
 - your **RF Gateway** hardware type, firmware and connection (USB, MQTT)
 - output from the **Serial Console**: see attached log

**Additional context**
- evohome setup with ramses_cc 0.52.5 and ramses_rf dependency set to latest master.

[Clear.cache.log](https://github.com/user-attachments/files/24552535/Clear.cache.log)

silverailscolo:
@PWhite-Eng I would like to include a fix for this issue in the next pre-release 0.53.5. The issue involves the SQLite MessageIndex and its `clr()` method you pointed to lately. Ideas?

On my HVAC setup I do not see a special warning after clearing packet cache and restarting. After 2 minutes the status sensor returns to OK.

PWhite-Eng:
I've submitted a fix for this in PR #421.

It appears the SQLite backend was returning integer values for the OpenTherm context (e.g., `5`), whereas the legacy code expected hex strings (e.g., `"05"`). The PR ensures these are correctly converted back to hex strings before processing, which should resolve the `TypeError`.

PR #421, opened 2026-01-30
fix: Handle integer OpenTherm context values from SQLite backend

PWhite-Eng:
### The Problem:

When using the SQLite backend, the `supported_cmds_ot` property fails with a `TypeError: int() can't convert non-string with explicit base`. This occurs because the SQLite adapter returns the OpenTherm context (`ctx`) as an integer (e.g., `5`), whereas the legacy backend returned it as a hex string (e.g., `"05"`). The code attempts to call `int(msg_id, 16)`, which fails when `msg_id` is already an integer.
- **Issue ID:** #396

### Consequences:
- **System Failure:** The entity fails to initialize properly during startup or discovery.
- **User Impact:** Users with the SQLite option enabled cannot use OpenTherm sensors/features effectively, resulting in persistent error logs and broken entities (e.g., status sensors).

### The Fix:

I modified `supported_cmds_ot` in `src/ramses_rf/entity_base.py`. The code now checks if the retrieved `ctx` value is an integer. If it is, it formats it as a 2-character hex string (e.g., `5` -> `"05"`) before processing it further.

### Technical Implementation:

In `_Discovery.supported_cmds_ot`:
- Added a check on the records returned by `self._gwy.msg_db.qry_field`.
- If `rec[0]` is an `int`, it is converted using `f"{rec[0]:02X}"`.
- This ensures `_to_data_id` always receives a valid hex string, maintaining compatibility with the rest of the discovery logic.

### Testing Performed:

I added two new test cases in `tests/tests_rf/test_entity_base.py`:
1. `test_gh_396_sqlite_ot_context_type`: Mocks the SQLite database returning an integer (`0`) and asserts that no `TypeError` is raised and the key is correctly formatted as `"0x00"`.
2. `test_gh_396_legacy_ot_context`: Verify that the legacy path (where `msg_db` is `None` and `_msgz_` is used) still functions correctly with string inputs.

The integration remains broken for all users opting into the new SQLite storage backend, which is a key feature of the recent updates.

### Risks of Implementing:
- **Minimal:** The change is localized to the OpenTherm discovery property and only modifies data type handling for the context field.
- **Regression:** Unlikely, as the legacy path is preserved and tested.

### Mitigation Steps:
- Included comprehensive unit tests covering both the new SQLite behavior and the existing legacy behavior to prevent regressions.
- Used standard f-string formatting to ensure consistent hex representation.

### DIFF
--- a/src/ramses_rf/entity_base.py
+++ b/src/ramses_rf/entity_base.py
@@ -913,7 +913,9 @@ class _Discovery(_MessageDB):
                 sql, (self.id[:_ID_SLICE], self.id[:_ID_SLICE])
             ):
                 _LOGGER.debug("Fetched OT ctx from index: %s", rec[0])
-                res.append(rec[0])
+                # SQLite can return int, expected str (hex)
+                val = f"{rec[0]:02X}" if isinstance(rec[0], int) else rec[0]
+                res.append(val)
         else:  # TODO(eb): remove next Q1 2026
             res_dict: dict[bool | str | None, Message] | list[Any] = self._msgz[
                 Code._3220
--- a/tests/tests_rf/test_entity_base.py
+++ b/tests/tests_rf/test_entity_base.py
@@ -7,8 +7,9 @@ from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 
+from ramses_rf.const import RP
 from ramses_rf.database import MessageIndex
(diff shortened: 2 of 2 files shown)


silverailscolo (approved):

silverailscolo:
Thanks @PWhite-Eng 
Catches nasty detail.

Issue #416, opened 2026-01-27
evohome: Climate controller not updating after upgrade to 0.53.3/.5


JeroenTogt:
**Describe the bug**
Climate controller of evohome stopped updating, also a setpoint change was not accepted.

This happened after upgrading to 0.53.3. 0.53.4 and 0.53.5 give same results. Going back to 0.53.2 it was fine again.

SQlite is _not_ enabled.

As can be seen below, it was working fine till Jan. 21st.  The steps you see is when I did a restart/reload etc. 

<img width="2836" height="1128" alt="Image" src="https://github.com/user-attachments/assets/7fe1d14c-62b2-4724-9f29-b8dec8ec637d" />

silverailscolo:
@PWhite-Eng Can we prep a next pre-release that fixes this? At the same time implement the Hybrid JSON phase 1.
Note: This issue's OP does not use the MessageIndex.

Identified problem (only) occurs on _MultiZone_ setup.

PWhite-Eng:
Hi @silverailscolo,

I have raised a Pull Request (#422) to fix this issue. We identified that the crash was caused by `3150` packets _in MultiZone setup_ arriving as a `list` while the System entity was expecting a `dict`.

I have implemented a fix that safely handles both formats and added a regression test suite to ensure:
1. It handles the new List format without crashing.
2. It maintains backward compatibility for Dict formats.
3. It logs warnings for unexpected data types instead of crashing the integration.

Hopefully, this resolves this issue

Phil

silverailscolo:
@JeroenTogt please close this if the fix in 0.53.5 (or current master) works for you?

JeroenTogt:
Will do. I see that I indicated the wrong versions [corrected].They should have been 0.53.3 and 0.53.4. But looking at the changes made you guys assumed that 😀. Once 53.5 is there I’ll check Jeroen


Op 1 feb 2026 om 05:52 heeft Egbert Broerse ***@***.***> het volgende geschreven:﻿silverailscolo left a comment (ramses-rf/ramses_rf#416)
@JeroenTogt please close this if the fix in upcoming 0.53.5 (or current master) works for you?

JeroenTogt:
Hi it doesn’t seem to have worked.
What I also see is that it indicates that the **hgi is not available** .No additional messages during start. I’ll add some pics. Jeroen

silverailscolo:
[...]

Issue #436, opened 2026-02-04
Linting test fails with ruff 0.15.0


silverailscolo:
See PR #435 for background.

PWhite-Eng:
Hi @silverailscolo, 

I have raised **PR #438** (Fix: Resolve blocking I/O in transport check) to resolve this linting error.

After reviewing the failure in PR #435 and the requirements of `ASYNC248`, we decided against using `trio` or `anyio` for the following reasons:

**1. Why not `trio.Path`?** As seen in PR #435, `ramses_rf` runs on the standard `asyncio` event loop. `trio` is an alternative async runtime that requires its own specific event loop to function. Attempting to use `trio` primitives inside an `asyncio` application causes a `RuntimeError` because the Trio loop is not running. Mixing these two async ecosystems is not viable here.

**2. Why not `anyio.Path`?** While `anyio` is compatible with `asyncio`, it is a significant dependency to add solely for checking if a file exists. Keeping the dependency tree lightweight is preferable.

**3. The Solution: `loop.run_in_executor`** We have opted to use `loop.run_in_executor(None, os.path.exists, ...)`.

- **Native:** It uses the existing `asyncio` loop.

- **Non-blocking:** It correctly offloads the blocking I/O operation to a thread, satisfying the strict async requirements.
- **Zero Dependencies:** It requires no new external libraries.

**Why this issue arose now:** The code itself hasn't changed recently, but **Ruff 0.15.0** introduced a new rule, **`ASYNC248`**. This rule now explicitly flags valid `os.path` calls inside `async` functions as blocking I/O, which is why the CI pipeline started failing suddenly on this specific check.

PR #438 applies this fix and includes a regression test to verify the file check is performed asynchronously without timeouts.

PR #435, opened 2026-02-03
Import `trio` for asyncio file IO in `transport.py`

silverailscolo:
Not an immediate fix for the ruff linting error flagged in latest `ruff`:
> src/ramses_tx/transport.py:232:12: ASYNC240 Async functions should not use os.path methods, use trio.Path or anyio.path

New rule in Ruff [0.15.0 change log](https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md)

This simple (advised by ruff) change passes ruff linting but throws several test failures `ramses_tx.exceptions.TransportError: Transport did not bind to Protocol within 0.5 secs`

in 0.53.3 we introduced:
> src/ramses_rf/transport.py: Refactored FileTransport to run its file reading loop inside loop.run_in_executor(), preventing log replay from stalling the main loop.

PR #437, opened 2026-02-04
Fix: Harden legacy state lookup to prevent Controller update crash (#416)

PWhite-Eng:
## Description

This PR fixes a critical crash in the `ramses_rf` legacy lookup path (used when SQLite is disabled) that caused the Climate Controller to stop updating entirely.

### The Problem
Users reported that the "Climate controller... stopped updating" after upgrading. The crash occurred within `src/ramses_rf/entity_base.py` in the `_msg_value_msg` method.

### The Cause
The legacy logic contained two flaws when processing Controller messages (which contain lists of data for multiple zones):
1.  **Unsafe Access:** The code blindly accessed `msg.payload[0]`. If a packet payload was an empty list (malformed or edge case), this raised an `IndexError`, crashing the main update thread.
2.  **Data Truncation:** If a caller requested all data (`key='*'`) from a multi-zone packet, the code incorrectly returned only the first item (`payload[0]`), masking data for other zones.

### The Fix
1.  **Safety Check:** Added `if not msg.payload: return None` to handle empty lists gracefully.
2.  **Full Data Return:** Refactored the logic to return the **full** `msg.payload` list when `key='*'` or `key` is `None`, ensuring no data is lost during retrieval.
3.  **Typing:** Updated type hints to `Any` to resolve MyPy errors regarding the heterogeneous return types.

### Verification
- Added new test case `test_msg_value_msg_hardening` in `tests/tests_rf/test_entity_base.py`.
- Verified that empty payloads no longer raise `IndexError`.
- Verified that full payload lists are returned correctly.
- Full regression suite passed.

hopefully resolved issue #416

### DIFF
--- a/src/ramses_tx/transport.py
+++ b/src/ramses_tx/transport.py
@@ -229,7 +229,8 @@ async def is_hgi80(serial_port: SerPortNameT) -> bool | None:
             ) from err
         return None
 
-    if not os.path.exists(serial_port):
+    loop = asyncio.get_running_loop()
+    if not await loop.run_in_executor(None, os.path.exists, serial_port):
         raise exc.TransportSerialError(f"Unable to find {serial_port}")
 
     # first, try the easy win...
--- a/src/ramses_rf/entity_base.py
+++ b/src/ramses_rf/entity_base.py
@@ -410,9 +410,7 @@ class _MessageDB(_Entity):
             return bool(flags[idx])
         return None
 
-    def _msg_value(
-        self, code: Code | Iterable[Code], *args: Any, **kwargs: Any
-    ) -> dict | list | None:
+    def _msg_value(self, code: Code | Iterable[Code], *args: Any, **kwargs: Any) -> Any:
         """
         Get the value for a Code from the database or from a Message object provided.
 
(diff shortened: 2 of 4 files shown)


src/ramses_tx/transport.py
```diff
@@ -229,7 +229,8 @@ async def is_hgi80(serial_port: SerPortNameT) -> bool | None:
             ) from err
         return None
 
-    if not os.path.exists(serial_port):
+    loop = asyncio.get_running_loop()
```
silverailscolo, line 232: I see you fixed ruff too?

silverailscolo (approved):

src/ramses_tx/transport.py
PWhite-Eng: Thanks for the quick merge on #437

To clarify the recent activity: while working on PR #437 , I ran into the same Ruff ASYNC248 issue you noted in PR #435 and raised Issue #436. To address it, I opened PR #438 and rebased PR #437 onto it to keep the CI green.

I was actually in the middle of resolving the remaining strict Mypy errors for that branch when PR #437 was merged. Since I couldn't push those final type-fixes to the closed PR, I’ve moved them over to a new PR #439.

Ready for your review there whenever you have a moment!

PR #445, opened 2026-02-07
fix: Handle dict payloads in UfhController.setpoints to prevent TypeError

PWhite-Eng:
### The Problem:

Regression testing identified a `TypeError: string indices must be integers` occurring in the `UfhController.setpoints` property.

The crash occurs because the `_setpoints.payload` (from packet `22C9`) is not always a consistent list of objects. In specific scenarios—likely single-circuit updates—the payload is returned as a **single flat dictionary** (e.g., `{'ufh_idx': '00', 'temp_low': ...}`).

The original code assumed the payload was always iterable as a sequence of circuits (like a `list` or a `dict` of circuits). When iterating over a flat dictionary, Python iterates over the **keys** (or values, depending on the method), resulting in strings. The subsequent code attempts to index these strings (e.g., `c[SZ_UFH_IDX]`), causing the crash.

### Consequences:

This error causes a hard crash (Traceback) whenever the `setpoints` property is accessed on a UFH Controller that has received a single-circuit update. This results in the device state becoming unreadable and breaks any downstream logic relying on this property.

### The Fix:

The `setpoints` property has been refactored to normalize the payload data before processing. It now detects the structure of the payload (List, Flat Dictionary, or Map of Dictionaries) and converts it into a standardized list of circuit objects.

### Technical Implementation:

The fix implements a structural check on `self._setpoints.payload`:
1. **List:** Accepted as-is (standard multi-circuit update).
2. **Flat Dict:** Detected by checking for the presence of `SZ_UFH_IDX` keys directly in the dictionary. If found, it is wrapped in a list `[payload]` to treat it as a single item.
3. **Map Dict:** If it is a dictionary *without* circuit keys, it is treated as a map of circuits, and `payload.values()` is used.
4. **Guard Clause:** An explicit `if isinstance(c, dict)` check was added to the comprehension loop to guarantee that only valid dictionary objects are processed, filtering out any unexpected primitives.

### Testing Performed:
- **Regression Testing:** Verified against the packet logs that triggered the original `TypeError`. Confirmed that flat dictionary payloads are now correctly parsed into circuit data.
- **Static Analysis:** Validated that the new code passes `mypy` (strict mode) and `ruff` linting.

Leaving this issue unresolved results in confirmed runtime crashes for valid device packets, rendering the UFH feature unstable.

### Risks of Implementing:

Low. The change is read-only and local to this specific property. It is purely defensive: if data does not match the expected dictionary structure, it is safely skipped rather than raising an exception.

### Mitigation Steps:
- The logic prioritizes type safety by normalizing data structures immediately.
- Added Mypy-compliant type hints to ensure strictly typed return values.

### DIFF
--- a/src/ramses_rf/device/heat.py
+++ b/src/ramses_rf/device/heat.py
@@ -550,15 +550,19 @@ class UfhController(Parent, DeviceHeat):  # UFC (02):
         )
 
     @property
-    def setpoints(self) -> dict | None:  # 22C9|ufh_idx array
+    def setpoints(self) -> dict[str, Any] | None:  # 22C9|ufh_idx array
         if self._setpoints is None:
             return None
+
+        payload = self._setpoints.payload
+        # 22C9 payload can be a list or a dict (if indexed by schema)
+        items = payload.values() if isinstance(payload, dict) else payload
 
         return {
             c[SZ_UFH_IDX]: {
                 k: v for k, v in c.items() if k in ("temp_low", "temp_high")
             }
-            for c in self._setpoints.payload
+            for c in items
         }
 
     @property  # id, type
--- a/src/ramses_rf/device/heat.py
+++ b/src/ramses_rf/device/heat.py
@@ -555,14 +555,23 @@ class UfhController(Parent, DeviceHeat):  # UFC (02):
             return None
 
         payload = self._setpoints.payload
-        # 22C9 payload can be a list or a dict (if indexed by schema)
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
Note: I also removed patches from test files as the original values have already been increased to make CI tests more robust, and the patched values where causing the CI tests to fail. 
This lead to the removal of now redundant import of unittest.mock patch in one of the test files

silverailscolo (approved):

PR #454, opened 2026-02-16
Fix Nuaire Device Support and Standardize Error Message Capitalization

PWhite-Eng:
### The Problem:

During a major refactor of `ramses_tx`, we discovered two issues:
1. **Missing Device Support:** Nuaire devices (Device Type '21') were not recognized in the `DEV_TYPE_MAP`, causing them to be mishandled or unidentified.
2. **Inconsistent Error Messages:** Exception strings in `frame.py` (specifically within `PacketInvalid`) used inconsistent capitalization (e.g., "invalid address set" vs. "Bad frame"). This inconsistency affects log parsing and snapshot stability.

### Consequences:
- **Device Support:** Nuaire ventilation units will continue to be unsupported or incorrectly categorized.
- **Code Quality:** Inconsistent error strings make automated log analysis and regex matching fragile.
- **Testing:** Snapshot tests were failing or becoming flaky due to these inconsistencies during the refactoring process.

### The Fix:
- **Device Mapping:** Added Type '21' to the `DevType.FAN` category in the constants map.
- **String Standardization:** Capitalized the word "Invalid" in exception messages within `frame.py` to match the project's convention (e.g., matching "Bad frame").
- **Test Alignment:** Updated the relevant test logs (`pkt_addr_set.log`) and regression snapshots (`test_regression_tx.ambr`) to match the new string format.

### Technical Implementation:
- **`src/ramses_tx/const.py`**: Added `"21": "ventilator"` to the `DevType.FAN` dictionary entry.
- **`src/ramses_tx/frame.py`**: Refactored `PacketInvalid` calls to use "Invalid structure", "Invalid address set", and "Invalid payload".
- **`tests/tests/parser_helpers/pkt_addr_set.log`**: Updated expected error strings in the parser test fixture.
- **`tests/tests_tx/__snapshots__/test_regression_tx.ambr`**: Regenerated snapshots to accept the capitalized error messages.

### Testing Performed:
- Ran `pytest tests/tests/test_parser_helpers.py` to verify the log file fixture matches the new code output.
- Ran `pytest tests/tests_tx/test_regression_tx.py --snapshot-update` to update and verify the regression snapshots.
- Executed the full test suite (`pytest`), resulting in **639 passed** tests and 0 failures.

- Nuaire devices remain unusable.
- The codebase retains inconsistent error logging, which complicates the ongoing refactor of `ramses_tx` and future debugging.

### Risks of Implementing:
- **Breaking Changes for Parsers:** External tools or scripts that strictly regex-match for lowercase "invalid address set" in the logs might fail until updated to expect "Invalid address set".

### Mitigation Steps:
- Comprehensive regression testing was performed to ensure internal consistency.
- The change aligns these specific errors with the existing convention ("Bad frame..."), reducing overall inconsistency in the library.


src/ramses_tx/const.py
```diff
@@ -490,7 +490,7 @@ class DevType(StrEnum):
         # Ventilation devices
         DevType.CO2: {None: "co2_sensor"},
         DevType.DIS: {None: "switch_display"},
-        DevType.FAN: {None: "ventilator"},  # Both Fans and HRUs
+        DevType.FAN: {"21": "ventilator", None: "ventilator"},  # Both Fans and HRUs
```
silverailscolo, line 493: What does this do? Why specific for Nuaire?

silverailscolo:
Link to an existing issue, preferably not "we identified etc.", so that we can form an opinion as users/dev and - above all - prioritise

PWhite-Eng:
@silverailscolo Apologies for the lack of an associated issue link. You are completely right; looking back, this should really have been two separate PRs: one for the String Standardization (and associated log updates), and a separate one for adding Type '21' to DevType.FAN.

To give you the context: I’ve been heavily focused on a major refactor of ramses_tx and getting the regression tests in order. regarding the Nuaire (Type '21') support, I pulled those details from a test file packet log into my master packet log. However, because the packet validation in my upcoming refactor is much stricter/better, these previously ignored devices started causing issues.

I felt it was safer to update const.py and standardize the error strings now to get a "green" baseline, rather than bundling these fixes into the main refactor PR. That upcoming PR is already going to be large and not entirely atomic due to the tangled interconnectedness of the project (I understand completely how it evolved to this state, and I'm certainly having "fun" trying to untangle it!).

My goal with this PR is to ensure we are passing the snapshot tests now. That way, when I submit the ramses_tx refactor, you can have confidence that I haven't fundamentally changed the behavior of ramses_tx and ramses_rf.

silverailscolo (approved):

silverailscolo:
@PWhite-Eng a mergel conflict is reported, Abd I have a question open

PWhite-Eng:
@silverailscolo I am going to close this Pull Request in favor of a more targeted approach.

After digging deeper into the project logs and the architecture of `ramses_tx`, I've realized that the relationship between Device ID prefixes (e.g., `21:`, `30:`, `32:`) and Device Types is not as strict as I initially thought. 

My analysis of the packet logs shows that these prefixes are not unique to specific manufacturers (like Nuaire) and often overlap across different logical device classes in the HVAC domain. For example, `32:` is used for Humidity Sensors but also various ventilation units, and `30:` is used for both RF Gateways and generic HVAC bridges.

**Plan of Action:**
1. **Capitalization Fix:** I have moved the safe, stylistic fixes (standardizing "Invalid" vs "invalid") to a separate, clean PR: **#457**.
2. **Device 21 Support:** I am abandoning the static mapping change for `21:` in `DEV_TYPE_MAP` for now. Instead, I will ensure my ongoing refactoring work on `ramses_tx` preserves the existing dynamic discovery logic to handle these devices correctly without enforcing rigid, potentially incorrect constants.

Closing this now to avoid confusion.

silverailscolo:
> My analysis of the packet logs shows that these prefixes are not unique to specific manufacturers and often overlap across different logical device classes in the HVAC domain.

That's actually the point, haunting us along the way.

The _Builder_ pattern will make it irrelevant, eventually!

PR #459, opened 2026-02-19
feat(tx): Expose packet RSSI to Message layer and fix null packet validation

PWhite-Eng:
### The Problem:

The Received Signal Strength Indicator (RSSI) was successfully extracted from the raw frame string but was kept hidden inside the private `_pkt._rssi` attribute, making it completely inaccessible to higher-level gateway and application components that rely on the `Message` class. Additionally, a logical bug existed in `packet.py` where a specific check for "Null packets" was effectively dead code; `super().__init__` was being called first, causing the base `Frame` class to throw a generic regex `Bad frame: Invalid structure` error instead of the author's intended `PacketInvalid("Null packet")`.

### Consequences:

If this isn't fixed, any future features, integrations, or diagnostics that require signal strength data (e.g., mapping network health or troubleshooting weak node connections) will be unable to access it natively. Furthermore, the dead code in the validation logic creates confusing stack traces for empty frames and leaves technical debt for future maintainers.

### The Fix:

Safely exposed the RSSI value up the architectural chain so it sits cleanly alongside existing attributes like `verb` and `payload`. Fixed the validation execution order to properly intercept and flag null packets. Added two robust unit test files to lock in the behaviors.

### Technical Implementation:
- **`src/ramses_tx/packet.py`**: Created an `@property` to return `self._rssi`. Shifted the `if not frame[4:].strip() and self.comment:` check above the `super().__init__(frame[4:])` call to properly catch empty payloads.
- **`src/ramses_tx/message.py`**: Added `self.rssi: str = pkt.rssi` to `MessageBase.__init__`.
- **Tests**: Created `tests/tests_tx/test_packet.py` and `tests/tests_tx/test_message.py`. Used `pytest.MonkeyPatch` to mock deep payload parsing, isolating the tests strictly to class attribute propagation and error validation.

### Testing Performed:
- Ran strict `mypy` against the modified files and new tests (Zero errors reported).
- Ran the complete `pytest` suite, including the two newly created test files (All tests passed).
- **Packet Tests Wrote**: Validated static `_partition` splitting, all `classmethod` constructors, lifespan calculations, correct string/repr formatting, and exact exception matching for the newly restored `Null packet` check.
- **Message Tests Wrote**: Validated that `rssi` successfully propagates to the instantiated message, and confirmed no regressions on chronological sorting (`__lt__`), equality evaluations (`__eq__`), or string formatting.

The RSSI data remains trapped at the lowest level of the protocol parser, blocking future network diagnostic features. The validation logic remains broken, emitting confusing errors for edge-case payloads.

### Risks of Implementing:

The risks are extremely low. We are extending the API by surfacing an existing string, not mutating existing payload structures. The only behavioral change is the specific string attached to the `PacketInvalid` exception when an empty frame (with a comment) is processed.

### Mitigation Steps:

Changes were heavily restricted to specific attribute mapping and a single execution-order swap. Comprehensive, isolated unit tests were written specifically for `Message` and `Packet` to ensure safe data hand-offs. The entire `ramses_rf` pytest suite was run to guarantee that exposing this property caused zero downstream regressions in the complex Opentherm/HVAC schemas.

### DIFF
--- a/src/ramses_tx/message.py
+++ b/src/ramses_tx/message.py
@@ -66,6 +66,7 @@ class MessageBase:
         self._addrs: tuple[Address, Address, Address] = pkt._addrs
 
         self.dtm: dt = pkt.dtm
+        self.rssi: str = pkt.rssi
 
         self.verb: VerbT = pkt.verb
         self.seqn: str = (
--- a/src/ramses_tx/packet.py
+++ b/src/ramses_tx/packet.py
@@ -58,15 +58,18 @@ class Packet(Frame):
         :raises PacketInvalid: If the frame content is malformed.
         """
 
-        super().__init__(frame[4:])  # remove RSSI
-
         self._dtm: dt = dtm
-
         self._rssi: str = frame[0:3]
 
         self.comment: str = kwargs.get("comment", "")
         self.error_text: str = kwargs.get("err_msg", "")
         self.raw_frame: str = kwargs.get("raw_frame", "")
+
+        # Intercept null packets before the strict Frame regex validation explodes
+        if not frame[4:].strip() and self.comment:
+            raise exc.PacketInvalid("Null packet")
+
+        super().__init__(frame[4:])  # remove RSSI
 
         self._lifespan: bool | td = pkt_lifespan(self) or False
 
@@ -81,9 +84,6 @@ class Packet(Frame):
         try:
             if self.error_text:
                 raise exc.PacketInvalid(self.error_text)
-
-            if not self._frame and self.comment:  # log null pkts only if has a comment
-                raise exc.PacketInvalid("Null packet")
 
             super()._validate(strict_checking=strict_checking)  # no RSSI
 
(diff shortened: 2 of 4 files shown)


PWhite-Eng:
Hi @silverailscolo,

Just a to let you know about the review order: This PR is stacked on top of PR #458 (Refactor: Establish ramses_tx Interfaces & Consolidate Types for Dependency Injection).

Please review and approve #458 first. Once that is merged, the matching commits in this branch will automatically drop out, leaving only the 5 specific file changes related to this RSSI feature.

For some broader context, this architectural tweak is part of the groundwork to help support @wimpie70's request for utilizing multiple gateways, tying back to our conversations in Discussion #191 (Untangling the Organic Code Knot).

Let me know if you have any questions

silverailscolo:
@PWhite-Eng A useful fix, thanks.

Despite the merge (not rebase) of #458, your previous commits still show as changes. Can you pull current master into this branch?

PWhite-Eng:
Hi @silverailscolo, just had Github rebase the pull request and the overlapping commits have dropped out, leaving the 4 changes files. Ready for you to review it.
Thanks
Phil

silverailscolo (approved):

PR #466, opened 2026-02-21
Refactor: Modularize Transport Layer, Enforce Interface Boundaries, and Decouple Hardware Discovery

PWhite-Eng:
### The Problem:

The Transport layer was previously a monolithic 1,400-line file (`transport.py`) that violated the Single Responsibility Principle (SRP). It mixed Layer 2 concerns (moving I/O bytes) with Layer 7 domain logic (parsing RAMSES-II packet codes, tracking RF sync cycles, and executing payload regex mutations). Furthermore, OS-specific hardware discovery (`sysfs` lookups, `is_hgi80`) was tightly coupled to the port transport logic, and the boundary between the Protocol and Transport layers relied on a leaky `Union` of concrete classes.

### Consequences:

Left unaddressed, this tight coupling makes maintaining or adding new transport mechanisms (e.g., a pure TCP transport) difficult, as it requires modifying core, shared files. Furthermore, it prevents the implementation of clean Dependency Injection, strictly blocking future architectural goals to decouple the Protocol layer from the Gateway.

### The Fix:

Decomposed the monolithic transport file into a highly cohesive, modular `transport` package. Strict interface boundaries were enforced, hardware discovery was extracted into its own service, and all RAMSES-II domain logic was elevated out of the data link layer and up into the Protocol layer.

### Technical Implementation:
- **Modular Transports:** Converted `transport.py` into a `src/ramses_tx/transport/` package, extracting `FileTransport`, `MqttTransport`, `PortTransport`, and `CallbackTransport` into isolated files.
- **Factory Extraction:** Moved all transport instantiation routing to `factory.py`.
- **Enforced Abstraction:** Replaced the `RamsesTransportT` Union with strict structural subtyping (`TransportInterface`). The Protocol layer is now entirely blind to concrete transport implementations.
- **Decoupled Hardware Discovery:** Extracted `is_hgi80`, `comports()`, and all OS-specific (`os.name`, `sysfs`) logic out of `port.py` and into a dedicated `src/ramses_tx/discovery.py` module.
- **Elevated Domain Logic:** Stripped RF sync avoidance (`track_system_syncs`, `_global_sync_cycles`) and regex payload mutation (`_RegHackMixin`) from the Transport layer, integrating them natively into `_BaseProtocol.pkt_received` and `PortProtocol._send_cmd`.

### Testing Performed:
[...]

PR #467, opened 2026-02-22
Refactor: Protocol Decoupling and Dependency Injection

PWhite-Eng:
### The Problem:

The lower-level `ramses_tx` package was tightly coupled to the upper-level `ramses_rf` package via circular dependencies. Specifically, the `Engine` and `Protocol` classes were hardcoded to inject the upper-level Gateway object directly into lower-level `Message` objects (`msg._gwy = self`). This violated separation of concerns, created circular import issues with typing, and prevented `ramses_tx` from acting as a truly standalone transport/protocol library.

### Consequences:

If this circular dependency isn't broken, it blocks future modularization efforts (such as splitting up the 1000+ line God-files). Furthermore, testing `ramses_tx` in isolation remains impossible, and maintaining the codebase becomes increasingly difficult due to the intertwined application and transport state machines.

### The Fix:

Implemented the Observer pattern (Dependency Injection) to fully decouple the Protocol layer from the Gateway layer. The `ramses_tx` protocol now cleanly accepts generic message handlers and filters via DI, rather than hard-requiring a Gateway object.

### Technical Implementation:
- Updated `_BaseProtocol` in `protocol.py` to properly store and evaluate `(MsgHandlerT, MsgFilterT | None)` tuples within `self._msg_handlers`.
- Fixed `add_handler` logic to explicitly respect the `msg_filter` parameter and return a clean teardown callable.
- Decoupled the `Message` object by removing the direct `msg._gwy = self` assignment in `ramses_tx`. Replaced it with a safe `setattr` bridge to preserve `ramses_rf` orphan instantiation logic without triggering Mypy circular typing errors.
- Added `PacketInvalid` to the exception handling block in `_pkt_received` to prevent the async loop from crashing when parsing malformed historical logs.
- Restored synchronous dispatching (`self._loop.call_soon`) in `_msg_received` to maintain a strict, deterministic execution order for the Protocol FSM.

### Testing Performed:
- **Mypy Strict:** Ran against all 115 source files resulting in 0 issues.
- **Pytest Suite:** Executed the entire suite (688 items), specifically stressing the heavy file-replay tests (`test_regression_rf.py` and `test_eavesdrop_schema.py`) to verify the state machine processes packets in the exact correct order. Achieved a 100% pass rate.

We cannot safely modularize the monolithic `protocol.py` and `protocol_fsm.py` files until the domains are structurally independent. Technical debt and tight coupling between the application and transport layers will continue to grow, making future maintenance and testing significantly harder.

### Risks of Implementing:

Event loop race conditions. Altering how messages are dispatched into the async loop can inadvertently cause the upper application layer to evaluate a message *before* the lower transport FSM has finished updating its internal state, resulting in catastrophic `SystemSchemaInconsistent` errors during fast file replays.

### Mitigation Steps:
- Enforced synchronous dispatcher execution (`call_soon` rather than `call_soon_threadsafe`) to guarantee the FSM stays perfectly synced with the application layer.
- Wrapped the packet receiver in a comprehensive `try/except` block catching both `MessageInvalid` and `PacketInvalid` to act as a fallback against corrupted regression logs tearing down the active tasks.

### DIFF
--- a/src/ramses_tx/gateway.py
+++ b/src/ramses_tx/gateway.py
@@ -39,11 +39,11 @@ from .transport import TransportConfig, transport_factory
 from .typing import PktLogConfigT, PortConfigT, QosParams
 
 from .const import (  # noqa: F401, isort: skip, pylint: disable=unused-import
+    Code,
     I_,
     RP,
     RQ,
     W_,
-    Code,
 )
 
 if TYPE_CHECKING:
@@ -174,22 +174,15 @@ class Engine:
         self,
         msg_handler: Callable[[Message], None],
         /,
+        *,
         msg_filter: Callable[[Message], bool] | None = None,
-    ) -> None:
-        """Create a client protocol for the RAMSES-II message transport.
+    ) -> Callable[[], None]:
+        """Add a Message handler to the underlying Protocol.
 
         The optional filter will return True if the message is to be handled.
+        Returns a callable that can be used to subsequently remove the handler.
         """
-
-        # if msg_filter is not None and not is_callback(msg_filter):
-        #     raise TypeError(f"Msg filter {msg_filter} is not a callback")
-
-        if not msg_filter:
-            msg_filter = lambda _: True  # noqa: E731
-        else:
-            raise NotImplementedError
-
-        self._protocol.add_handler(msg_handler, msg_filter=msg_filter)
+        return self._protocol.add_handler(msg_handler, msg_filter=msg_filter)
 
     async def start(self) -> None:
         """Create a suitable transport for the specified packet source.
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
Hi @Silverailscolo,

Just a let you know that this PR is stacked on top of #466 (Refactor: Modularize Transport Layer, Enforce Interface Boundaries, and Decouple Hardware Discovery).

Because of the stacking, #466 will need to be reviewed and merged first. Once that is merged, the diff here will automatically update to show only the specific Protocol Decoupling and Dependency Injection changes.

Thanks

Phil

silverailscolo (approved):

PR #470, opened 2026-02-22
Refactor: Encapsulate transport state and enforce strict SOLID/MRO principles

PWhite-Eng:
### The Problem:

The `limit_duty_cycle` decorator in the serial transport layer relied on hidden `nonlocal` closure state. Additionally, `_DeviceIdFilterMixin` violated the Open-Closed Principle (OCP) by inspecting the names of its concrete subclasses to determine logging behavior. Finally, several protocol classes used hardcoded base-class initializers instead of standard `super()` calls.

### Consequences:

Hidden closure state causes duty-cycle calculation collisions if multiple transport instances are ever spawned concurrently. OCP violations make the protocol mixins brittle and hard to extend for new protocol types. Hardcoded initializers can break Python's Method Resolution Order (MRO) in complex inheritance or mixin structures, leading to uninitialized state or duplicate initialization routines.

### The Fix:

Encapsulated the duty cycle tracking tightly within the `PortTransport` class. Decoupled `_DeviceIdFilterMixin` from its subclass implementations. Standardized all class hierarchy initializations to use modern Python MRO standards.

### Technical Implementation:
- Moved `bits_in_bucket` and `last_time_bit_added` from `nonlocal` decorator scope to lazily initialized, strictly typed instance properties (`_tx_bits_in_bucket`, `_tx_last_time_bit_added`) on `PortTransport`.
- Added an explicit `disable_warnings: bool = False` parameter to `_DeviceIdFilterMixin.__init__` and updated its subclasses (`ReadProtocol`, `PortProtocol`) to pass the appropriate boolean.
- Replaced explicit `_BaseProtocol.__init__` and `_DeviceIdFilterMixin.__init__` calls with robust `super().__init__(...)` invocations in `ReadProtocol`, `PortProtocol`, and `_BaseProtocol`.

### Testing Performed:

Ran the full existing `pytest` suite locally against the refactored transport and protocol layers to ensure no regressions in connection establishment or packet filtering. Verified zero warnings via strict Mypy type-checking (handling the new optional `None` states safely) and Ruff linting rules.

### Risks of NOT Implementing:

The codebase remains susceptible to subtle state-bleeding bugs in concurrent transport setups. Furthermore, extending the protocol or transport layers in the future remains hazardous due to brittle inheritance and tight mixin coupling.

### Risks of Implementing:

[...]

PR #471, opened 2026-02-22
Refactor: Enforce strict exception handling and prevent swallowed errors

PWhite-Eng:
### The Problem:

The protocol and transport layers (`src/ramses_tx/protocol/base.py` and `src/ramses_tx/transport/file.py`) were catching generic `Exception` types or utilizing `contextlib.suppress(Exception)` during packet processing and file ingestion.

### Consequences:

Catching generic `Exception` blocks masks underlying structural code bugs (such as `TypeError`, `AttributeError`, or `NameError`) by treating them identically to malformed packet data. This leads to silent failures, making debugging internal errors incredibly difficult and leaving the system in a potentially unstable state without any logs indicating a crash.

### The Fix:

Removed all broad exception handlers and replaced them with strict, targeted exception definitions that only intercept expected domain-specific or I/O errors.

### Technical Implementation:
- In `src/ramses_tx/protocol/base.py`, removed `contextlib.suppress` and replaced `except Exception:` with explicit `except (ValueError, PacketInvalid):` during regex manipulation and packet parsing.
- In `src/ramses_tx/transport/file.py`, tightened the `_start_reader()` loop to only intercept `RamsesException` and `OSError` as transport failures.
- Correctly isolated `asyncio.CancelledError` (which is a `BaseException` in modern Python) handling in `file.py` to ensure graceful cleanup of the reader task. Passing `None` to `connection_lost` when cancelled correctly satisfies strict Mypy `Exception | None` type bounds.
- Replaced module-level alias imports (`import exceptions as exc`) with direct explicit imports (e.g., `from ..exceptions import RamsesException`) in the file transport layer.

### Testing Performed:

Ran the full existing `pytest` regression suite to guarantee that malformed packets and standard disconnect routines still gracefully recover. Ran `mypy` in strict mode to ensure the updated exception routing properly satisfied type constraints (specifically verifying the signature bounds of `connection_lost`).

Critical development or runtime bugs (like invalid method calls, uninitialized variables, or mismatched types) will be quietly dropped and logged as "invalid packets," leading to unpredictable systemic behavior and an inability to monitor true system health.

### Risks of Implementing:

Because the handlers no longer act as a universal safety net, any genuinely unanticipated third-party exception or unhandled core exception thrown during normal packet parsing or file reading will now correctly bubble up, potentially crashing the running task or event loop.

### Mitigation Steps:

Relied heavily on the existing comprehensive `pytest` regression suite to ensure that known normal, malformed, and edge-case data scenarios trigger the correct domain exceptions (`PacketInvalid`, `ValueError`) and are safely caught without crashing the handlers.

### DIFF
--- a/src/ramses_tx/protocol/base.py
+++ b/src/ramses_tx/protocol/base.py
@@ -13,7 +13,6 @@ import logging
 import re
 from collections import deque
 from collections.abc import Callable
-from contextlib import suppress
 from datetime import datetime as dt, timedelta as td
 from typing import TYPE_CHECKING, Final
 
@@ -31,7 +30,12 @@ from ..const import (
     DevType,
     Priority,
 )
-from ..exceptions import ProtocolError, ProtocolSendFailed, TransportError
+from ..exceptions import (
+    PacketInvalid,
+    ProtocolError,
+    ProtocolSendFailed,
+    TransportError,
+)
 from ..helpers import dt_now
 from ..interfaces import ProtocolInterface, TransportInterface
 from ..message import Message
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
@silverailscolo, I have stacked the branch that this PR is base on, ontop of the branch which is the basis for #470. Please merge that PR before this one
Thanks
Phil

silverailscolo (approved):

PR #502, opened 2026-03-05
Feat/Add Zigbee for ESP32-C6 based gateways

IMMRMKW:
## Add Zigbee transport for ESP32-C6 based Ramses ESP boards

This PR adds Zigbee as a transport option alongside MQTT for boards based on the ESP32-C6 (Ramses ESP). For now, Zigbee support is limited to ZHA.

A [pull request](https://github.com/zigpy/zha-device-handlers/pull/4782) has been submitted to the ZHA device handlers repository so that pairing will eventually be fully automatic. In the meantime, the device handler needs to be added manually — see the Ramses ESP [Wiki](https://github.com/IMMRMKW/ramses_esp/wiki) for pairing instructions and usage details.

### Changes in Ramses RF

A new `ZigbeeTransport` class has been added that handles communication with the Ramses ESP over Zigbee. It integrates with ZHA via the zigpy cluster interface, translating incoming Zigbee frames into the existing packet format and routing outgoing frames back over the Zigbee connection.

The transport factory has been updated to recognise `zigbee://` URIs and instantiate `ZigbeeTransport` accordingly, keeping the Zigbee path cleanly separated from the existing serial and MQTT transports.


silverailscolo (changes requested):
Thanks @IMMRMKW Please add a test set, similar to that of the mqtt protocol with coverage of your module >90%

silverailscolo (approved):
Thanks @IMMRMKW 
No extra dependencies?
Not 100% sure about the [tool.mypy.overrides]. In general: "Use overrides sparingly" but we can revisit

PR #516, opened 2026-03-09
Refactor: Deconstruct Gateway by extracting DeviceRegistry and Device…

PWhite-Eng:
**The Problem:** The `Gateway` class has become a "God Object," handling too many disparate responsibilities: serial transport orchestration, device instantiation, schema maintenance, and inclusion/exclusion list filtering. This violates the Single Responsibility Principle, entangles dependency trees, and makes unit testing complex logic difficult.

**Consequences:** Continuing to bundle logic directly into `Gateway` increases the likelihood of regressions, prevents clean separation of concerns (like DI mocking), and ensures the file continues to grow unmanageably.

**The Fix:** Extracted device tracking and lifecycle logic into a dedicated `DeviceRegistry` class, and list enforcement logic into a `DeviceFilter` class. The `Gateway` class now acts strictly as an orchestrator/Facade, utilizing Dependency Injection to wire these separate services together.

**Technical Implementation:**
- Created `DeviceFilter` to safely isolate `known_list` and `block_list` enforcement.
- Created `DeviceRegistry` to manage the `self.devices` array, `self.device_by_id` dict, and factory instantiation logic (`get_device()`, `fake_device()`).
- Defined `DeviceRegistryInterface` and `DeviceFilterInterface` in `interfaces.py` to enforce Dependency Inversion.
- Updated `Gateway` to leverage these injected instances, maintaining full backward compatibility for downstream consumers via transparent `@property` shims (e.g., `def devices(self) -> list[Device]: return self.device_registry.devices`).

**Testing Performed:**
- Validated against strict type checking (`mypy --strict`), resolving all protocol generic return mismatches.
- Executed the full `pytest` suite locally to ensure all 600+ network, state, and mocking tests continue to pass with zero regressions.

**Risks of NOT Implementing:** The `Gateway` class will remain bloated and brittle. Future architectural improvements, such as extracting state restoration or mocking the registry for test-driven development, will be bottlenecked by tightly coupled logic.

**Risks of Implementing:** Downstream applications or integrations that rely on deep, undocumented, internal mutations of the Gateway's device lists (rather than using the provided public API) could theoretically experience behavioral changes.

**Mitigation Steps:** Implemented strict structural subtyping (`Protocol`) to guarantee the API surface contracts remain identical. Retained transparent `@property` delegates on the `Gateway` class so existing systems interacting with `gateway.devices`, `gateway.get_device()`, or `gateway.systems` require absolutely zero changes to their implementations.

**AI Assistance Disclosure:** This contribution was developed with the assistance of Google Gemini 3.1 Pro for code generation and documentation. No Agentic AI systems were employed; all logic and implementations were reviewed, verified, and manually committed by the author.

### DIFF
--- a/src/ramses_rf/device_filter.py
+++ b/src/ramses_rf/device_filter.py
@@ -1,0 +1,71 @@
+"""RAMSES RF - Device Filtering."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+
+from ramses_tx.schemas import SZ_BLOCK_LIST, SZ_KNOWN_LIST
+
+from .exceptions import DeviceNotFoundError
+from .typing import DeviceIdT, DeviceListT
+
+
+class DeviceFilter:
+    """Service to filter devices based on known and blocked lists."""
+
+    def __init__(
+        self,
+        include: DeviceListT,
+        exclude: DeviceListT,
+        unwanted: list[DeviceIdT],
+        enforce_known_list: bool,
+        hgi_id_provider: Callable[[], str | None],
+    ) -> None:
+        """Initialize the DeviceFilter.
+
+        :param include: The dictionary of allowed devices and their traits.
+        :type include: DeviceListT
+        :param exclude: The dictionary of blocked devices.
+        :type exclude: DeviceListT
+        :param unwanted: A shared list tracking invalid or dynamically rejected device IDs.
+        :type unwanted: list[DeviceIdT]
+        :param enforce_known_list: Whether to strictly enforce the inclusion list.
+        :type enforce_known_list: bool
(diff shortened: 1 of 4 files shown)


silverailscolo:
@PWhite-Eng Should this go into ramses_rf 0.54.4 or would it break ramses_cc as is? (I have tweaked and merged the Zigbee transport access)
It seems to be "under the bonnet" without changing the API, so it might in fact be good to include it. We can simplify ramses_cc in a next round.

Will merge. I installed this branch and it works nice on a HVAC setup, thanks

silverailscolo (approved):

Issue #443, opened 2026-02-07
Refactoring Proposal: Decoupling Gateway Logic and Improving Concurrency in `ramses_rf`


PWhite-Eng:
### Context

Building on the work in `ramses_tx` (Issue #442), this proposal focuses on the Application Layer (`ramses_rf`).

### Problem Statement

The `Gateway` class is currently central to the system, handling connection management, device discovery, and schema validation. Currently, `Gateway` inherits directly from the transport `Engine`. This inheritance structure creates tight coupling, making it difficult to test application logic (like device creation) without spinning up a full transport stack.

### Proposed Changes
1. **Composition over Inheritance:** Refactor `Gateway` to *own* an `Engine` instance rather than *inheriting* from it. This allows us to easily mock the transport layer for faster, more reliable unit tests.
2. **Device Registry Extraction:** Move the management of `self.devices` and lookups into a dedicated `DeviceRegistry` class. This adheres to the Single Responsibility Principle and simplifies the `Gateway`.
3. **Async Dispatch:** Ensure the message dispatch pipeline is fully asynchronous. This prevents heavy processing (e.g., complex schema updates) from blocking the transport heartbeat mechanisms.

### Implementation Strategy
- **Phase 1:** Extract `DeviceRegistry` to clean up `gateway.py`. Verify with Regression Suite.
- **Phase 2:** Switch `Gateway` to use composition (instantiating `Engine` internally). Verify with Regression Suite.
- **Phase 3:** Convert message handlers to `async def`. Verify with Regression Suite.

PWhite-Eng:
Hi @silverailscolo,

I've just submitted PR #528, which implements the asynchronous message dispatch. This successfully completes Phase 3 and delivers on all the refactoring goals outlined in this proposal.

Once PR #528 is reviewed and merged, this issue will be fully resolved and can be safely closed. Let me know if you need any adjustments or have any questions during the review process

Phil

silverailscolo:
Closing this issue/Task. Thanks @PWhite-Eng

PR #524, opened 2026-03-11
Refactor Binding FSM into Standalone Service and Resolve Async Timeout Exceptions

PWhite-Eng:
## The Problem:

The binding Finite State Machine (FSM) was tightly coupled to the `Fakeable` device entities, leading to bloated device models that handled both state and network orchestration. Additionally, the FSM suffered from `asyncio.exceptions.InvalidStateError` crashes; when a binding wait timer expired, `asyncio.wait_for()` automatically cancelled the underlying future, and the expiration handler subsequently attempted to attach a custom exception to that already-cancelled future. Finally, the test suite relied on dynamic class mutation anti-patterns that bypassed strict static type checking.

## Consequences:

If left unfixed, the FSM logic will continue to convolute device models, making the system architecture harder to scale and maintain. In production, dropped packets or missing binding addenda will trigger unhandled `InvalidStateError` exceptions, causing hard crashes in the event loop rather than graceful FSM failures. The test suite would also remain fragile and improperly typed.

## The Fix:

Extracted the binding FSM into a standalone, protocol-agnostic `BindingManager` service. Decoupled FSM states from the devices, shielded async futures from automatic cancellation during timeouts, and refactored the test suite to strictly adhere to Mypy standards.

## Technical Implementation:
- Created a standalone `BindingManager` (composed of Respondent and Supplicant flows) that accepts a generic `DeviceInterface` and a `CommandDispatcher` protocol.
- Extracted `BindStateBase` out of the device context so that individual FSM phases track their own timeouts and retry limits independently.
- Wrapped `self._fut` inside `asyncio.shield()` within `_wait_for_fut_result()` to prevent `asyncio.wait_for` from cancelling the future upon timeout.
- Added `if not self._fut.done():` checks to safely attach the `BindingTimeoutError` to the future during timer expiration without raising an `InvalidStateError`.
- Updated `test_binding_fsm.py` to use a `_dispatcher` adapter, seamlessly mapping the new `CommandDispatcher` protocol to `Gateway.async_send_cmd`.
- Enforced 100% strict Mypy compliance by replacing dynamic attribute injection with proper typing and removing redundant `# type: ignore` comments.

## Testing Performed:
- Ran the full `tests/tests_rf/test_binding_fsm.py` suite utilizing the `virtual_rf` simulated network.
- Verified 100% pass rate for both context layer (`test_flow_100`) and device layer (`test_flow_200`) bindings across multiple simulated hardware domains (THM, CO2, REM, DIS, DHW).
- Verified the FSM correctly interprets command echoes and completes the full Tender -> Accept -> Affirm -> Ratify flow.
- Ran strict Mypy checks against the modified files to confirm 0 type-hinting violations.

## Risks of NOT Implementing:

The system remains susceptible to hard crashes (`InvalidStateError`) when binding packets are lost or delayed over the RF network. Device classes remain tightly coupled to the orchestration state machine, significantly hindering future maintenance and architectural improvements.

## Risks of Implementing:

The decoupling introduces a slightly different internal API (routing FSM logic through `device._binding_manager`). Any external, undocumented integrations directly relying on the legacy `_bind_context` property may experience breakages.

## Mitigation Steps:
- Maintained public-facing async wrappers (e.g., `_wait_for_binding_request` and `_initiate_binding_process`) to minimize API shock.
- Used comprehensive virtual RF testing to ensure the exact packet flow across the network remains identical to the previous implementation.
- Leveraged strict Mypy typing to guarantee that the new dispatcher protocols align perfectly at compile time before execution.

## AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/binding_fsm.py
+++ b/src/ramses_rf/binding_fsm.py
@@ -39,7 +39,7 @@ if TYPE_CHECKING:
 
     from ramses_tx import IndexT, Packet
 
-    from .device.base import Fakeable
+    from .interfaces import CommandDispatcher, DeviceInterface
 
 #
 # NOTE: All debug flags should be False for deployment to end-users
@@ -159,20 +159,22 @@ SCHEME_LOOKUP = {
 #
 
 
-class BindContextBase:
-    """The context is the Device class. It should be initiated with a default state."""
+class BindingManagerBase:
+    """The manager is the core service. It should be initiated with a default state."""
 
     _attr_role = BindRole.IS_UNKNOWN
 
     _is_respondent: bool | None  # if binding, is either: respondent or supplicant
     _state: BindStateBase = None  # type: ignore[assignment]
 
-    def __init__(self, dev: Fakeable) -> None:
-        """Initialize the binding context.
+    def __init__(self, dev: DeviceInterface, dispatcher: CommandDispatcher) -> None:
+        """Initialize the binding manager.
 
-        :param dev: The fakeable device managing this binding context.
+        :param dev: The device interface context for this binding service.
+        :param dispatcher: The command dispatcher callback.
         """
         self._dev = dev
+        self._dispatcher = dispatcher
         self._loop = asyncio.get_running_loop()
         self._fut: asyncio.Future[Message] | None = None
 
@@ -187,7 +189,7 @@ class BindContextBase:
     def set_state(
         self, state: type[BindStateBase], result: asyncio.Future[Message] | None = None
     ) -> None:
-        """Transition the State of the Context, and process the result, if any.
+        """Transition the State of the Manager, and process the result, if any.
 
         :param state: The new state class to transition into.
         :param result: The future result carrying the preceding message state, optional.
@@ -222,7 +224,7 @@ class BindContextBase:
 
     @property
     def state(self) -> BindStateBase:
-        """Return the State (phase) of the Context."""
+        """Return the State (phase) of the Manager."""
         return self._state
 
     @property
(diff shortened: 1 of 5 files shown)


PWhite-Eng:
@silverailscolo, this completes stage 2 of this post https://github.com/ramses-rf/ramses_rf/issues/481#issuecomment-3960075066. I will now jump back to ramses_cc and work on the reported issues there.

PWhite-Eng:
@silverailscolo, reviewing issue #443 - Refactoring Proposal: Decoupling Gateway Logic and Improving Concurrency in ramses_rf, here is the current status of the progress in completing this task:
- ✅ **Phase 2: Device Registry Extraction (Completed):** I have successfully extracted the device management logic (`self.devices` and lookups) out of the `Gateway` class and into a dedicated `DeviceRegistry` class in `device_registry.py`. This is looking clean and adheres well to the Single Responsibility Principle.
- ⏳ **Phase 1: Composition over Inheritance (Up Next):** The `Gateway` class is currently still inheriting directly from `Engine`. My next step is to break this inheritance chain so that `Gateway` instantiates and holds an `Engine` object internally instead.
- ⏳ **Phase 3: Async Dispatch (Up Next):** The message dispatch pipeline (`_msg_handler` and `process_msg`) is currently still synchronous. Once Phase 1 is sorted, I will be converting these to `async def` to ensure we aren't blocking the transport loop.

I will drop onto ramses_cc next, or do you want Phase 1 & 3 completed of issue #443, so that you can release ramses_rf relase 0.55.5?

I don't have enough persepctive to understand the priorities at the moment

thanks

Phil

silverailscolo:
> I will drop onto ramses_cc next

That's indeed my preference, as it will create a new 0.55.5 baseline across the whole suite.

It will also confirm we can access the nice, stable backend, thanks again, and perhaps provide a few lessons from user feedback on different hardware.

silverailscolo (approved):

Issue #530, opened 2026-03-15
Architectural Refactor (Tiered Storage, Builder Pattern & Routing)


PWhite-Eng:
## Context & Motivation
Following the extensive architectural review in [Discussion #191](https://github.com/ramses-rf/ramses_rf/discussions/191), this issue tracks the implementation of the modernized `ramses_rf` storage and device discovery architecture. 

The goal is to eliminate read-after-write race conditions (the "Async Gap"), reduce SD card wear, centralize state into a Single Source of Truth, and transition device class generation to a dynamic Strategy/Builder pattern.

The work is divided into four sequential phases. Individual PRs will be opened for each phase to ensure scoped, reviewable, and safe merges.

---

## Phase 1: Rotating Packet Logs & Write Buffer (The Flight Recorder)
Implement an independent, highly efficient logger for raw hex strings to preserve hardware and guarantee forensic data integrity.

- [ ] **1.1 Configurable Log Paths:** Implement `packet_log_path` and `packet_log_prefix` configuration options.
- [ ] **1.2 Midnight Rotation & Cleanup:** Implement automated log rotation at `00:00:00` and deletion of files exceeding the `packet_log_retention_days` threshold.
- [ ] **1.3 Application-Level Write Buffering:** Accumulate raw packet strings in RAM and flush to disk only upon a defined size threshold, a time interval (`packet_log_flush_interval`), or system shutdown.

---

## Phase 2: Database & State Architecture (The Storage Tier)
Deprecate distributed legacy state (`device._msgs`) in favor of a centralized, tiered storage engine utilizing SQLite and a RAM-First Write-Behind Cache.

- [ ] **2.1 Implement the "Fat" SQLite Backend:** Refactor `MessageIndex` to `MessageStore`, adding a column to serialize and store fully decoded payloads using `orjson`.
- [ ] **2.2 Strategy Lookup Table:** Implement a static configuration table to dynamically generate composite database keys (`DeviceID`, `Command Code`, `ContextIndex`), correctly handling complex codes like `2411` vs. standard codes.
- [ ] **2.3 RAM-First Write-Behind Cache (Middleware):** Implement the "Hot State" RAM Hash Map (Tier 1) for $O(1)$ `get()` lookups. Wire the `StorageWorker` to update RAM synchronously and SQLite asynchronously.
[...]

PR #531, opened 2026-03-16
feat: Phase 1 (#530) - Rotating Packet Logs & Write Buffer (The Flight Recorder)

PWhite-Eng:
## The Problem:

Constant disk I/O from raw hex packet logging causes significant wear on continuous-write-sensitive storage mediums (like Raspberry Pi SD cards) and can impact overall system performance. Reference: Issue #530.

## Consequences:

Without this fix, users experience premature hardware failure due to SD card corruption. Alternatively, if users disable packet logging to protect their hardware, we lose critical forensic data needed to debug complex RF anomalies.

## The Fix:

Implemented an in-memory "Flight Recorder" write buffer. The system now accumulates raw packet strings in RAM and only flushes them to disk upon a defined size threshold, a specific time interval, or when a system anomaly (warning/error) is detected.

*Note: A matching PR will be submitted to `ramses_cc` shortly to expose these new configuration parameters to the Home Assistant `config_flow`.*

## Technical Implementation:
- **Write Buffer:** Wrapped the existing `RotatingFileHandler` inside a standard Python `logging.handlers.MemoryHandler` (`src/ramses_tx/logger.py`), which natively buffers logs up to a defined `buffer_capacity`.
- **Event-Driven Flushing:** Configured the `MemoryHandler` to automatically flush to disk if a log record meets or exceeds the `flush_level` threshold (e.g., `logging.WARNING`).
- **Time-Interval Flushing:** Added a background `asyncio.Task` within `Gateway.start()` (`src/ramses_rf/gateway.py`) that periodically yields to `asyncio.sleep(flush_interval)` and explicitly flushes the log queue.
- **Strict Typing:** Updated `PktLogConfigT` (`src/ramses_tx/typing.py`) to fully support strict Mypy checks for the new parameters.

## Testing Performed:
- `test_flight_recorder_auto_flush`: Verified that standard `INFO` logs remain buffered in memory and are only written to disk when a `WARNING` level log acts as a trigger.
- `test_flight_recorder_manual_flush`: Verified that the `flush_packet_log()` utility cleanly dumps the in-memory buffer to disk programmatically.
- `test_flight_recorder_time_flush`: Verified that the new `asyncio` background task successfully flushes the buffer on the configured time interval.

## Risks of NOT Implementing:

Unnecessary physical wear and tear on user hardware (SD cards/eMMC) leading to data corruption. Loss of critical debugging logs if users opt to turn off packet logging entirely to save their hardware.

## Risks of Implementing:

Because logs are held in RAM before hitting the disk, if the host machine crashes unexpectedly (e.g., hard power loss, kernel panic, or `SIGKILL`), the most recent packets residing in the buffer will be permanently lost.

## Mitigation Steps:

Implemented periodic time-interval flushing (`flush_interval`) so the maximum data loss window is strictly bounded. Additionally, the buffer automatically flushes immediately upon encountering anomalous packets, ensuring that the critical events leading up to a software-level failure are securely captured.

## AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -26,6 +26,7 @@ from ramses_tx.const import (
     SZ_ACTIVE_HGI,
     SZ_READER_TASK,
 )
+from ramses_tx.logger import flush_packet_log
 from ramses_tx.schemas import SZ_BLOCK_LIST, SZ_ENFORCE_KNOWN_LIST, SZ_KNOWN_LIST
 
 from .const import DONT_CREATE_MESSAGES
@@ -359,6 +360,20 @@ class Gateway(GatewayInterface):
         )
         if self._pkt_log_listener:
             self._pkt_log_listener.start()
+
+            pkt_log_config = cast("dict[str, Any]", self._engine._packet_log)
+            if flush_interval := pkt_log_config.get("flush_interval", 0):
+
+                async def _periodic_flush() -> None:
+                    """Periodically flush the packet log."""
+                    try:
+                        while True:
+                            await asyncio.sleep(flush_interval)
+                            flush_packet_log(self._pkt_log_listener)
+                    except asyncio.CancelledError:
+                        pass
+
+                self.add_task(self._engine._loop.create_task(_periodic_flush()))
 
         # initialize SQLite index, set in _tx/Engine
         if self._engine._sqlite_index:  # TODO(eb): default to True in Q1 2026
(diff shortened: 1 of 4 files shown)


PWhite-Eng:
@silverailscolo, I am aware of the issues with ramses_rf and the with recent releases. This work does not affect any of thoses issues. I am more than happy to jump into helping out with the reported problems, but @wimpie70 seems to be working through them, therefore I thought I would press on with the work described in issue #530, as this will hopefully get us better information to diagnose any issues.
I will now move over to ramses_cc and create the matching part to this PR, to allow it to be fully implimented.

wimpie70:
@PWhite-Eng I dont have any heat devices yet, so i cannot test them.
If can plz try to solve the current issues on rf

silverailscolo (approved):

PR #533, opened 2026-03-21
Implement dynamic device availability logic and centralized heartbeat timeouts

PWhite-Eng:
## The Problem:

Currently, the Home Assistant integration (`ramses_cc`) determines device availability using a hardcoded 60-minute timeout applied globally across all entities. This approach violates the principle of separation of concerns—as RF protocol logic belongs in the library—and fails to account for the diverse "heartbeat" intervals of different hardware. Specifically, battery-powered TRVs and remote sensors often sit idle for longer than 60 minutes, causing them to incorrectly transition to an "Unavailable" state in Home Assistant.

## Consequences:

If this is not fixed, users with low-traffic RF devices will continue to experience "fragile" entity availability. This results in broken Home Assistant automations, incorrect state history, and a poor user experience as healthy devices appear offline simply because they have not needed to transmit a message recently.

## The Fix:

This PR pushes the responsibility for availability logic down into `ramses_rf`. It introduces a mechanism to track the latest message timestamp per device and evaluates availability against a `heartbeat_timeout` that is specific to the device class (e.g., 12 hours for TRVs, 24 hours for HVAC remotes).

## Technical Implementation:
- **Centralized Configuration:** Defined `HEARTBEAT_TIMEOUT_*` constants in `src/ramses_rf/const.py`.
- **State Tracking:** Modified `DeviceBase` to capture the timestamp of the latest processed message within the `_handle_msg` pipeline.
- **Logic Delegation:** Implemented a base `is_available` property in `DeviceBase` that compares the current time against the last message received plus the allowed timeout.
- **Class Overrides:** Provided specific `heartbeat_timeout` overrides for `OtbGateway`, `TrvActuator`, `HvacRemoteBase`, and `HvacSensorBase` to reflect real-world transmission intervals.

## Testing Performed:
- **Unit Testing:** Ran the full `pytest` suite; all existing protocol tests passed.
- **Type Checking:** Verified the changes are `mypy` compliant (strict mode).
- **Linting:** Verified code adheres to `ruff` standards.
- **Manual Verification:** Confirmed that `_last_msg_dtm` correctly updates upon message reception and that the `is_available` logic handles `None` cases (pre-telemetry) gracefully.

## Risks of NOT Implementing:

Persistent bugs regarding entity availability in the Home Assistant integration and continued violation of architectural separation between the protocol library and the HA component.

## Risks of Implementing:

There is a minor risk that a specific device model within a class might have a heartbeat longer than the assigned constant, leading to it still going "Unavailable" eventually. However, by using centralized constants, these can now be adjusted in a single location.

## Mitigation Steps:
- Implemented a conservative default timeout of 1 hour for unknown devices.
- Centralized all timeout values in `const.py` to allow for rapid tuning based on user feedback without modification of core logic.
- Ensured that devices are considered "Available" if no message has been received yet to prevent premature unavailability during system startup.

## AI Assistance Disclosure:

This contribution was developed with the assistance of Google Gemini 1.5 Pro for code generation and documentation. No Agentic AI systems were employed; all logic and implementations were reviewed, verified, and manually committed by the author.

### DIFF
--- a/src/ramses_rf/const.py
+++ b/src/ramses_rf/const.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+from datetime import timedelta
 from enum import IntEnum
 from typing import TYPE_CHECKING, Final
 
@@ -155,3 +156,10 @@ WB_STATUS_CODES: Final[dict[str, str]] = {
     "357": "Appliance in air purge mode. Primary heat exchanger air venting program active - approximately 100 seconds.",
     "358": "Three way valve kick. If the 3-way valve hasn't moved in within 48 hours, the valve will operate once to prevent seizure",
 }
+
+# Device Availability Timeouts
+HEARTBEAT_TIMEOUT_DEFAULT = timedelta(hours=1)
+HEARTBEAT_TIMEOUT_OTB = timedelta(hours=24)
+HEARTBEAT_TIMEOUT_TRV = timedelta(hours=12)
+HEARTBEAT_TIMEOUT_REMOTE = timedelta(hours=24)
+HEARTBEAT_TIMEOUT_SENSOR = timedelta(hours=12)
--- a/src/ramses_rf/const.py
+++ b/src/ramses_rf/const.py
@@ -3,7 +3,7 @@
 
 from __future__ import annotations
 
-from datetime import timedelta
+from datetime import timedelta as td
 from enum import IntEnum
 from typing import TYPE_CHECKING, Final
 
@@ -158,8 +158,8 @@ WB_STATUS_CODES: Final[dict[str, str]] = {
 }
 
 # Device Availability Timeouts
-HEARTBEAT_TIMEOUT_DEFAULT = timedelta(hours=1)
-HEARTBEAT_TIMEOUT_OTB = timedelta(hours=24)
-HEARTBEAT_TIMEOUT_TRV = timedelta(hours=12)
-HEARTBEAT_TIMEOUT_REMOTE = timedelta(hours=24)
-HEARTBEAT_TIMEOUT_SENSOR = timedelta(hours=12)
+HEARTBEAT_TIMEOUT_DEFAULT = td(hours=1)
+HEARTBEAT_TIMEOUT_OTB = td(hours=24)
+HEARTBEAT_TIMEOUT_TRV = td(hours=12)
+HEARTBEAT_TIMEOUT_REMOTE = td(hours=24)
+HEARTBEAT_TIMEOUT_SENSOR = td(hours=12)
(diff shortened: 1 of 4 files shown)


PWhite-Eng:
@silverailscolo, please note this Pull Request requires a matching Pull Request in `ramses_cc` to delegate the availability check and enable full functionality.

silverailscolo:
> @silverailscolo, please note this Pull Request requires a matching Pull Request in `ramses_cc` to delegate the availability check and enable full functionality.

@PWhite-Eng Is that matching PR on ramses_cc upcoming shortly, so both could go into a release 0.55.6 tonight?

silverailscolo (dismissed):

PWhite-Eng:
@silverailscolo, I tried to get the matching one for ramses_cc sorted, but I am out of time. I am off on holiday now so I can't finish it, sorry. It will have to wait until I get back on the 29th March.
Phil

silverailscolo:
> off on holiday now

Enjoy! ~We'll set this aside np~ matching PR for ramses_cc (ramses-rf/ramses_cc#561)

wimpie70:
Here's an idea for the future: some devices do respond to certain msgs, so this could be a check for it's status.
Enjoy your holidays :D

PWhite-Eng:
> Here's an idea for the future: some devices do respond to certain msgs, so this could be a check for it's status. Enjoy your holidays :D

@wimpie70 , we need to be really careful with this for two reasons: first, the RAMSES protocol rate-limits the number of transmissions (duty cycle), and second, certain devices limit their radio traffic to conserve battery power. If we start pinging them, we risk blocking other traffic and could significantly shorten their battery life.

PWhite-Eng:
@silverailscolo, I have now created the matching pull request for ramses_cc (ramses-rf/ramses_cc#561), so if you have happy with both, can you merge them and include a version bump?
thanks
Phil

src/ramses_rf/device/base.py
```diff
@@ -9,10 +9,16 @@
 import asyncio
 import logging
 from collections.abc import Iterable
+from datetime import UTC, datetime, timedelta
```
silverailscolo: @PWhite-Eng We use dt and td in our code, see for example [gateway.py line 14](https://github.com/ramses-rf/ramses_rf/blob/master/src%2Framses_tx%2Fgateway.py#L14).

silverailscolo (changes requested):

src/ramses_rf/const.py
```diff
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+from datetime import timedelta
```
silverailscolo: Here also import timedelta as td

silverailscolo (changes requested):
Similar in const.py

silverailscolo:
In fact, specifically for the Gateway Status binar_sensor, I suggest to always have that available (because there's always a gateway in config). Just vary the status binary state to OK<>Problem.

For the System binary_sensor, Availability seems more appropriate.

silverailscolo (approved):
Thanks

PR #539, opened 2026-03-29
Fix: Offload packet log flushing and shutdown to executor to prevent blocking I/O

PWhite-Eng:
## The Problem:

When the packet log rotates (typically around 1 AM), the `_periodic_flush` task synchronously calls `flush_packet_log`. This forces the `MemoryHandler` to flush to the `TimedRotatingFileHandler`, which in turn makes blocking calls to `os.listdir` and `open`  directly on the main asyncio event loop.

This issue was reported in `ramses_cc` Issue ramses-rf/ramses_cc#546.

## Consequences:

Home Assistant detects these blocking operations inside the event loop and throws stability warnings. If left unaddressed, these blocking I/O calls freeze the event loop, causing delayed packet processing, missed RF signals, and general system degradation for the user.

## The Fix:

Offloaded all synchronous, blocking logging I/O operations (flushing and closing file handlers) to a background thread using the event loop's default executor.

## Technical Implementation:
- In `gateway.start()`, the `_periodic_flush` background task now wraps the `flush_packet_log(self._pkt_log_listener)` call in `await self._engine._loop.run_in_executor(None, ...)`.
- In `gateway.stop()`, created a synchronous `_stop_listener` helper function that calls `listener.stop()` and iterates through `listener.handlers` to call `.close()`. This helper is then executed via `await self._engine._loop.run_in_executor(None, ...)`.

## Testing Performed:
- Added `test_gateway_stop_closes_listener_in_executor`: Uses `patch.object` on the event loop to verify that `run_in_executor` is awaited exactly once during `gwy.stop()`, and validates the correct arguments (the stop function and the listener instance) are passed.
- Added `test_gateway_start_initiates_periodic_flush`: Mocks the configuration to include a `flush_interval` and verifies that the `QueueListener` is started and the background `_periodic_flush` task is successfully registered to the loop via `add_task`.
- Ran full pytest suite locally (100% pass) and verified 0 errors under Mypy `--strict`.

## Risks of NOT Implementing:

The integration will continue to cause Home Assistant to log severe stability warnings, and users will experience micro-stutters or lockups in their Home Assistant instance every time the packet log flushes or rotates.

## Risks of Implementing:

There is a minor risk of a race condition if the gateway attempts to log a final packet at the exact millisecond the background executor thread is iterating through and closing the file handlers during system shutdown.

## Mitigation Steps:

To mitigate the shutdown race condition, `await self._engine.stop()` is explicitly called *before* the executor is spun up to tear down the `QueueListener`.
This guarantees that the transport and protocol layers are fully shut down and no new messages are being generated or pushed to the logging queue before the handlers are closed.

## AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/const.py
+++ b/src/ramses_rf/const.py
@@ -3,7 +3,7 @@
 
 from __future__ import annotations
 
-from datetime import timedelta
+from datetime import timedelta as td
 from enum import IntEnum
 from typing import TYPE_CHECKING, Final
 
@@ -158,8 +158,8 @@ WB_STATUS_CODES: Final[dict[str, str]] = {
 }
 
 # Device Availability Timeouts
-HEARTBEAT_TIMEOUT_DEFAULT = timedelta(hours=1)
-HEARTBEAT_TIMEOUT_OTB = timedelta(hours=24)
-HEARTBEAT_TIMEOUT_TRV = timedelta(hours=12)
-HEARTBEAT_TIMEOUT_REMOTE = timedelta(hours=24)
-HEARTBEAT_TIMEOUT_SENSOR = timedelta(hours=12)
+HEARTBEAT_TIMEOUT_DEFAULT = td(hours=1)
+HEARTBEAT_TIMEOUT_OTB = td(hours=24)
+HEARTBEAT_TIMEOUT_TRV = td(hours=12)
+HEARTBEAT_TIMEOUT_REMOTE = td(hours=24)
+HEARTBEAT_TIMEOUT_SENSOR = td(hours=12)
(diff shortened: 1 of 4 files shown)


PWhite-Eng:
@silverailscolo, I have stacked this PR on PR #538  - Refactor: Asynchronous Protocol Alignment and Standardized Command Dispatching, please review and merge that PR before this PR, hopefully the commits from that PR should drop out of this PR.

silverailscolo:
This PR and #538 are stacked on (include same changes as) #533.
I just made a comment on #533 regarding `import timedelta as td`.

Please fix that and _rebase_ these afterwards.

silverailscolo (approved):

PR #546, opened 2026-03-31
Feat: Implement `set_fan_mode` for HvacVentilator to enable native fan control

PWhite-Eng:
### The Problem:

Home Assistant users attempting to control the fan speed of ventilation devices (like the Siber DF Evo 4) via standard UI climate cards were encountering a `NotImplementedError` (see `ramses_cc` Issue #210). While the `ramses_tx` layer already had the `Command.set_fan_mode` packet builder for `22F1` commands, the `HvacVentilator` class lacked the corresponding device-level method to invoke it.

### Consequences:

Without this method, users cannot control their ventilation speeds using standard smart home interfaces. They are forced to rely on complex, error-prone manual `send_command` YAML scripts to inject raw hex frames.

### The Fix:

Added the `set_fan_mode` asynchronous method to the `HvacVentilator` class in `src/ramses_rf/device/hvac.py`.

### Technical Implementation:
- Created `async def set_fan_mode(self, fan_mode: str | int)`.
- For a `22F1` command to be accepted by the ventilator, it typically expects the packet to originate from a bound remote (`REM`). The implementation first queries `self.get_bound_rem()` to impersonate the remote.
- If no remote is bound, the logic gracefully falls back to using the Gateway's own ID (`self.hgi.id`).
- The resolved `src_id` and requested `fan_mode` are passed to `Command.set_fan_mode()`, and the resulting packet is queued via `self._gwy.async_send_cmd()` using `Priority.HIGH` and 2 retries.
- If neither a remote nor an HGI is available, the method raises `exc.CommandInvalid`.

### Testing Performed:

Added the following tests to `tests_rf/device/test_hvac_ventilator.py` to ensure robust coverage:
- `test_set_fan_mode_with_bound_rem`: Verifies correct packet construction and QoS parameters when a bound remote is present.
- `test_set_fan_mode_with_hgi_fallback`: Verifies successful fallback to the HGI ID when no remote is bound in the system topology.
- `test_set_fan_mode_no_src_id_raises`: Ensures a `CommandInvalid` exception is cleanly raised if neither a bound remote nor an HGI is available to act as the packet source.

The `ramses_cc` integration will remain incapable of fully supporting standard Home Assistant climate entities for HVAC devices, degrading the user experience.

### Risks of Implementing:

Extremely low risk. This PR adds a discrete, previously missing control method and does not modify the existing message parsing or state management loops.

### Mitigation Steps:

Implemented strict `src_id` validation before packet construction. By safely raising a `CommandInvalid` exception when the topology lacks a valid source ID, we prevent the library from crafting malformed packets or crashing.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/device/hvac.py
+++ b/src/ramses_rf/device/hvac.py
@@ -857,6 +857,34 @@ class HvacVentilator(FilterChange):  # FAN: RP/31DA, I/31D[9A], 2411
         _LOGGER.debug("No bound REM or DIS devices found for FAN %s", self.id)
         return None
 
+    async def set_fan_mode(self, fan_mode: str | int) -> Packet:
+        """Set the operating mode/speed of the ventilator.
+
+        :param fan_mode: The desired fan mode (e.g., 'low', 'medium', 'high', 'boost').
+        :return: The sent packet.
+        :raises CommandInvalid: If unable to determine a valid source ID.
+        """
+        # 22F1 commands to a FAN typically must originate from a bound Remote (REM)
+        # We attempt to impersonate the first bound REM. If none exists, fallback to HGI.
+        src_id = self.get_bound_rem()
+
+        if not src_id:
+            if self.hgi:
+                src_id = self.hgi.id
+            else:
+                raise exc.CommandInvalid(
+                    f"{self}: Cannot set fan mode without a bound REM or HGI"
+                )
+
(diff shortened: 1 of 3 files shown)


silverailscolo (approved):

silverailscolo:
Thanks @PWhite-Eng Nice

Issue #93, opened 2023-12-05
DT4R - Unexpected code for dst to Rx


BillComben:
I don't know if you are still interested in this, but my Ramses_RF script advised that I reported the  following message
```
11:03:56.594 ||  22:006128 |            |  I | device_battery   |      || {'battery_low': False, 'battery_level': 0.0}
11:03:56.716 ||  22:006100 |  01:023621 |  W | ufh_setpoint     |  00  || {'ufh_idx': '00', 'mode': 'heat', 'setpoint_bounds': (16.0, 25.5)}
11:04:02.185  I --- 22:006100 63:262142 --:------ 10E0 038 000002FF1F00FFFFFFFFFFFFFFFFFFFFFFFF4454342030313700C95817D4D7B6CD12902AED17 < Support the development of ramses_rf by reporting this packet, with the make/model of device: DT2:006100 (device type 22 not known to have signature: 0002FF1F00FFFFFFFF)
11:04:02.185  I --- 22:006100 63:262142 --:------ 10E0 038 000002FF1F00FFFFFFFFFFFFFFFFFFFFFFFF4454342030313700C95817D4D7B6CD12902AED17 < AssertionError(C95817D4D7B6CD12902AED17)
Traceback (most recent call last):
  File "C:\Users\WDC\Evohome\ramses_rf-master/src\ramses_tx\message.py", line 259, in _validate
    result = parse_payload(self)
             ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\WDC\Evohome\ramses_rf-master/src\ramses_tx\parsers.py", line 2840, in parse_payload
    result = _PAYLOAD_PARSERS.get(msg.code, parser_unknown)(msg._pkt.payload, msg)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\WDC\Evohome\ramses_rf-master/src\ramses_tx\parsers.py", line 959, in parser_10e0
    assert msg.verb == RP or not unknown, f"{unknown}"
AssertionError: C95817D4D7B6CD12902AED17
11:04:02.442 ||  22:006100 |            |  I | relay_failsafe   |      || {'failsafe_enabled': False, 'unknown_0': 'FF'}
```
It is monitoring a newly installed 12 Zone Evohome system, The type 22 devices on the system are actually Honeywell DT4R wireless thermostats.

[...]

PR #549, opened 2026-04-01
Fix: Allow CTL to receive DT4R setpoint bounds (Resolves Issues #93, #542)

PWhite-Eng:
### The Problem:

The upstream `ramses_rf` library uses strict validation to ensure only authorized device classes receive specific packet types. Controllers (`01:`) were not permitted to receive `W` (Write) packets for codes `22C9` and `2209`. When a DT4R Thermostat (`22:`) transmitted these local setpoint bounds to the Controller, the `ramses_tx` layer dropped them, throwing a `PacketInvalid: Unexpected code for dst (CTL) to Rx` exception.

### Consequences:

Because these essential configuration packets were discarded, the schema setup was interrupted. TRVs fell into `orphans_heat`, and the HA integration (`ramses_cc`) failed to populate the climate entities entirely (Issue #542). Issue #93 logged the exact same failure pattern for older DT4R firmware using code `2209`.

### The Fix:

Whitelisted the `22C9` and `2209` command codes in the `ramses_tx` routing matrix to allow `DevType.CTL` to receive these bounds, preventing the `PacketInvalid` exception.

### Technical Implementation:
- Added `_2209 = "2209"` to the `Code` `StrEnum`.
- Duplicated the `22C9` `CODES_SCHEMA` regex mapping to handle `2209`.
- Updated `_DEV_KLASSES_HEAT` to allow `DevType.CTL` to process `{I_: {}, W_: {}}` for both `22C9` and `2209`.
- Aliased `parser_2209 = parser_22c9` so both codes resolve to the same dictionary payload.

### Testing Performed:
- **CLI Replay:** Used `client.py parse` on user-provided logs to verify the `PacketInvalid` exception is no longer raised and the payload safely traverses the state machine.
- **Snapshot Testing:** Added real-world `22C9` payloads to `code_22c9.log` and created `code_2209.log` for the legacy variant. Ran `pytest tests/tests/test_parsers.py` to pass the raw hex through the validation matrix and parser via `_proc_log_line`. The smoke tests passed perfectly, proving the parser handles the new rules without throwing internal assertions.

DT4R thermostat users will continue to suffer silent validation failures and remain unable to populate their `climate` entities in Home Assistant.

### Risks of Implementing:

Allowing a Thermostat (`THM`) to send a Write (`W_`) verb to the Controller (`CTL`) could theoretically cause unintended state mutations in the upper `ramses_rf` state machine.

### Mitigation Steps:

We verified the architecture of `zones.py` and `heat.py`. Because the `Zone` class uses `super()._handle_msg(msg)` inherited from `Entity`, the extracted `zone_idx` simply allows the parsed dictionary payload to safely inject into the zone's `state_store` cache without disrupting other state logic. The fix remains strictly scoped to the `ramses_tx` validation matrix.

### AI Assistance Disclosure:


PWhite-Eng:
@silverailscolo, this change is based on the logs in issues ramses-rf/ramses_cc#542 and ramses-rf/ramses_cc#93. This is the first time I have added/modified a device, so if you need any further justification for this change/addtion, please let me know
Phil

silverailscolo (approved):

PR #551, opened 2026-04-01
Feat: add support for local thermostat setpoint bounds (22C9/2209)

PWhite-Eng:
### The Problem:

ramses_rf currently only reads zone setpoint bounds from the central controller's config (000A packet). Local boundary overrides physically configured on local thermostats (like the DT4R) use 22C9 or 2209 packets, which are currently unmapped in the zone API. (Addresses Issue #575).

### Consequences:

Upstream clients (like Home Assistant) cannot access the true local setpoint boundaries. The UI will show standard limits, allowing users to attempt to set targets that the local thermostat hardware will actively reject.

### The Fix:

Exposed the 22C9 and 2209 packet payloads through the Zone API without disrupting the existing 000A configuration structure.

### Technical Implementation:

Added a setpoint_bounds async property to the Zone class in zones.py which retrieves data from the 22C9/2209 state store. Extended the params() payload dictionary to include setpoint_bounds so it is readily available to downstream integration coordinators.

### Testing Performed:

Created tests/tests_rf/test_system_system.py to achieve 100% code coverage for zones.py. Tests explicitly cover the new setpoint_bounds property, state store mock integration, and boundary precedence logic.

Upstream components are forced to rely on inaccurate or superseded limits, leading to an inconsistent state between the physical hardware and the software UI.

### Risks of Implementing:

Minimal. This is purely an additive feature. Existing 000A payload structures used by older systems remain entirely untouched.

### Mitigation Steps:

Separated the local bounds into a new setpoint_bounds property rather than mutating the established config payload. This guarantees zero regressions for integrations expecting legacy 000A structures.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/system/zones.py
+++ b/src/ramses_rf/system/zones.py
@@ -737,6 +737,15 @@ class Zone(ZoneSchedule):
             ),
         )
 
+    async def setpoint_bounds(self) -> dict[str, Any] | None:  # 22C9, 2209
+        """Return the zone's local setpoint bounds if defined by a thermostat."""
+        return cast(
+            dict[str, Any] | None,
+            await self.state_store._msg_value(
+                (Code._22C9, Code._2209), zone_idx=self.idx
+            ),
+        )
+
     async def set_setpoint(self, value: float | None) -> Packet | None:  # 000A/2309
         """Set the target temperature, until the next scheduled setpoint."""
         if value is None:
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
@silverailscolo, please note this is one of a pair of matching PRs to enable this feature, I will shortly upload the other PR to ramses_cc
Phil

silverailscolo (approved):

Issue #537, opened 2026-03-27
Add status to gateway device


silverailscolo:
In order to refactor the Gateway status sensor in ramses_cc from direct gwy querying (not async) to async state, the Gateway Entity should expose a status (recent messages via the device = True) as a property. 

Currently, the integration (client) tries to access gateway.engine.this_message and compare that message's timestamp to a cutoff datetime. This no longer works in 0.55.6

Copy pattern from battery_state

PR #557, opened 2026-04-03
Fix: State restoration, active fault polling, and BDR91 relay demand evaluation (#583, #537)

PWhite-Eng:
### The Problem:

Following a system reboot, `system_mode` and faked sensor temperatures evaluate to `None`. The Evohome `active_fault` entity gets permanently stuck as `Unknown`. Furthermore, BDR91 devices fail to report `relay_demand` when they only broadcast `3EF0`/`3EF1` modulation packets instead of standard `0008` packets. (Resolves ramses_cc issues #583 and #537).

### Consequences:

Home Assistant startup automations fail due to null system states. Bound heating relays show unknown demand states. Users have no visibility into active controller faults, rendering the integration unreliable for critical heating automations.

### The Fix:

Wired device and system state getters to seamlessly fall back to the SQLite `msg_db`. Patched `BdrSwitch` to calculate demand from modulation packets. Re-enabled fault log polling during discovery.

### Technical Implementation:
- Added explicit `self._gwy.msg_db.get()` lookups in `SysMode.system_mode()`, `Temperature.temperature()`, `DhwTemperature.temperature()`, and `Weather.temperature()` before deferring to the volatile `state_store`.
- Overrode `BdrSwitch.relay_demand()` to check `SZ_RELAY_DEMAND` and seamlessly fall back to `MODULATION_LEVEL` via `3EF0`/`3EF1` packets.
- Uncommented the `asyncio.create_task(self.get_faultlog())` call inside `Logbook._setup_discovery_cmds()` to ensure active faults are queried on startup.

### Testing Performed:
- Built `tests/tests_rf/test_device_heat.py` to achieve 100% execution coverage on `device/heat.py`. This explicitly tests `msg_db` fallbacks, uses `PropertyMock` to validate `is_faked` logic, and thoroughly tests the `OtbGateway` modulation priority hacks.
- Expanded `tests/tests_rf/test_system_heat.py` for full coverage of `SysMode` caching and asynchronous `Logbook` polling initialization.
- Executed `pytest --snapshot-update` on `test_regression_rf.py` to capture the newly corrected BDR91 relay demand outputs generated by historical packet replays.

Core state attributes remain permanently broken upon cache restoration and HA restarts. Faked sensors drop offline, and BDR91 modules remain silently degraded, causing severe user confusion.

### Risks of Implementing:

Relying on `msg_db` fallbacks could theoretically surface stale data if the SQLite database is not flushed or pruned correctly between extreme system downtimes.

### Mitigation Steps:

Cached database lookups are strictly evaluated as fallbacks; active packet data in the volatile `state_store` always takes precedence. Existing `ramses_rf` database lifecycle management naturally handles the pruning of expired packets.

### AI Assistance Disclosure:


PWhite-Eng:
Hi @silverailscolo,

Just adding a quick note here to explicitly document and justify the update to the `test_gateway_replay_regression` snapshot included in this PR.

**Why the snapshot was updated:**
During the replay of the historical packet log (`FIXTURE_FILE`), the regression suite processes `3EF0`/`3EF1` modulation packets broadcast by a BDR91 relay. 

Previously, `BdrSwitch` strictly inherited its `relay_demand()` method from `RelayDemand`, meaning it *only* looked for `0008` packets. Because there were no `0008` packets in that specific log sequence for the device, it silently failed to report a demand, and the old snapshot baked in that missing key.

In this PR, we patched `BdrSwitch` to correctly fall back to evaluating `3EF0`/`3EF1` packets if `0008` is absent. As a result, the BDR91 now successfully parses the replay log and outputs `'relay_demand': 1.0` (and `0.0`) in its state dictionary. 

The snapshot diff strictly consists of these newly populated, valid `relay_demand` keys. Updating the snapshot locks in this corrected behavior as our new permanent baseline.

Phil

silverailscolo:
@PWhite-Eng Thanks for this fix. I suggest we include it in Release 0.56.0.
Hoping @peternash has a chance to install ramses_rf@master before I finalise that release.

silverailscolo (approved):

peternash:
@silverailscolo - I’m travelling at the moment but hopefully will get a chance to install master later today

PR #559, opened 2026-04-06
Refactor: Phase 2.2 & 2.5 Big Bang Cutover — RAM-First CQRS MessageStore, Stateless Entities & Collision Abstention

PWhite-Eng:
### The Problem:

This PR resolves the ongoing architectural bottlenecks identified in Issue #530, specifically targeting the goals of Phases 2.2 and 2.5.

Historically, `ramses_rf` maintained system state via localized, distributed `_MessageDB` dictionaries attached to every individual `Device`, `Zone`, and `System` entity. As the project attempted to introduce a centralized, persistent SQLite backend, we adopted a "strangler fig" pattern—running the new SQLite asynchronous disk-writer alongside the legacy distributed dictionaries.

This hybrid approach resulted in severe state fragmentation and catastrophic testing deadlocks. The system suffered from three fatal flaws:
1. **Testing Hell (Thread Contention):** The asynchronous `SQLiteWorker` background thread was fighting the blazing-fast, synchronous `asyncio` event loop during `pytest` runs. This caused cascading timeouts, orphaned threads, and deadlocks. Hacking the tests with `time.sleep()` or forced `flush()` calls merely moved the deadlocks around.
2. **The "Coincidental Match" Eavesdrop Collision:** The heuristic for discovering TRV-to-Zone bindings via `30C9` (Temperature) packets was strictly order-dependent. If a TRV and a Zone shared the exact same temperature (a very common occurrence at thermal equilibrium), the system would blindly bind them. When explicit topology packets (`000C`) arrived later, strict schema validation rejected the heuristic correction, crashing the event loop with a fatal `SystemSchemaInconsistent` exception.
3. **Ghost Packets in Test Logs:** Our chronological test logs contained artificially generated packets with exact, duplicate microsecond timestamps. Dictionary-based caches silently overwrote these packets, causing the gateway to completely "miss" devices during test replays.

### Consequences:

Attempting to maintain three separate storage mechanisms (legacy dicts, RAM cache, and SQLite) became mathematically unsustainable. Leaving the code as-is guaranteed ongoing CI/CD thread deadlocks, flaky eavesdropping tests, and a fragmented source-of-truth for device states. The library would occasionally drop packets, silently permit illegal mid-stream device topology mutations (e.g., TRVs changing parent controllers), and freeze device updates upon encountering schema inconsistencies.

### The Fix:

[...]

PR #560, opened 2026-04-06
Fix: Graceful handling of unsupported OpenTherm message types to prevent sensor drops

PWhite-Eng:
### The Problem:

There is a parser regression affecting OpenTherm sensors. When the gateway encounters certain unsupported or reserved `msg_id` values (e.g., `OEMDiagnosticCode` / 115 returning as `-reserved-` or `Unknown-DataId`), the decoder drops the value keys entirely from the returned dictionary. This causes `KeyError`s in upstream consumers trying to read the state. (References ramses-rf/ramses_cc#583, Point 2).

### Consequences:

Because the keys are missing, upstream integrations (like Home Assistant) fail to parse the sensor state, resulting in roughly half of the OpenTherm sensors becoming permanently stuck in an `Unknown` state.

### The Fix:

Updated the decoder to guarantee that the expected value keys (`SZ_VALUE`, `SZ_VALUE_HB`, and `SZ_VALUE_LB`) are always populated—even for unsupported message types—by explicitly mapping them to `None`.

### Technical Implementation:

Modified `decode_frame` in `src/ramses_tx/opentherm.py`. Replaced the simple early-return block for error/reserved `msg_type`s (`0b000`, `0b010`, `0b011`, `0b110`, `0b111`) with logic that inspects the `msg_schema`. It now explicitly injects the expected dictionary keys with `None` values before returning, allowing upstream consumers to fail gracefully rather than crashing.

### Testing Performed:

Created a brand new test suite at `tests/tests_tx/test_opentherm.py` achieving 100% code coverage of the modified parsing logic. Built a custom frame generator that calculates dynamic 31-bit parity to mimic real hardware. Tests comprehensively cover `READ_DATA` null injection, `READ_ACK` payload processing, boundary limits, and corrupt schema fallbacks across every distinct schema type (`FLAG8`, `F8_8`, `U16`, `S8`).

Users will continue to experience broken OpenTherm sensors (`Unknown` state) and silent parsing exceptions in the background.

### Risks of Implementing:

Minimal. The change only affects the edge-case handling of OpenTherm frames that are already flagged by the protocol as invalid, reserved, or unknown.

### Mitigation Steps:

Implemented a highly rigid unit test suite to guarantee that valid `READ_ACK` messages continue to parse flawlessly and that only the explicitly targeted error types receive the `None` injection.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_tx/opentherm.py
+++ b/src/ramses_tx/opentherm.py
@@ -1164,6 +1164,24 @@ def decode_frame(
     if msg_type in (0b000, 0b010, 0b011, 0b110, 0b111):
         # if frame[4:] != "0000":  # NOTE: this is not a hard rule, even for 0b000
         #     raise ValueError(f"Invalid data-value for msg-type: 0x{frame[4:]}")
+
+        # Ensure expected value keys are populated to prevent KeyErrors upstream
+        if not msg_schema:
+            data_value[SZ_VALUE] = None
+        elif isinstance(msg_schema[VAL], dict):
+            val_hb = msg_schema[VAL].get(HB, msg_schema[VAL])
+            val_lb = msg_schema[VAL].get(LB, msg_schema[VAL])
+            if val_hb == FLAG8 and val_lb == FLAG8:
+                data_value[SZ_VALUE] = None
+            else:
+                data_value[SZ_VALUE_HB] = None
+                data_value[SZ_VALUE_LB] = None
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
@silverailscolo, I have branched this off the current master, not off the branch for PR #559, as it does not interfer with any files which that PR changes, therefore you can happy merge this one ahead of #559 if you wish
Phil

silverailscolo (approved):

Issue #545, opened 2026-03-31
Architecture: Move packet caching from "Stringly Typed" logs to structured Data Transfer Objects (DTOs)


PWhite-Eng:
Hi @silverailscolo,

While debugging a downstream issue in `ramses_cc` where cached packets were failing known-list validation, we uncovered an architectural fragility in how packets are serialized to persistence.

Currently, the `client_state` cache (managed by `storage.py`) flattens `Packet` objects into raw strings. We recently noticed that `Packet._from_cmd()` injects a dummy RSSI (`"... "`) to maintain visual alignment for human-readable logs:

``` Python
return cls.from_port(dtm, f"... {cmd._frame}")

```

Because the storage layer persists this exact string, downstream consumers (like HA) that rely on strict character indexing (`pkt[11:20]`) to parse out device IDs suddenly broke.

Using human-readable string representations (`__str__` / `__repr__`) for data persistence is a "stringly typed" anti-pattern. We should separate the visual representation of a packet from its structured data.

#### Options for Resolution

**Option 1: Structured DTOs / JSON Serialization (Recommended)** Modify `storage.py` and the `Packet` class to serialize state as structured dictionaries rather than concatenated strings. For example: `{"dtm": "...", "rssi": null, "verb": "I", "src": "04:071355", "code": "2309", "payload": "030640"}`
- **Pros:** Industry best practice (Parse Early, Serialize Late). It guarantees data integrity. We can change the console logging format at will without ever breaking the persistence layer or downstream integrations.
- **Cons:** This is a breaking change for `client_state`. It will require migration logic (or a cache wipe) for existing users, and updates to the `ramses_cc` cache-reading logic.

**Option 2: Standardize the Dummy RSSI** We enforce a strict, predictable rule for missing RSSIs in `packet.py` (e.g., exactly 3 spaces, or `---`, with no trailing spaces) so that character indices remain absolute across the entire ecosystem.
- **Pros:** Quick fix. No breaking changes to the database schema or cache files.
- **Cons:** It's a band-aid. It keeps us locked into string-slicing for business logic, meaning the next time we add parser hints or formatting, we risk breaking things again.

**Option 3: Explicit `.to_cache()` / `.from_cache()` methods on `Packet`** We create a highly rigid, unchanging string format specifically for the cache that strips out parser hints, dummy RSSIs, and comments, returning *only* the raw frame.
- **Pros:** Solves the immediate issue without requiring a full JSON schema migration.
[...]

PR #562, opened 2026-04-06
Feature: Migrate client state caching to structured JSON DTOs (#545)

PWhite-Eng:
### The Problem:

Currently, the integration cache (managed by `get_state`) flattens physical packets into raw log strings. Because the downstream persistence layer relies on exact character formatting, rigid string slicing (e.g., `pkt[11:20]`) fails whenever parser hints or formatting changes occur (such as the recently added dummy RSSI padding). (References `ramses_rf` Issue #545, addresses `ramses_cc` Issue ramses-rf/ramses_cc#588 ).

### Consequences:

If not fixed, simple cosmetic changes to packet logs will continue to break the persistence layer. During Home Assistant restarts, cached packets are silently discarded, causing Climate and DHW entities to fail restoration and appear as `Unknown`.

### The Fix:

Migrated the entire cache serialization pipeline away from "stringly-typed" logs to structured Data Transfer Objects (DTOs) via JSON serialization.

### Technical Implementation:
- Added strict `PacketDTO` and `DeviceAddress` models to `packet.py`.
- Implemented `to_dto()` and `to_dict()` methods on the `Packet` class. These map physical frame components into explicit keys (`dtm`, `verb`, `addr1`, `code`), while accepting `parsed_payload` injected directly from the higher-level `Message` wrapper.
- Added a `frame` column to `SQLiteWorker` to persistently store the exact raw packet string. Updated `MessageStore._hydrate_ram()` to reconstruct packets losslessly using this column instead of generating dummy string fragments.
- Updated `Gateway._restore_cached_packets()` to intelligently parse either the new dictionary DTOs or the legacy string format, mapping them back into native Python strings for the underlying transport layer.

### Testing Performed:

Ran `mypy --strict` with zero errors. Wrote targeted unit tests in `test_packet.py` for DTO serialization/deserialization and timezone-aware handling. Updated `test_sqlite_worker.py` to ensure lossless `frame` hydration across the SQL boundary. Verified backward compatibility for legacy string caches in `test_gateway.py`. The full pytest suite passes seamlessly.

The architecture remains brittle, and cache persistence remains permanently coupled to visual log formatting, guaranteeing future entity drops.

### Risks of Implementing:

The JSON output structure fundamentally changes the payload exported to Home Assistant. Downstream integrations must be updated to consume this dictionary format.

### Mitigation Steps:

Wrote highly defensive unpacking logic in `Packet.from_dict` and `Gateway._restore_cached_packets`. The implementation natively checks `isinstance(state, dict)` and gracefully falls back to legacy string parsing, guaranteeing that existing users upgrading to this version will not experience cache corruption or require a storage wipe.

### AI Assistance Disclosure:

---

### DIFF
--- a/src/ramses_tx/packet.py
+++ b/src/ramses_tx/packet.py
@@ -6,7 +6,11 @@ Decode/process a packet (packet that was received).
 
 from __future__ import annotations
 
+import contextlib
+from dataclasses import asdict, dataclass
 from datetime import datetime as dt, timedelta as td
+from enum import StrEnum
+from typing import Any
 
 from .command import Command
 from .const import I_, RP, RQ, W_, Code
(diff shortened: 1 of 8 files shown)


PWhite-Eng:
@silverailscolo, this PR is **Part 1 of a paired update**. This introduces the JSON DTOs and structured backend serialization required to fix the cache loading bugs.

I have raised a companion PR in `ramses_cc` to update the Home Assistant coordinator to consume these new dictionaries instead of using strict string slicing. > https://github.com/ramses-rf/ramses_cc/pull/596

**Merge Strategy:** These two PRs should ideally be merged at the same time, and `ramses_cc` will require a version bump in its manifest/requirements to lock against this new `ramses_rf` release to ensure the schemas match.

src/ramses_tx/packet.py
```diff
@@ -29,6 +33,122 @@
...
 
 
+class Verb(StrEnum):
+    """Enumeration of standard RAMSES-RF packet verbs.
+
+    The whitespace in Information and Write verbs is explicitly preserved
+    as required by the protocol.
+    """
+
+    I_ = " I"
```
silverailscolo: This is also in ramses_tx/const.py:
```
@verify(EnumCheck.UNIQUE)
class VerbT(StrEnum):
    """Protocol verbs (message types)."""

    I_ = " I"
    RQ = "RQ"
    RP = "RP"
    W_ = " W"


I_: Final = VerbT.I_
RQ: Final = VerbT.RQ
RP: Final = VerbT.RP
W_: Final = VerbT.W_
```

silverailscolo (changes requested):
Good stuff, 1 comment to prevent duplication in the code

src/ramses_tx/packet.py
PWhite-Eng: I thought I saw that somewhere, thanks for pointing it out. I have updated the code to import it from const.py

PWhite-Eng:
@silverailscolo, I had to re-write `tests/tests_rf/test_hgi_behaviors.py` as Github's CI tests were not happy with it. I acknowledge that it should ideally be a separate PR, but I have included it in this PR as this PR was failing without it.

silverailscolo (approved):

PR #520, opened 2026-03-10
Refactor: Remove Legacy Gateway Bridges and Finalize DeviceRegistry DI

PWhite-Eng:
## The Problem:

During the previous phase of deconstructing the `Gateway` God Object (where we successfully extracted `DeviceRegistry` and `DeviceFilter`, and applied Composition Root), we temporarily left "bridge" methods on the `Gateway` class (e.g., `@property def devices`, `get_device()`). These were left in place so existing files wouldn't break immediately. However, leaving these legacy wrappers in place obscures the new architecture, keeps the `Gateway` API surface artificially bloated, and encourages continued tight coupling.

## Consequences:

If these bridge methods are not removed, new code will continue to rely on the `Gateway` for device management. This negates the architectural benefits of the Dependency Injection work, maintains technical debt, and prevents true interface segregation.

## The Fix:

Removed all temporary forwarding/bridge methods from the `Gateway` and its interface. Updated the entire codebase and test suite to explicitly interact with the newly decoupled `DeviceRegistry`.

## Technical Implementation:
- Removed `@property def devices`, `device_by_id`, `system_by_id`, `systems`, and functions like `get_device()`, `fake_device()`, `known_list()`, and `_add_device()` directly from `gateway.py`.
- Cleaned up `GatewayInterface` in `interfaces.py` to ensure it no longer mandates device-handling methods.
- Updated all internal routing (`dispatcher.py`, `schemas.py`, `device/base.py`, `system/heat.py`, and CLI modules) to explicitly call `gwy.device_registry.X`.
- Updated `GatewayStub` and `mock_gateway` in the test suite to expose a `device_registry` property, repairing the test doubles.
- Applied `typing.cast()` in internal builder modules (like `schemas.py`) to satisfy strict Mypy checks where concrete `Device` objects are required from the generalized interface.

## Testing Performed:
- Executed the full `pytest` suite (825 tests) with a 100% pass rate.
- Executed strict `mypy` static type checking across all 130 source files, resolving all 60+ initial fallout errors down to 0.

## Risks of NOT Implementing:

Technical debt will accumulate as developers continue using the old `gwy.get_device()` patterns instead of fully adopting the `DeviceRegistry` pattern.

## Risks of Implementing:

This is a high-touch cleanup PR impacting many files. Un-merged feature branches or downstream scripts relying on `gwy.devices`, `gwy.get_device()`, or `gwy.device_by_id` will experience `AttributeError` breakages upon rebasing.

## Mitigation Steps:
[...]

PR #572, opened 2026-04-08
Fix: restore `gwy.fake_device`, imports

silverailscolo:
**PR Summary**

Fixes ramses_cc issue https://github.com/ramses-rf/ramses_cc/issues/598

@PWhite-Eng Please. review as you took it out on March 10

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -60,7 +60,7 @@ from ramses_tx import (
 from ramses_tx.transport import TransportConfig
 from ramses_tx.typing import PktLogConfigT, PortConfigT
 
-from .device import HgiGateway
+from .device import Fakeable, HgiGateway
 from .device_filter import DeviceFilter
 from .device_registry import DeviceRegistry
 from .dispatcher import detect_array_fragment, process_msg
@@ -73,7 +73,7 @@ from .interfaces import (
 from .message_store import MessageStore
 from .schemas import load_schema
 from .system import Evohome
-from .typing import DeviceListT
+from .typing import DeviceIdT, DeviceListT
 
 if TYPE_CHECKING:
     from ramses_tx import RamsesTransportT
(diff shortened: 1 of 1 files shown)


silverailscolo:
Fix is working on HVAC setup, where it ends with an expected error as no fan is in pairing mode to reply:

```
Logger: homeassistant.components.websocket_api.http.connection
Bron: components/websocket_api/connection.py:315
integratie: Home Assistant WebSocket API (documentatie, problemen)
Eerst voorgekomen: 11:47:53 (1 gebeurtenis)
Laatst gelogd: 11:47:53

[546305272672] Error handling message: Unexpected error during binding for 37:117647: 37:117647: SuppSendOfferWaitForAccept: Failed to transition to <class 'ramses_rf.binding_fsm.SuppIsReadyToSendConfirm'>: expected message not received after 5.1 secs (home_assistant_error)
[...] Home Assistant/2026.2.1 (io.robbie.HomeAssistant; build:2026.1766; iOS 18.7.7) Mobile/HomeAssistant, like Safari)
```

PWhite-Eng:
Hi @silverailscolo,

Thank you for jumping on this so quickly to stop the bleeding for users facing the binding bug! I completely understand why you raised this PR—restoring the method directly to the Gateway is the fastest way to get `ramses_cc` working again.

However, I'd like to propose we fix this downstream in `ramses_cc` instead, and I have just raised PR ramses-rf/ramses_cc#598 over there to do exactly that.

**Context on the architecture:** Back in PR #520, the primary goal was to dismantle the `Gateway` God Object and strictly enforce Dependency Injection. By segregating all device management into the `DeviceRegistry`, we decoupled the modules. Leaving the temporary bridge methods (like `fake_device`) on the `Gateway` obscures this new architecture and encourages continued tight coupling.

While restoring `gwy.fake_device` as a passthrough fixes the immediate bug, it is a slight step backward architecturally, as it re-introduces the technical debt and API bloating we worked hard to remove in PR #520.

**The proposed path forward:** The architecturally "correct" fix is to update the downstream consumer (`ramses_cc`) so it calls `self._coordinator.client.device_registry.fake_device(...)` directly. My new PR in `ramses_cc` implements this along with strict testing to prevent future regressions.

If you agree with this approach, please review the `ramses_cc` PR. Once we merge that, we can safely close this PR without needing to revert our cleanups in `ramses_rf`.

Thanks

Phil

PWhite-Eng:
Hi @silverailscolo,

Just a quick follow-up to my previous message

To be absolutely certain we hadn't left any other PR #520 landmines hiding in the integration, I just completed a comprehensive regex audit of the `ramses_cc` codebase. I specifically searched for any lingering calls to the other legacy bridge methods that were removed from the Gateway object (such as `client.devices`, `client.device_by_id`, `client.get_device`, `client.systems`, etc.).

I'm happy to report that the search came up completely clean.

The rest of the `ramses_cc` coordinator and internal handlers have already been successfully updated to use the new `DeviceRegistry` Dependency Injection pattern. It turns out that `fake_device` was just an isolated oversight, likely because it is triggered manually via an edge-case Developer Tools service call rather than the core integration polling loop.

With that confirmed, we are highly confident there are no other outstanding downstream breakages in `ramses_cc` stemming from the `ramses_rf` architecture update.

Phil

PR #574, opened 2026-04-08
Fix `PortTransport` initialization race condition and resolve `pyserial` typing

PWhite-Eng:
### The Problem:

During startup, if the serial port receives data immediately upon connection, the `_read_ready()` callback fires before the transport has fully completed its async setup. This causes an `AttributeError: 'PortTransport' object has no attribute '_init_fut'`, crashing the integration (reported in Issue #583). Additionally, writing tests for this exposed missing type stubs for `pyserial`, resulting in scattered `# type: ignore` technical debt across the codebase.

### Consequences:

If this isn't fixed, users with active serial environments (where data is flowing the moment the port opens) will experience silent initialization failures or complete integration crashes upon restarting Home Assistant.

### The Fix:

Synchronously initialized the transport future to guarantee it exists before any I/O can trigger, and updated the development environment to natively support `pyserial` type hinting.

### Technical Implementation:
- Moved `self._init_fut = self._loop.create_future()` directly into `PortTransport.__init__` rather than waiting for the background `_create_connection()` tasks to spawn it.
- Updated `_pkt_read` to safely validate against `self._extra.get(SZ_SIGNATURE)` using `.get()` in case early packets arrive before the signature is written to the extra dict.
- Added `types-pyserial` to `requirements_dev.txt` and removed the legacy `# type: ignore[import-untyped]` comments across 13 affected files.

### Testing Performed:

Created `tests/tests_tx/test_transport_port.py` with 100% coverage.
- Wrote `test_port_transport_init_initializes_future_immediately` to explicitly trigger `_read_ready()` before background tasks run to guarantee the `AttributeError` is resolved.
- Mocked the underlying `asyncio` event loop read selectors to ensure tests safely mock serial file descriptors without throwing internal `PermissionError`s.

Persistent, environmentally dependent startup crashes for users based on how chatty their physical serial devices are during boot.

### Risks of Implementing:

Very low. Moving the `Future` initialization synchronously is standard practice, though altering anything in the startup transport sequence carries a minor risk of altering connection timing.

### Mitigation Steps:

Wrote comprehensive async unit tests that intercept and manipulate the background initialization tasks specifically to ensure all edge cases gracefully time out or resolve as intended.

### AI Assistance Disclosure:


PWhite-Eng:
Hi @silverailscolo,

I've just submitted this PR to fix the startup crash reported by Peter in ramses-rf/ramses_cc#583 (`AttributeError: 'PortTransport' object has no attribute '_init_fut'`).

Full disclosure: I went down a bit of a rabbit hole with this one! While writing the test file to ensure 100% coverage of the fix, MyPy flagged the missing `serial` type stubs. Rather than just dropping another `# type: ignore[import-untyped]` on the new test file, I decided to fix the root cause. I added `types-pyserial` to `requirements_dev.txt` and cleaned up the ignore comments across the 13 files where we previously had to bypass it.

I know the typing cleanup really should have been a separate PR, but since it was a direct blocker for getting a clean MyPy `--strict` run on the new transport test, I bundled it here.

**Quick reminder:** When you pull this branch down, please remember to run `pip install -r requirements_dev.txt` in your venv so your local MyPy picks up the new `types-pyserial` package and gives you a clean check

Phil

silverailscolo (approved):

PR #578, opened 2026-04-09
Fix: Resolve SQLite database locking by routing deletions through single-writer queue

PWhite-Eng:
### The Problem:

Users are experiencing frequent `sqlite3.OperationalError: database table is locked` tracebacks. This occurs because `message_store.rem()` was bypassing the `SQLiteWorker` daemon thread and using `asyncio.to_thread` to spawn unmanaged threads for `DELETE` queries. When these rogue threads collided with the background worker (which holds the WAL lock for inserts/pruning), SQLite rejected the concurrent write attempts.

### Consequences:

If left unfixed, this concurrency collision causes constant database write failures. On slower hardware (like Raspberry Pi SD cards), the resulting disk I/O bottleneck aggressively starves the Home Assistant asyncio event loop, causing system-wide timeouts, dropped RF telemetry, and corrupted serial buffers.

### The Fix:

Enforced a strict single-writer policy for the database. All `DELETE` operations are now routed through the `queue.SimpleQueue` so they are executed sequentially by the dedicated `SQLiteWorker` background thread, completely eliminating concurrent write collisions.

### Technical Implementation:
- Created a strongly-typed `DeleteMessageRequest(NamedTuple)` in `sqlite_worker.py` to safely pass SQL queries and parameters into the queue.
- Added `submit_delete_message()` to expose this functionality to the frontend.
- Refactored `message_store.rem()` to drop its `asyncio.to_thread` logic and instead push the dynamically generated `DELETE` query to the worker queue.
- Replaced the generic tuple used for flushing the queue with a strictly-typed `FlushRequest(NamedTuple)` to satisfy `mypy --strict`.

### Testing Performed:
- **`test_message_store.py::test_rem_msg`**: Added to verify that messages are accurately removed from the RAM cache (`self._message_log` and `self._state_cache`) using both exact message objects and arbitrary `kwargs` (e.g., `code="2309"`).
- **`test_sqlite_worker.py::test_storage_worker_delete`**: Added to verify that the async worker successfully consumes the `DeleteMessageRequest`, executes the SQL, and permanently removes the record from the physical `.sqlite` file.
- Verified that the entire project (137 source files) successfully passes `mypy --strict` with no issues.

Users will continue to experience fatal SQLite locking tracebacks. On lower-end hardware, the resulting I/O wait times will continue to crash surrounding Home Assistant integrations and corrupt incoming RF packet streams.

### Risks of Implementing:

If the `SQLiteWorker` background thread were to silently crash, deletions (and insertions) would continue to succeed in the RAM cache but would never persist to the disk, creating a silent failure where the queue grows indefinitely in memory.

### Mitigation Steps:

The `SQLiteWorker` includes broad exception handling within its `_run` loop to catch and log specific `sqlite3.Error` tracebacks without killing the daemon thread itself. Furthermore, we implemented explicit `flush()` mechanics in the test suite to physically verify that queue items are drained and committed to disk rather than just hanging in memory.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/message_store.py
+++ b/src/ramses_rf/message_store.py
@@ -516,9 +516,8 @@ class MessageStore:
                 ctx=ctx,
                 hdr=hdr,
             )
-            cx = getattr(self, "_cx", None)
 
-            if self._worker and cx is not None:
+            if self._worker:
                 if msg:
                     kwargs["dtm"] = msg.dtm
 
@@ -527,13 +526,7 @@ class MessageStore:
 
                 sql_params = tuple(kwargs.values())
 
-                def _execute_delete(
-                    conn: sqlite3.Connection, query: str, params: tuple[Any, ...]
-                ) -> None:
-                    with self._db_lock:
-                        conn.execute(query, params)
-
-                await asyncio.to_thread(_execute_delete, cx, sql, sql_params)
+                self._worker.submit_delete_message(sql, sql_params)
 
             msgs = tuple(msgs_to_remove)
 
--- a/src/ramses_rf/sqlite_worker.py
+++ b/src/ramses_rf/sqlite_worker.py
@@ -41,7 +41,27 @@ class SnapshotRequest(NamedTuple):
     pass
 
 
-QueueItem = PacketLogEntry | PruneRequest | SnapshotRequest | tuple[str, Any] | None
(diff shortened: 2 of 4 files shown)


silverailscolo (approved):
Thanks!

Issue #580, opened 2026-04-11
Comprehensive architectural review and diagnostic report of the ramses_rf and ramses_tx codebases


PWhite-Eng:
Here is the comprehensive architectural review and diagnostic report of the `ramses_rf` and `ramses_tx` codebases undertaken by Gemini 3.1. 

The set objective was twofold:
1. Analyze the codebase to pinpoint the root causes of 5 specific, known performance/architectural issues.
2. Perform a codebase-wide audit against 7 strict industry best-practice pillars.

### PART 1: Specific Known Issues to Locate and Analyze

It was asked to review the following 5 specific known Issues:

**Issue 1: State Rehydration Bloat (Startup)**
When `ramses_rf` loads its historical packet cache (typically the last 24 hours), it needlessly instantiates thousands of high-volume sync/status packets (specifically message codes `0004`, `1060`, and `2349`), which eats memory and blocks the event loop.
* **Goal:** Locate the cache loading mechanism. Formulate a plan to KEEP packets from the last 1 hour, STRIP the specific high-volume codes for hours 2-24, define these as readable constants, and enforce `await asyncio.sleep(0)` yielding.

**Issue 2: Event Loop Starvation on `get_state()`**
Home Assistant calls `client.get_state()` every 5 minutes to save the cache to disk. If this dictionary compilation is synchronous, it freezes the host system.
* **Goal:** Locate `get_state()`. Check if it is truly `async def` and if its internal loops properly yield (`await asyncio.sleep(0)`). Warn if it relies on blocking CPU-heavy dictionary copies.

**Issue 3: Zone Name Resolution Looping**
There is a severe performance bottleneck in `ramses_rf.system.zones` related to `await self.name()`. The system repeatedly executes "Pick Zone.name from:" logic, redundantly evaluating dozens of duplicate `0004` (Zone Name) packets restored from the cache for every single zone (00 to 0B).
* **Goal:** Locate this logic. Formulate a plan to implement a "short-circuit" return if the name is resolved, and deduplicate cached `0004` packets early.

**Issue 4: QoSProtocol Timeouts (Cascade Effect)**
Outgoing radio packets fail to receive timely ACKs. This is likely a cascade effect from Issues 1, 2, and 3 starving the event loop, leaving `ramses_tx` without compute time.
* **Goal:** Verify if `ramses_tx` initialization and packet parsing have proper loop integrity and yielding.

**Issue 5: Packet to Message Routing Pipeline & Heartbeats**
[...]

PR #581, opened 2026-04-11
Fix: Resolve async starvation and optimize state cache restoration

PWhite-Eng:
### The Problem:

During gateway startup and shutdown, the loading and saving of the state cache relied on synchronous dictionary comprehensions and tight `for` loops. Additionally, the cache restoration process inefficiently converted structured DTOs back into strings to funnel them through a legacy mock transport layer, which subjected them to heavy regex parsing. Finally, the cache was infinitely accumulating high-frequency status messages, exacerbating the load.

### Consequences:

Processing large state caches monopolized the Python thread, causing severe async starvation. Because the event loop was entirely blocked, other critical background tasks—such as network I/O, heartbeats, or user interactions—would experience significant latency, timeouts, or silent failures during gateway initialization and shutdown.

### The Fix:

Implemented cooperative multitasking by chunking the cache loops, actively pruned stale/noisy telemetry data, and bypassed the legacy string-based transport layer entirely to hydrate `Packet` objects directly from the DTOs.

### Technical Implementation:
- **Data Pruning:** Defined `HIGH_VOLUME_STATUS_CODES` in `const.py` (`_0004`, `_1060`, `_2309`, `_2349`, `_30C9`).
- **Cooperative Multitasking:** Replaced the synchronous dictionary comprehension in `gateway.py -> get_state()` with an `async for` loop, executing `await asyncio.sleep(0)` every 100 iterations. Applied the same chunking to `_restore_cached_packets()`.
- **Direct DTO Hydration:** Removed the `transport_factory` dependency in `_restore_cached_packets()`. The loop now iterates over the DTOs and instantiates packets directly via `Packet.from_dict(dtm, state)`, feeding them straight into `protocol.pkt_received()`.
- **Time-based Expiration:** Added logic to evaluate the `dtm` of cached packets, dropping any that match `HIGH_VOLUME_STATUS_CODES` if they are older than `dt.now() - timedelta(hours=1)`.
- **Test Fixes:** Updated `test_gateway_restore_cached_packets_dto` to mock `Packet.from_dict` to prevent strict regex validation errors on synthetic test frames, verifying the new direct-to-protocol routing logic.

### Testing Performed:
- Updated `tests/tests_rf/test_gateway.py::test_gateway_restore_cached_packets_dto` to verify that `Packet` objects are instantiated directly from DTOs and injected into the protocol handler.
- Ran `mypy --strict` across the codebase (0 errors).
- Ran the full `pytest` suite (100% pass rate).

Leaving the code as is guarantees event loop blocking on systems with large caches. As users accumulate larger databases of packet history, gateway startup times will degrade to the point of triggering application watchdogs or causing disconnects on related integrations.

### Risks of Implementing:

Because we are bypassing the legacy transport emulator's parsing buffer, there is a small risk that malformed or legacy cache entries (which the transport layer might have silently sanitized or dropped) could throw exceptions during direct `Packet.from_dict()` hydration.

### Mitigation Steps:

Wrapped the new direct `Packet.from_dict()` instantiation and `pkt_received()` injection in a robust `try/except` block. Any packet that fails strict structural hydration will be caught, logged as a debug warning, and skipped, ensuring the primary restoration loop continues uninterrupted.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/const.py
+++ b/src/ramses_rf/const.py
@@ -139,22 +139,51 @@ DONT_UPDATE_ENTITIES: Final[int] = 1
 
 SCHED_REFRESH_INTERVAL: Final[int] = 3  # minutes
 
+HIGH_VOLUME_STATUS_CODES: Final = (
+    Code._0004,
+    Code._1060,
+    Code._2309,
+    Code._2349,
+    Code._30C9,
+)
+
 # Status codes for Worcester Bosch boilers - OT|OEM diagnostic code
 WB_STATUS_CODES: Final[dict[str, str]] = {
     "200": "CH system is being heated.",
     "201": "DHW system is being heated.",
-    "202": "Anti rapid cycle mode. The boiler has commenced anti-cycle period for CH.",
+    "202": (
+        "Anti rapid cycle mode. The boiler has commenced anti-cycle period for CH."
+    ),
     "203": "System standby mode.",
     "204": "System waiting, appliance waiting for heating system to cool.",
     "208": "Appliance in service Test mode (Min/Max)",
-    "265": "EMS controller has forced stand-by-mode due to low heating load (power required is less than the minimum output)",
-    "268": "Component test mode (is running the manual component test as activated in the menus).",
+    "265": (
+        "EMS controller has forced stand-by-mode due to low heating "
+        "load (power required is less than the minimum output)"
+    ),
+    "268": (
+        "Component test mode (is running the manual component test as "
+        "activated in the menus)."
+    ),
     "270": "Power up mode (appliance is powering up).",
     "283": "Burner starting. The fan and the pump are being controlled.",
-    "284": "Gas valve(s) opened, flame must be detected within safety time. The gas valve is being controlled.",
-    "305": "Anti fast cycle mode (DHW keep warm function). Diverter valve is held in DHW position for a period of time after DHW demand.",
-    "357": "Appliance in air purge mode. Primary heat exchanger air venting program active - approximately 100 seconds.",
-    "358": "Three way valve kick. If the 3-way valve hasn't moved in within 48 hours, the valve will operate once to prevent seizure",
+    "284": (
+        "Gas valve(s) opened, flame must be detected within safety "
+        "time. The gas valve is being controlled."
+    ),
+    "305": (
+        "Anti fast cycle mode (DHW keep warm function). Diverter valve "
+        "is held in DHW position for a period of time after DHW demand."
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
address PR1 in #580

PWhite-Eng:
@silverailscolo, please note that I have instructed my friend to write code that is PEP8 compliant, therefore they are implimenting line lengths that are <=79 characters long, and comments and docstrings are <=74 characters long. This is causing alot of formatting changes in the file, but all code and comments are preserved. It does make it harder to find the actual changes. 
Sorry
Phil

silverailscolo (approved):
PEP reformat np to include

silverailscolo:
> PEP8 reformat np to include

Could we add this to the ruff  formatter settings?

PWhite-Eng:
@silverailscolo, I have just been down the rabbit hole of looking at implimenting the line length standard. It would be using ruff. I have turned it on in ramses_rf and there was nearly 4000 (yes, that is not a typo, four thousand) lines which were too long. I then turned off the line length check and there was no problems reported by VS Code or by mypy, all was good!
I will bravely leave this for someone else to take up!

silverailscolo:
> implimenting the line length standard. It would be using ruff. 

Can you leave the required change commented out in pyproject.toml?

PWhite-Eng:
Deep in the middle of PR3, but this is what you need to change/add to the pyproject.toml
Change this section:
```
[tool.ruff.lint]
  select = [
    "ASYNC",  # flake8-async
    "B",  #     flake8-bugbear
    "E",  #     pycodestyle
    "F",  #     Pyflakes
    "I",  #     isort
    "SIM",  #   flake8-simplify
    "UP",  #    pyupgrade
  ]
  ignore = ["ASYNC109", "ASYNC110", "B011", "SIM102", "SIM114", "UP040"]

  # B011   - Do not call assert False since python -O removes these calls
  # E501   - Line too long (Note: Removed from ignore list to enable check)
  # SIM102 - Use a single `if` statement instead of nested `if` statements
  # SIM114 - Combine `if` branches using logical `or` operator
  # UP040  - Type alias uses `TypeAlias` annotation instead of the `type` keyword
```
The change is in the ignore list, you need to remove E501

To add just below the above section
```
[tool.ruff.format]
  # Enforce PEP 8 docstring and comment wrapping
  docstring-code-format = true


[tool.ruff.lint.pycodestyle]
  # PEP 8 recommends 72 characters for docstrings and comments
  max-doc-length = 72
```

PR #582, opened 2026-04-11
perf: Eliminate Zone Name database query bottleneck via active RF caching

PWhite-Eng:
### The Problem:

Fetching a zone name previously resulted in repeated, expensive database queries to the `message_store` (or `entity_state`). Because zone names are requested frequently by downstream UI layers, this caused a significant performance bottleneck and unnecessary I/O looping.

### Consequences:

If left unfixed, the system suffers from degraded performance, high CPU utilization, and unnecessary SQLite database locks, particularly on larger setups with many zones being polled frequently by integrations like Home Assistant.

### The Fix:

Implemented a real-time, event-driven caching mechanism for Zone Names. The zone name is now cached in memory (`self._name`). It is fetched from the database exactly once on startup (if needed) and is subsequently kept perfectly synchronized by passively eavesdropping on `0004` (Zone Name) RF packets as they arrive.

### Technical Implementation:
- Initialized `self._name: str | None = None` in `ZoneBase`.
- Refactored `Zone.name()` to return early if `self._name` is populated.
- If unpopulated, `Zone.name()` executes `message_store.get()` once and caches the result.
- Added logic in `Zone._handle_msg()` to intercept `Code._0004`. When a name broadcast is detected, it parses the payload and updates `self._name`, ensuring the cache never drifts from the physical controller's state.
- Updated `syrupy` snapshot tests to accept the new end-state names (see PR comments for detailed justification).

### Testing Performed:
- Ran `mypy --strict` to ensure complete type compliance.
- Ran `pytest` regression suite.
- Validated that the `test_gateway_replay_regression` snapshot correctly reflects the *final* chronological state of the system names rather than the historical first state.

Continuous CPU and I/O bloat, creating lag in downstream integrations and risking thread blocking on the SQLite database during high RF traffic periods.

### Risks of Implementing:

Very low. The primary risk is a malformed `0004` packet temporarily overwriting a zone name with garbage data, but the strict packet validation in the dispatcher heavily mitigates this.

### Mitigation Steps:

Payload structures for `Code._0004` (both list and dict formats) are safely type-checked (`isinstance`) before assignment to prevent runtime crashes from unexpected payload shapes.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/const.py
+++ b/src/ramses_rf/const.py
@@ -5,13 +5,22 @@ from __future__ import annotations
 
 from datetime import timedelta as td
 from enum import IntEnum
-from typing import TYPE_CHECKING, Final
+from typing import Final
 
-from ramses_tx.const import (  # noqa: F401
+from ramses_tx.const import (
     DEFAULT_MAX_ZONES as DEFAULT_MAX_ZONES,
+    DEV_ROLE_MAP as DEV_ROLE_MAP,
+    DEV_TYPE_MAP as DEV_TYPE_MAP,
     DEVICE_ID_REGEX as DEVICE_ID_REGEX,
     DOMAIN_TYPE_MAP as DOMAIN_TYPE_MAP,
+    F9 as F9,
+    FA as FA,
     FAN_MODE as FAN_MODE,  # deprecated, use SZ_FAN_MODE, to be removed in Q1 2026
+    FC as FC,
+    FF as FF,
+    I_ as I_,
+    RP as RP,
+    RQ as RQ,
     SYS_MODE_MAP as SYS_MODE_MAP,
     SZ_ACCEPT as SZ_ACCEPT,
     SZ_ACTUATORS as SZ_ACTUATORS,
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
**Context regarding the `test_regression_rf.py` snapshot updates:**

I wanted to provide a detailed explanation of why the snapshot for `test_gateway_replay_regression` required an update, as I know we are rightly cautious about changing snapshot outputs.

**The Root Cause:** The snapshot failure is a direct, positive result of the new caching mechanism correctly handling the progression of time within the regression log file.

**Old Behavior (The Bug):** Previously, when the snapshot test finished parsing the log and called `Zone.name()`, the code dynamically queried the database using `message_store.get(...)` and returned `msgs[0]`. Because `message_store.get()` returns historical packets, `msgs[0]` represented the **first/oldest** `0004` packet captured in that log for that zone.

**New Behavior (The Fix):** The new, highly efficient cache actively intercepts `0004` packets via `_handle_msg()` as they stream in. Every time a new `0004` packet arrives for a zone, `self._name` is updated. Therefore, when the snapshot test finishes its 45-second run and asks for the zone name, it receives the **last/newest** name recorded in the log.

**Why this matters:** The regression log file (`regression_packets_sorted.txt`) contains a long history of packets. Within that history, some zones were renamed over time (e.g., Zone 04 changing from `"Ed Office"` to `"Bathroom"`, and Zone 02 from `"Front room"` to `"Kitchen"`).

Our new cache accurately reflects the *current, final state* of the physical system at the end of the replay. The old code was accidentally locking onto a *historical start state*. Updating this snapshot proves that our digital twin is now perfectly synchronized with live physical hardware changes dynamically, without requiring a database query.

PWhite-Eng:
addresses PR2 in #580

silverailscolo (approved):
A lot of reformatting, and the `_name` fix, thanks

silverailscolo:
@PWhite-Eng Runs great on HVAC with all PRs up to now. Solid approach!
CPU 3% on average:
CPU HVAC 0 56 4+

PR #584, opened 2026-04-12
Refactor: Remove `__class__` mutation anti-pattern and strictly decouple transport/application messages

PWhite-Eng:
### The Problem:

The `ramses_tx` transport layer and `ramses_rf` application layer were tightly coupled via a dangerous runtime evaluation hack (`msg.__class__ = Message`) and arbitrary `setattr` injections (`_gwy`). This violated Object-Oriented design principles and bypassed static type checkers, creating a brittle architecture. Additionally, assigning `None` to the active `_msg_handler` during transport pauses created a race condition that could trigger a `TypeError: 'NoneType' object is not callable` crash in the asyncio event loop.

### Consequences:

Left unaddressed, the framework's core message routing relies on type obfuscation, making future feature expansions highly prone to silent attribute errors. Furthermore, the event loop race conditions during gateway pauses/resumes threaten the overall stability of the long-running process.

### The Fix:

Extracted all application-specific logic (`_gwy`, `_expired`) out of the base `ramses_tx.Message` and relocated it into a new, strictly typed `ApplicationMessage` subclass inside `ramses_rf.gateway`. Employed a factory pattern to safely promote incoming transport packets to domain-aware application messages at the architectural boundary, and implemented a thread-safe No-Op pattern for paused transport states.

### Technical Implementation:
- **Pure Transport:** Distilled `ramses_tx/message.py` down to a pure transport DTO and added a bound `TypeVar` to ensure generic classmethod factories (`_from_cmd`, `_from_pkt`) strictly respect subclass types.
- **Subclassing & Factories:** Created `ApplicationMessage(Message)` in `ramses_rf/gateway.py` with safe `_expired` calculation logic (including a 7-day fallback limit for unit testing).
- **Concurrency Safety:** Replaced `_msg_handler = None` with a No-Op `_drop_msg` pattern during transport pauses, acting as a safe drain for lingering in-flight packets. Added threading locks to the background task registry.
- **Liskov Substitution Principle (LSP):** Maintained LSP across `ramses_rf/device/*.py` and `entity_state.py` by keeping `msg: Message` in interface signatures while safely using `getattr(msg, "_expired", False)` internally, satisfying `mypy --strict`.

### Testing Performed:
- Ran the full `pytest` suite (~850+ tests), confirming 100% backward compatibility with existing mocked logs and packet replays.
- Added `test_pure_message_separation` to explicitly assert that the pure transport `Message` no longer contains application-layer properties (raising an `AttributeError`).
- Added `test_application_message_factory` to verify that `ApplicationMessage.from_message()` successfully duplicates transport data and accurately triggers the newly enforced 7-day hard expiration limit.

Continued technical debt, masking of type-related bugs, potential runtime crashes under heavy asynchronous load due to `NoneType` handler evaluations, and the inability to safely run strict static analysis on the core library.

### Risks of Implementing:

Subtle regressions in downstream message processing if an obscure, untested component strictly expects the old mutated base class structure instead of correctly interacting with the Liskov-compliant subclass (though extensive testing minimizes this).

### Mitigation Steps:

Enforced strict Mypy compliance (`mypy --strict` passes with 0 errors). Used the `getattr(msg, "_expired", False)` pattern to provide incredibly safe fallbacks for edge-case unit tests or deeply nested objects. Successfully validated the entire system against the extensive historical packet-log test suite.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/device/base.py
+++ b/src/ramses_rf/device/base.py
@@ -24,7 +24,7 @@ from ramses_rf.entity_base import Entity, class_by_attr
 from ramses_rf.exceptions import DeviceNotFaked, SchemaInconsistentError
 from ramses_rf.schemas import SZ_ALIAS, SZ_CLASS, SZ_FAKED, SZ_KNOWN_LIST
 from ramses_rf.topology import Child
-from ramses_tx import Command, Packet, Priority, QosParams
+from ramses_tx import Command, Message, Packet, Priority, QosParams
 from ramses_tx.ramses import CODES_BY_DEV_SLUG, CODES_ONLY_FROM_CTL
 from ramses_tx.typing import PayloadT
 
@@ -40,7 +40,7 @@ if TYPE_CHECKING:
     from ramses_rf import Gateway
     from ramses_rf.models import DeviceTraits
     from ramses_rf.system import Zone
-    from ramses_tx import Address, DeviceIdT, IndexT, Message
+    from ramses_tx import Address, DeviceIdT, IndexT
 
 
 BIND_WAITING_TIMEOUT = 300  # how long to wait, listening for an offer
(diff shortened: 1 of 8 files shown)


PWhite-Eng:
addresses PR3 in #580

silverailscolo (approved):
We must check by running the Ramses_cc tests to see if they will work with this update, or update there to match these improvements, thanks

PWhite-Eng:
Just ran the full pytest suite in ramses_cc (which has PR ramses-rf/ramses_cc#611 in it) and all test passed

PR #585, opened 2026-04-12
Refactor Dispatcher Pipeline & Implement Native Heartbeat Support

PWhite-Eng:
### The Problem:

The `process_msg` routing logic in `dispatcher.py` had grown into a monolithic "God function", making the lifecycle of a packet difficult to follow, maintain, and mock for unit testing. Furthermore, valid 1-byte `"00"` heartbeat payloads were failing strict frame validation, unnecessarily raising `PacketPayloadInvalid` exceptions. Finally, dynamic payload parsing was triggering `[no-any-return]` static typing errors under Mypy strict mode.

### Consequences:

Leaving the dispatcher as a monolith creates a massive bottleneck for future development and increases the cognitive load for developers trying to debug dropped packets. Furthermore, dropping valid 1-byte heartbeat packets due to strict regex validation creates blind spots in tracking device keep-alives and network health.

### The Fix:

Completely refactored the message dispatcher into a linear, mathematically isolated Chain of Responsibility pipeline. Simultaneously implemented safe interception of 1-byte heartbeat payloads across the packet validation and parser layers, ensuring they are cleanly processed without breaking legacy tests or the binding FSM.

### Technical Implementation:
- **Pipeline Architecture:** Split `dispatcher.py`'s `process_msg` into four discrete, sequential stages: `validate_addresses`, `instantiate_devices`, `validate_slugs`, and `route_payload`. Each stage acts as a boolean gate.
- **Heartbeat Interception (`packet.py`):** Wrapped the `super()._validate()` call in a `try/except` block to specifically catch and suppress `PacketPayloadInvalid` for payloads measuring exactly 1 byte and evaluating to `"00"`.
- **Parser Routing (`parsers.py`):** Added a dedicated `parser_heartbeat` function. Implemented logic in `parse_payload` to divert `"00"` payloads to this parser, explicitly excluding `Code._1FC9` to protect the binding state machine.
- **Type Guarding:** Implemented `isinstance()` checks at the tail of `parse_payload` to narrow dynamic dictionary generation, satisfying Mypy strict mode.

### Testing Performed:
- **`test_dispatcher.py`**: Updated to mock the new pipeline architecture. Added a `TestDispatcherHeartbeats` class to verify `"00"` payloads correctly reach the target device's `_handle_msg` queue.
- **`test_packet.py`**: Wrote `test_packet_heartbeat_payload_bypass` to explicitly assert that a `"00"` payload bypasses regex crash validation but maintains `_has_payload = False` for legacy test safety.
- **`test_parsers.py`**: Verified that returning `{'heartbeat': True}` successfully falls back to `{}` for legacy logs where necessary, yielding a 100% pass rate.

The dispatcher logic remains fragile and impossible to comprehensively unit-test. Heartbeat packets continue to be flagged as invalid protocol violations, degrading the system's ability to monitor device presence accurately.

### Risks of Implementing:

Because the pipeline enforces strict, sequential validation, if an edge-case protocol quirk relies on bypassing a slug check or address check in an undocumented way, the packet might be dropped prematurely (false positive validation failures).

### Mitigation Steps:

Maintained absolute feature parity with the previous monolithic dispatcher—no underlying validation logic was altered, only structurally isolated. Ensured the entire `pytest` suite passes with zero regressions. Explicitly bypassed heartbeat interception for `1FC9` packets to ensure the pairing/binding process remains completely unaffected.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/dispatcher.py
+++ b/src/ramses_rf/dispatcher.py
@@ -48,7 +48,14 @@ _DBG_INCREASE_LOG_LEVELS: Final[bool] = (
 _LOGGER = logging.getLogger(__name__)
 
 
-__all__ = ["detect_array_fragment", "process_msg"]
+__all__ = [
+    "detect_array_fragment",
+    "instantiate_devices",
+    "process_msg",
+    "route_payload",
+    "validate_addresses",
+    "validate_slugs",
+]
 
 
 MSG_FORMAT_18 = "|| {:18s} | {:18s} | {:2s} | {:16s} | {:^4s} || {}"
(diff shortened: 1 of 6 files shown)


PWhite-Eng:
addresses PR4 in #580

PWhite-Eng:
Just ran the full pytest suite in ramses_cc (which has PR https://github.com/ramses-rf/ramses_cc/pull/611 in it) and all test passed

PWhite-Eng:
@silverailscolo, please note that if you see deleted TODO/FIXME/HACK, this is intentional as we fixed them in this refactor, those that have not been solved have been left (I think one or two of them may have been relocated with the altered code)

silverailscolo (approved):

PR #586, opened 2026-04-12
Refactor MessageStore CQRS architecture to resolve asyncio event loop deadlocks

PWhite-Eng:
### The Problem:

The main `asyncio` event loop was stalling due to blocking synchronous database operations and `threading.Lock()` contention between the main application thread and the background `SQLiteWorker`. Furthermore, forceful CLI shutdowns (e.g., `Ctrl+C`) were instantly terminating the background worker before it could safely flush the pending write queue to disk. Sharing a single connection across threads was also violating Python 3.13's strict SQLite thread-safety rules, risking C-level segmentation faults.

### Consequences:

Left unfixed, the application would suffer from severe unresponsiveness (event loop freezing) during high-volume RF packet processing. Shutting down the application resulted in zombie threads or 15-minute deadlocks. Additionally, sudden terminations posed a severe risk of data loss due to dropped disk snapshots.

### The Fix:

Migrated the database logic to a strict Command Query Responsibility Segregation (CQRS) architecture. Read operations now instantly hit the in-memory RAM dictionaries, while all database writes and deletes are offloaded to a non-blocking `SimpleQueue` handled by the background worker. We removed manual Python thread locks, shifting concurrency management entirely to SQLite's native Write-Ahead Logging (WAL) mechanism. Finally, we added a graceful 7-second buffer to the application teardown sequence.

### Technical Implementation:
- **`message_store.py`**: Surgically removed `self._db_lock = threading.Lock()`. Rerouted blocking `DELETE` operations in `clr()` and `INSERT` operations in `add()` directly to the non-blocking `SQLiteWorker` queue. Fixed Mypy `[type-arg]` warnings and localized `UnboundLocalError` scoping bugs.
- **`sqlite_worker.py`**: Fully implemented non-blocking `queue.put()` handlers for payloads, prunes, and disk snapshots.
- **`client.py`**: Wrapped the teardown `gwy.stop()` call in `asyncio.wait_for(..., timeout=7.0)` to ensure the OS does not kill the Python process while the DB thread is finishing its final disk flush.
- **`gateway.py`**: Removed dangerous cross-thread cancellation logic (`asyncio.to_thread(self._message_store.stop)`) that previously corrupted the event loop state machine during teardown.
- **`test_client.py`**: Removed a flawed `asyncio.wait_for` mock in `test_async_main_real_gateway_init` that was swallowing coroutines and throwing `RuntimeWarning`s for un-awaited mocks.

### Testing Performed:

Executed the full Pytest suite to verify the new architecture. Confirmed that the `PYTEST_CURRENT_TEST` environment variable bypass correctly isolates the background worker to ensure synchronous test speeds. Fixed and successfully executed `tests/tests_cli/test_client.py::test_async_main_real_gateway_init` to validate the new graceful shutdown lifecycle. Ran and passed 100% strict Mypy type checking and Ruff formatting.

The system will continue to experience random event loop hangs, high latency on RF message processing, lost database writes on application exit, and hard segfaults on modern Python 3.13 installations.

### Risks of Implementing:

If the database disk is exceptionally slow, a heavy backlog of queued write operations might take longer than 7 seconds to flush during shutdown, which could still result in partial data loss if the timeout expires.

### Mitigation Steps:

Implemented a deterministic 7.0-second timeout on `gwy.stop()` using `asyncio.wait_for`. Handled `TimeoutError` cleanly in the `client.py` exception block. Retained explicit fallback `sqlite3.Error` handling in the worker to ensure corrupt payloads do not crash the background thread.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_cli/client.py
+++ b/src/ramses_cli/client.py
@@ -718,10 +718,10 @@ async def async_main(command: str, lib_kwargs: dict[str, Any], **kwargs: Any) ->
     else:  # if no Exceptions raised, e.g. EOF when parsing, or Ctrl-C?
         msg = "ended without error (e.g. EOF)"
     finally:
-        with contextlib.suppress(asyncio.CancelledError):
-            await (
-                gwy.stop()
-            )  # Task cancellation is expected during shutdown, especially after EOF
+        with contextlib.suppress(asyncio.CancelledError, TimeoutError):
+            # Task cancellation is expected during shutdown, especially
+            # after EOF
+            await asyncio.wait_for(gwy.stop(), timeout=7.0)
 
     print(f"\r\nclient.py: Engine stopped: {msg}")
 
--- a/src/ramses_rf/message_store.py
+++ b/src/ramses_rf/message_store.py
@@ -13,12 +13,13 @@ RAMSES RF - Message database and index.
    i7     get_rp_codes  src, dst     list(Code)        Discovery-supported_cmds
    =====  ============  ===========  ==========  ====  ========================
 
-[#fn1] A word of explanation.[^1]: This table documents the primary methods
-used by external components (like `EntityState`) to query the central
-message store. As of Phase 2.5, legacy SQL-based query methods
+[#fn1] A word of explanation.[^1]: This table documents the primary
+methods used by external components (like `EntityState`) to query the
+central message store. As of Phase 2.5, legacy SQL-based query methods
 (`qry`, `qry_field`, `_select_from`) have been removed. The system now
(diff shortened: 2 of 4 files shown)


PWhite-Eng:
addresses PR5 in #580

PWhite-Eng:
Just ran the full pytest suite in ramses_cc (which has PR https://github.com/ramses-rf/ramses_cc/pull/611 in it) and all test passed

silverailscolo (approved):
Makes SQLite solidly async

PR #590, opened 2026-04-12
Refactor Entity State caching to use strictly-typed `StateCache` object

PWhite-Eng:
### The Problem:

The `EntityState` component previously relied on a deeply nested, 3-dimensional dictionary (`dict[Code, dict[VerbT, dict[bool | str | None, ApplicationMessage]]]`) exposed via `get_state_cache_nested()`. This complex data structure leaked into external modules (`base.py`, `discovery.py`, etc.), breaking encapsulation, violating the Law of Demeter, and preventing the codebase from achieving strict type safety.

### Consequences:

Continuing to use the nested dictionary makes the codebase harder to maintain and difficult to type-hint correctly (causing Mypy strict failures). It also leaves the system prone to runtime `KeyError`s or `AttributeError`s when other modules try to navigate the complex, unregulated dict structure directly.

### The Fix:

Replaced the nested dictionary with a dedicated `StateCache` class. This class encapsulates the memory storage internally and exposes clean, strictly-typed methods for querying messages. External modules were updated to use public, flat-data APIs rather than traversing the internal cache structure.

### Technical Implementation:
- Created the `StateCache` class in `entity_state.py` with strict access methods: `get_message()`, `get_by_code()`, `get_all()`, and `get_records()`.
- Converted `get_state_cache_nested()` into a private `_build_state_cache()` method that exclusively returns the new `StateCache` object.
- Refactored internal `EntityState` methods (`_msg_dev_qry`, `find_latest_code`, `get_message_log_flat`) to iterate over the new `cache.get_records()` implementation.
- Updated `device/base.py` to query `get_message_log_flat()` instead of reaching into the raw nested cache.
- Updated `discovery.py` to utilize `get_all_messages()` for OpenTherm context extraction.
- Resolved Ruff `B007` warnings by marking unused loop control variables (e.g., `_ctx`).

### Testing Performed:
- Updated `test_entity_base.py` to instantiate and test against the new `StateCache` class API.
- Modified legacy dictionary structure assertions (e.g., `assert cache == {...}`) to safely use object method queries (e.g., `assert cache.get_message(...) == mock_msg`).
- Verified that the legacy OpenTherm context tests (`test_gh_396_legacy_ot_context`) continue to pass securely using mocked `StateCache` objects.
- Executed the full Pytest suite with a 100% pass rate.
- Verified zero typing errors across 138 source files using `mypy --strict`.

Leaving the nested dictionary structure in place perpetuates technical debt, prevents the codebase from achieving strict type compliance, and makes future refactoring of the central `MessageStore` highly brittle due to tightly coupled module dependencies.

### Risks of Implementing:

Upstream modules that relied heavily on the exact dictionary structure of the old cache might fail if they were missed during the refactor. There is also a slight risk of altered iteration order since the underlying internal storage changed to a tuple-keyed flat dictionary.

### Mitigation Steps:
- Strictly enforced `mypy --strict` to guarantee no hidden or legacy references to the old `get_state_cache_nested` method exist anywhere in the codebase.
- Maintained full test coverage passing 100% in Pytest.
- Transitioned external modules to heavily tested existing `EntityState` public methods (`get_all_messages`, `get_message_log_flat`) to ensure behavioral parity without coupling them to the new internal cache object.

### AI Assistance Disclosure:


PWhite-Eng:
addresses PR8 in #580

PWhite-Eng:

silverailscolo (approved):

PR #592, opened 2026-04-12
Refactor Command Address Mutation to Immutable Cloning & Fix Strict Mypy Issues

PWhite-Eng:
### The Problem:

The protocol layer (`base.py`) was directly mutating private, strictly-typed properties (`cmd._hdr_`, `cmd._addrs`) of the `Command` object when patching the active HGI address for `evofw3` devices. Additionally, minor technical debt items were lingering in the `Command` class, including a misleading `TODO` regarding `__str__` serialization.

### Consequences:

Direct mutation of an already-instantiated `Command` object violated the read-only properties of the parent `Frame` class and broke strict typing constraints (`mypy --strict`), specifically regarding `HeaderT` and `tuple` vs `list` assignments. Furthermore, attempting to force in-place mutations and altering the `__str__` return value broke the FSM transport serialization, resulting in severe `ProtocolTimeoutError` crashes across the entire virtual RF test suite.

### The Fix:

Transitioned the `Command` modification paradigm from "in-place mutation" to "immutable cloning." Replaced direct property hacking with a new, type-safe `clone_with_source()` method, and cleaned up the identified technical debt in `command.py`.

### Technical Implementation:
- Implemented `clone_with_source(self, new_src: DeviceIdT | str) -> Command` in `command.py`. This method reconstructs the raw frame string and re-runs the constructor (`type(self)(new_frame)`), ensuring mathematically perfect recalculation of `_hdr`, `_addrs`, and `_rx_header`.
- Updated `_patch_cmd_if_needed` in `base.py` to invoke this new cloning method instead of manually overriding object dictionaries.
- Reverted an erroneous `TODO` on `Command.__str__` that was breaking FSM transport serialization by outputting the header instead of the raw frame.
- Cleaned up dead code (`_source_entity`) and removed completely misleading comments.

### Testing Performed:
- Added `test_clone_with_source` to `test_command.py` to verify the cloned object is a distinct instance with updated addresses and fully recalculated headers, whilst confirming the original command remains completely untouched.
- Added `test_patch_cmd_if_needed_evofw3` to `test_protocol_base.py` to ensure the protocol correctly intercepts and replaces the default `HGI_DEV_ADDR` with the active HGI ID when `evofw3` is in use.
- Ran the full `pytest` suite (including `test_binding_fsm.py` and `test_protocol_fsm.py`), ensuring 0 timeouts and a 100% pass rate. Verified 100% compliance with `mypy --strict`.

The codebase would remain non-compliant with `mypy --strict`. Future protocol extensions would risk severe bugs or serialization timeouts if developers attempt to modify parsed, immutable packet frames in-place.

### Risks of Implementing:

Extremely low. Because we are creating entirely new instances of `Command` rather than modifying existing ones, downstream consumers holding references to the original command will not see unexpected mid-flight state changes (which is the desired, thread-safe behavior).

### Mitigation Steps:

Extensive unit testing of both the isolated `Command` clone logic and the actual protocol dispatch layer. Ran the full asynchronous FSM and virtual RF test suites to definitively ensure the newly instantiated cloned objects do not disrupt the transmit queue or transport layer serialization.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_tx/command.py
+++ b/src/ramses_tx/command.py
@@ -337,7 +337,6 @@ class Command(Frame):
             _LOGGER.warning(f"{self} < Command is potentially invalid: {err}")
 
         self._rx_header: HeaderT | None = None
-        # self._source_entity: Entity | None = None  # TODO: is needed?
 
     @classmethod  # convenience constructor
     def from_attrs(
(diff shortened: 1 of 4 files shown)


PWhite-Eng:
addresses PR11in #580

PWhite-Eng:

silverailscolo (approved):

PR #593, opened 2026-04-12
Refactor Zigbee transport to use explicit exception handling and routing

PWhite-Eng:
### The Problem:

The `zigbee.py` transport module previously relied on broad `except Exception:` and `except Exception as err:` clauses. This violates Python PEP 8 best practices and creates a dangerous environment in asynchronous event-driven applications, as it can silently swallow critical system faults (e.g., `KeyboardInterrupt`, `SystemExit`) or obscure unexpected underlying framework errors.

### Consequences:

If left unfixed, the system is vulnerable to silent failures. Unrelated background errors or critical event-loop crashes could be swallowed, resulting in silent hangs without any corresponding error logs. It also makes debugging third-party library failures significantly harder due to missing stack traces.

### The Fix:

We refactored the transport exception handling to strictly filter for expected runtime errors. Broad catches were replaced with explicit exception tuples for internal callbacks. Where broad catches were strictly required for opaque third-party library calls, the exceptions were wrapped and bubbled up as domain-specific transport errors with full logging.

### Technical Implementation:
- **Explicit Filtering:** Updated standard execution paths and callbacks (`attribute_updated`, `cluster_command`) to catch a specific tuple of expected errors: `(KeyError, ValueError, TypeError, AttributeError, RuntimeError, exc.RamsesException)`.
- **Safe Error Wrapping:** Maintained broad `except Exception as err:` clauses *only* around unpredictable `zigpy`/`zha` I/O triggers, explicitly wrapping them with `raise exc.TransportZigbeeError(...) from err` to preserve the traceback chain.
- **Stack Trace Preservation:** Ensured all broad error catches utilize `_LOGGER.exception(...)` instead of `_LOGGER.debug` or `_LOGGER.warning`.
- **Test Implementation:** Added `TestExceptionPropagation` utilizing a dynamically generated `CustomUnhandledError` to prove that exceptions outside the explicit whitelist propagate correctly out of the callback.

### Testing Performed:
- Validated the full `pytest` suite locally, achieving a 100% pass rate.
- Wrote `test_unhandled_exception_propagates_out_of_callback` using a custom exception class to verify that the new explicit exception tuples correctly allow non-whitelisted errors to escape natively.
- Verified 100% compliance with `mypy --strict` and Ruff linting standards.

Leaving the code as is means the transport layer remains brittle. Future bugs introduced in the `zigpy` stack or internal logic might be silently swallowed, leading to "ghost" bugs in production that are incredibly difficult to reproduce and diagnose.

### Risks of Implementing:

Removing a catch-all `except Exception:` always carries the risk that a previously swallowed (but functionally harmless) error originating from a third-party library might escape the try block and inadvertently crash the callback dispatcher or the active event loop.

### Mitigation Steps:

To mitigate the risk of unexpected crashes, we included `RuntimeError` and our internal base `exc.RamsesException` in the explicit whitelist. Furthermore, at the direct I/O boundary (where unpredictable third-party errors are most likely to occur), we maintained the broad catch but converted it into a safe, domain-specific `TransportZigbeeError` to ensure safe propagation without destabilizing the application.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_tx/command.py
+++ b/src/ramses_tx/command.py
@@ -337,7 +337,6 @@ class Command(Frame):
             _LOGGER.warning(f"{self} < Command is potentially invalid: {err}")
 
         self._rx_header: HeaderT | None = None
-        # self._source_entity: Entity | None = None  # TODO: is needed?
 
     @classmethod  # convenience constructor
     def from_attrs(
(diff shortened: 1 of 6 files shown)


PWhite-Eng:
addresses PR12 in #580

PWhite-Eng:

silverailscolo (approved):

PR #594, opened 2026-04-12
Refactor payload parsing to Chain of Responsibility and remove packet validation hacks

PWhite-Eng:
### The Problem:

`Packet._validate` relied on a hardcoded, structural bypass hack to permit 1-byte '00' heartbeat payloads, violating separation of concerns. Furthermore, payload parsing logic in `Message` and `parsers.py` relied on monolithic, procedural functions (`parse_payload`, `_check_msg_payload`) that heavily entangled regex schema validation, null-payload checking, and device parsing logic, making the code difficult to extend and maintain.

### Consequences:

Leaving the parsing logic entangled makes it error-prone and brittle when adding new RAMSES RF protocol rules or handling specific device quirks. Hardcoded bypasses at the lower `Packet` layer bleed application-level context into network-level framing, increasing the risk of malformed packets slipping through baseline validation checks.

### The Fix:

Migrated the payload validation and parsing sequence into a highly typed, structured Chain of Responsibility pipeline, and removed the frame-level bypass hacks from the `Packet` object.

### Technical Implementation:
- **`packet.py`:** Removed the `try/except PacketPayloadInvalid` bypass block targeting `parts[-1] == "00"`, restoring pure frame-level validation.
- **`parsers.py`:** Introduced an abstract `PayloadDecoder` class. Implemented the logic into sequential chain links: `RegexValidatorDecoder` (handles schema validation and early exits for null payloads), `HeartbeatDecoder` (intercepts 1-byte heartbeats), and `StandardParserDecoder` (routes to specific 4-digit code parsers).
- **`parsers.py`:** Wired the chain together via the `PayloadDecoderPipeline` class.
- **`message.py`:** Replaced legacy procedural calls with the invocation of `PayloadDecoderPipeline().decode(self)`.
- **`tests_tx/test_message.py`:** Refactored the `patch_parsers` fixture to mock `PayloadDecoderPipeline.decode` directly, resolving `AttributeError` exceptions and restoring isolated test coverage.

### Testing Performed:

Ran the complete data-driven log parsing suite (`test_parsers.py`) against thousands of real-world historical packets to guarantee 100% behavioral parity with the old monolithic logic. Executed the full `pytest` suite ensuring all mocked unit tests, integration tests, and packet evaluations pass flawlessly. Verified complete strict type compliance via `mypy --strict`.

Technical debt will continue to accumulate. The `Packet` layer remains improperly coupled to payload-specific logic, and the central parsing monolith will become increasingly fragile as new device fingerprints, Edge cases, and payload schemas are introduced.

### Risks of Implementing:

A logic error in the pipeline sequence (e.g., yielding a value too early or failing to yield `None` when bypassing) could cause valid payloads to be ignored, or trigger false-positive `PacketPayloadInvalid` exceptions, potentially leading to dropped messages across the gateway.

### Mitigation Steps:

The pipeline order was carefully sequenced (Regex Validation -> Heartbeats -> Standard Parsers) to ensure malformed payloads are safely rejected before any heartbeat logic is applied. The continued success of the extensive log-driven `pytest` suite proves that the new architectural flow perfectly mimics the outputs and edge-case handling of the original codebase.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_tx/message.py
+++ b/src/ramses_tx/message.py
@@ -14,8 +14,8 @@ from .address import Address
 from .command import Command
 from .const import DEV_TYPE_MAP, SZ_DHW_IDX, SZ_DOMAIN_ID, SZ_UFH_IDX, SZ_ZONE_IDX
 from .packet import Packet
-from .parsers import parse_payload
-from .ramses import CODE_IDX_ARE_COMPLEX, CODES_SCHEMA, RQ_IDX_COMPLEX
+from .parsers import PayloadDecoderPipeline
+from .ramses import CODE_IDX_ARE_COMPLEX, CODES_SCHEMA
 
 # TODO:
 # long-format msg.__str__ - alias columns don't line up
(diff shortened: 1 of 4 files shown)


PWhite-Eng:
addresses PR13 in #580

PWhite-Eng:
Just ran the full pytest suite in ramses_cc (which has PR https://github.com/ramses-rf/ramses_cc/pull/618 in it) and all test passed

silverailscolo (approved):

Issue #583, opened 2026-04-11
Chore: Apply max. line length, reformat code


silverailscolo:
Currently 4k lines do no pass an industry standard rule for line length <72 chars

- medium priority (improves maintainability)
- use ruff format
- apply as prek hook:

In `pyproject.toml`
Change this section:
```
[tool.ruff.lint]
  select = [
    "ASYNC",  # flake8-async
    "B",  #     flake8-bugbear
    "E",  #     pycodestyle
    "F",  #     Pyflakes
    "I",  #     isort
    "SIM",  #   flake8-simplify
    "UP",  #    pyupgrade
  ]
  ignore = ["ASYNC109", "ASYNC110", "B011", "SIM102", "SIM114", "UP040"]

  # B011   - Do not call assert False since python -O removes these calls
  # E501   - Line too long (Note: Removed from ignore list to enable check)
  # SIM102 - Use a single `if` statement instead of nested `if` statements
  # SIM114 - Combine `if` branches using logical `or` operator
  # UP040  - Type alias uses `TypeAlias` annotation instead of the `type` keyword
```
The change is in the ignore list, you need to remove E501

Just below the above section, add:
```
[tool.ruff.format]
  # Enforce PEP 8 docstring and comment wrapping
  docstring-code-format = true


[tool.ruff.lint.pycodestyle]
  # PEP 8 recommends 72 characters for docstrings and comments
  max-doc-length = 72
```

_Originally posted by @PWhite-Eng in https://github.com/ramses-rf/ramses_rf/pull/581#issuecomment-4229731248_

silverailscolo:
79 char limit enforced in PR #1065 , closing

PR #602, opened 2026-04-13
Fix System Mode initializing as Null on startup (#583)

PWhite-Eng:
### The Problem:

As reported in Issue #583, the Evohome controller initializes its `system_mode` as `null` after a Home Assistant restart. The integration currently fails to proactively fetch the controller's operating mode upon initialization, forcing Home Assistant to wait indefinitely for the physical panel to spontaneously broadcast a mode change.

### Consequences:

If this isn't fixed, Home Assistant loses state synchronization every time it restarts. Any automations relying on the system mode (e.g., triggering actions based on 'Away' or 'Auto' modes) will silently fail or behave unpredictably until a user physically interacts with the controller to force a broadcast.

### The Fix:

Instructed the `Controller` entity to proactively poll for its system mode immediately upon initialization, and introduced an exponential backoff to the discovery service to guarantee this startup request succeeds even during heavy RF traffic.

### Technical Implementation:
- **`heat.py` (Controller):** Overrode `_setup_discovery_cmds` in the `Controller` class to actively queue an `RQ` packet for `Code._2E04` (System Mode). This forces a state fetch at startup, backed by a 60-minute repeating safety loop in case future broadcasts are dropped.
- **`discovery.py`:** Replaced the disabled debug-flagged backoff logic with a production-ready exponential timer (1m, 4m, 7.5m, 15m, 30m, 60m). This ensures that if the initial `2E04` startup query is lost to radio collisions, it gently paces RF retries until it successfully retrieves the state.

### Testing Performed:
- Added a dedicated async test (`test_controller_discovers_system_mode`) to `tests/tests_rf/test_device_heat.py`.
- The test explicitly triggers the discovery setup phase in isolation and asserts that the `2E04` packet is correctly queued into the `DiscoveryService`.
- Full `pytest` suite passes with 100% success locally.
- Verified strict Python 3.10+ typing compliance (`mypy --strict`) and Ruff linting passed with zero errors across the modified files.

The Home Assistant integration will remain unreliable post-restart, continually forcing users to manually intervene with their physical thermostat panels just to sync their smart home state.

### Risks of Implementing:

The new exponential discovery backoff logic could slightly increase RF traffic for the first hour if the controller is completely offline, powered down, or dead during the HA startup sequence.

### Mitigation Steps:

The exponential backoff in `discovery.py` is strictly capped at a 1-hour maximum and includes a `min()` safety constraint. This guarantees the retry pacing safely limits itself and never exceeds the standard configured polling interval (60 minutes) established by the `Controller`.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/device/heat.py
+++ b/src/ramses_rf/device/heat.py
@@ -367,6 +367,15 @@ class Controller(DeviceHeat):  # CTL (01):
         self.tcs = None  # TODO: = self?
         self._make_tcs_controller(**kwargs)  # NOTE: must create_from_schema first
 
+    def _setup_discovery_cmds(self) -> None:
+        super()._setup_discovery_cmds()
+
+        if not self.is_faked:
+            self.discovery.add_cmd(
+                Command.from_attrs(RQ, self.id, Code._2E04, PayloadT("00")),
+                60 * 60,  # Poll every 60 minutes after initial startup query
+            )
+
     def _handle_msg(self, msg: Message) -> None:
         super()._handle_msg(msg)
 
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
@silverailscolo, please note this is a stacked PR on top of #601
Thanks
Phil

silverailscolo (approved):

PR #604, opened 2026-04-14
Fix: Resolve asyncio task leaks and unhandled FSM protocol timeouts

PWhite-Eng:
### The Problem:

There were two critical issues regarding asynchronous exception handling. First, `_get_temp` in `zones.py` lacked an exception handler for `ProtocolTimeoutError`, causing the main asyncio task runner to crash when RF timeouts occurred. Second, in `fsm.py`, unexpected exceptions (e.g., `RuntimeError`) within the `_send_cmd` background task were being silently swallowed by asyncio because the `try/except` block only caught `TransportError`. This left futures hanging unresolved and the FSM stuck permanently in the `WantEcho` state.

### Consequences:

If left unfixed, these unhandled exceptions either crash the event loop or silently stall the FSM. This leads to severe CPU leaks, permanently stalled background polling, and a completely frozen state for the end-user, requiring manual system reboots to recover.

### The Fix:

Implemented graceful degradation for zone temperature polling and established a proper "Task Exception Gateway" in the protocol FSM to ensure all background task failures are correctly routed back to the main event loop to resolve pending futures.

### Technical Implementation:
- **`src/ramses_rf/system/zones.py`**: Wrapped the `await self._gwy.async_send_cmd(...)` call inside `_get_temp` with a `try/except` block that specifically catches `ramses_tx.exceptions.ProtocolTimeoutError`. It now logs a warning and safely returns `None` instead of crashing.
- **`src/ramses_tx/protocol/fsm.py`**: Broadened the exception handler in the `send_fnc_wrapper` task inside `_send_cmd` from `except TransportError as err:` to `except Exception as err:`. This ensures *any* unexpected failure correctly transitions the FSM back to the `IsInIdle` state and injects the exception into the pending future to unblock the queue.

### Testing Performed:

Fully strictly-typed, Ruff-compliant Pytest suites were written using the TDD (Red-Green) methodology:
- **`tests/tests_rf/test_system_zones.py`**: Added `test_zone_get_temp_handles_protocol_timeout` which mocks a 20-second RF timeout to guarantee `_get_temp` absorbs the error gracefully and completes the task.
- **`tests/tests_tx/test_protocol_fsm.py`**: Added `test_fsm_unhandled_exception_in_send_task` which injects a `RuntimeError` into the transport layer. This test mathematically proves that the FSM no longer hangs until the global timeout expires, but instead immediately rejects the future and cleanly resets the FSM to `IsInIdle`.

The application will continue to suffer from fatal CPU leaks and silent FSM stalls in production environments, particularly in deployments with poor RF connectivity where transport delays and timeouts are frequent occurrences.

### Risks of Implementing:

Broadening the exception handler in `fsm.py` to `except Exception` means that unexpected system-level or syntax errors within the transport function will now be routed through the FSM state machine rather than crashing the task immediately. If the caller does not monitor the rejected futures properly, deeply rooted architectural errors could theoretically be overlooked.

### Mitigation Steps:

The risk of masking errors is mitigated because the FSM explicitly injects the caught exception into the active future (`self._qos_mgr.fut.set_exception(exception)`). This ensures the error bubbles up cleanly to the caller's `await` point, preserving the traceback and failing loudly where appropriate, rather than failing silently in the background. Full regression tests were also added to ensure standard `TransportError` behavior remains unaffected.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_tx/protocol/fsm.py
+++ b/src/ramses_tx/protocol/fsm.py
@@ -374,7 +374,7 @@ class ProtocolContext(StateMachineInterface):
 
             try:
                 await self._send_fnc(cmd)
-            except TransportError as err:
+            except Exception as err:
                 self.set_state(IsInIdle, exception=err)
 
         try:
--- a/src/ramses_rf/system/zones.py
+++ b/src/ramses_rf/system/zones.py
@@ -49,6 +49,7 @@ from ramses_rf.schemas import (
 )
 from ramses_rf.topology import Child, Parent
 from ramses_tx import Address, Command, Message, Priority
+from ramses_tx.exceptions import ProtocolTimeoutError
 from ramses_tx.typing import HeaderT, PayDictT, PayloadT
 
 from .schedule import InnerScheduleT, OuterScheduleT, Schedule
@@ -892,9 +893,13 @@ class Zone(ZoneSchedule):
 
     async def _get_temp(self) -> Packet | None:
         """Get the zone's latest temp from the Controller."""
-        return await self._gwy.async_send_cmd(
-            Command.get_zone_temp(self.ctl.id, self.idx)
-        )
+        try:
+            return await self._gwy.async_send_cmd(
+                Command.get_zone_temp(self.ctl.id, self.idx)
+            )
+        except ProtocolTimeoutError as err:
+            _LOGGER.warning(f"{self}: _get_temp timed out: {err}")
+            return None
 
     async def reset_config(self) -> Packet:  # 000A
         """Reset the zone's parameters to their default values."""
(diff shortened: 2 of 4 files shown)


src/ramses_tx/protocol/fsm.py
```diff
@@ -374,7 +374,7 @@ def is_imminent(p: Packet) -> bool:
 
             try:
                 await self._send_fnc(cmd)
-            except TransportError as err:
+            except Exception as err:
```
silverailscolo, line 377: You might add a comment on this broad catch

silverailscolo (approved):

PR #611, opened 2026-04-17
Refactor: Resolve Priority 1 Transport Tech Debt, QoS Modernization, and Split Command God Object (Fixes #591)

### LINKED ISSUE #591: Refactor Technical Debt in QoS, Transport bindings, and FSM Logic
**Description:**
As part of our ongoing effort to stabilize the protocol layer and reduce tech debt, we've identified several areas related to the QoS configuration and transport FSM bindings that require structural updates. Leaving these issues unaddressed prevents us from rolling out the newer, more robust QoS mechanism and poses risks of silent failures or dropped packets during edge-case transport disconnects.

#### 1. FSM Early Write Binding (`F-01`)
* **Affected File:** `src/ramses_tx/protocol/base.py`
* **Location:** Line ~64 (`self._pause_writing = False`)
* **Description:** The base protocol initializes in a Read/Write state (`False`) *before* the transport connection is fully confirmed or bound.
* **Risk/Effects:** If a command is pushed to the protocol layer before `connection_made()` has fully established the pipe, attempting to write to the transport could result in dropped packets or unhandled internal buffer errors.
* **Proposed Fix:** Initialize `self._pause_writing = True`. The `connection_made()` method (or the attached FSM context) should then appropriately call `resume_writing()` to transition the state only once the transport is ready.

#### 2. Missing Return Types in FSM Transmit (`H-01`)
* **Affected File:** `src/ramses_tx/protocol/base.py`
* **Location:** Line ~288 (`if not pkt: # HACK: temporary workaround...`)
* **Description:** The internal `_send_cmd` coroutine occasionally returns `None` (e.g., if QoS is disabled or it silently fails during an internal timeout), forcing a hacky manual exception check in the parent method.
* **Risk/Effects:** Violates strict typing contracts (`-> Packet`) and creates unpredictable control flow. If the underlying FSM quietly swallows transmit failures, diagnosing dropped commands becomes incredibly difficult.
* **Proposed Fix:** Refactor `fsm.py` and the `_send_cmd` contract so they *always* return a valid `Packet` on success, or explicitly raise a typed exception (such as `ProtocolSendFailed`) on failure. Remove the manual `None` check from `base.py`.

#### 3. Outdated QoS Object Architecture (`T-01` & `T-02`)
* **Affected File:** `src/ramses_tx/command.py`
* **Location:** Lines ~92 (`# TODO: this needs work`), ~115-118 (`# TODO: deprecate`)
* **Description:** The `Qos` class heavily relies on an outdated architectural pattern, and its initialization arguments (priority, max_retries, timeout, backoff) are slated for deprecation.
* **Risk/Effects:** Creating new technical debt around QoS. As the FSM evolves, having multiple modules rely on legacy QoS class signatures creates brittle dependencies and slows down FSM improvements.
* **Proposed Fix:** Redesign the `Qos` class as a modern, strict configuration object (e.g., a frozen Python 3.10 `dataclass`). Remove the deprecated parameters and update all internal instantiations across `command.py`, `fsm.py`, and `core.py` to use the new pattern.

---

### Prompt for Gemini to Undertake the Work
*You can use the prompt below when you are ready to assign this ticket to Gemini:*

> "I need you to act as a Senior Python Developer. I am providing you with the source code for `base.py`, `command.py`, `core.py`, `fsm.py`, and `typing.py` from the `ramses_tx` project.
> 
> We need to resolve four interconnected technical debt issues regarding FSM logic and QoS structure:
> 1. In `base.py`, change the initialization of `self._pause_writing` to `True` to prevent early write binding before the transport is confirmed.
> 2. Redesign the `Qos` class in `command.py` into a modern frozen dataclass. Strip out the deprecated init arguments (`priority`, `max_retries`, `timeout`, `backoff`) and implement the new pattern.
> 3. Refactor the typed signature and logic of `_send_cmd` in both `base.py` and `fsm.py` so that it is guaranteed to return a `Packet` object, rather than `None`. If a packet cannot be returned, it must raise a `ProtocolSendFailed` exception internally.
> 4. Remove the `if not pkt: # HACK` workaround from `base.py`.
> 
> Please provide the fully refactored, Ruff/Mypy-strict compliant code for all modified files. Ensure all internal class usages of `Qos` are updated to match the new dataclass signature."

PWhite-Eng:
### The Problem:

This PR addresses several interconnected areas of technical debt outlined in Issue #591 that hinder protocol stability. Specifically:
1. **FSM Early Write Binding:** The base protocol was initializing its transport pipe in a Read/Write state (`_pause_writing = False`) before `connection_made()` had fully established the connection.
2. **Loose Transmit Contracts:** FSM transmit sequences occasionally swallowed failures and returned `None` instead of raising typed exceptions, necessitating hacky workarounds downstream.
3. **Outdated QoS Architecture:** The `Qos` configuration object was brittle, relying on deprecated `__init__` arguments and scattered logic.
4. **The God Object:** `command.py` had grown into a massive anti-pattern, violating the Single Responsibility Principle (SRP) by housing the core `Command` representation alongside dozens of domain-specific payload constructors.

### Consequences:

If left unaddressed, the FSM early write binding risks silent packet drops and internal buffer errors during edge-case transport disconnects. Silent `None` returns mask critical transmission failures, making dropped commands incredibly difficult to diagnose. Finally, the "God Object" architecture of `command.py` makes testing fragile and code maintenance increasingly risky for future refactoring phases.

### The Fix:

We systematically decoupled the structural architecture from the protocol logic to stabilize the transport layer:
- Split the monolithic `command.py` file into a modular package utilizing domain-specific Mixins.
- Relocated and modernized the `Qos` class.
- Secured the transport binding initialization state to prevent premature writes.
- Enforced strict, typed return contracts across the FSM and `_BaseProtocol`.
- Reorganized test files to mirror the source directory structure (industry best practice).

### Technical Implementation:
- **SRP Split:** Converted `command.py` into a `command/` directory package. Extracted domain constructors into `dhw.py`, `hvac.py`, `system.py`, and `zones.py` as Mixin classes. The primary `Command` object inherits these mixins, completely preserving the backward-compatible API signature.
- **Strict Typing:** Resolved all `mypy --strict` errors by implementing generic `TypeVar` bounding (`_T = TypeVar("_T", bound=CommandBase)`) across the new Mixin architecture.
- **QoS Modernization:** Moved `Qos` to `protocol/qos.py` alongside `QosManager`. Refactored `Qos` into a strictly typed `@dataclass(frozen=True)`, stripping deprecated initialization arguments and enforcing parameters via `__dataclass_fields__` reflection. Added Sphinx-compliant docstrings.
- **Transport Binding:** Initialized `self._pause_writing = True` in `base.py` to start the protocol in a safe, read-only mode until `connection_made()` is confirmed.
- **FSM Contracts:** Refactored `_send_cmd` to guarantee a `Packet` return type. Explicitly trapped `PacketPayloadInvalid` in `fsm.py` (`WantEcho` and `WantRply` states) to gracefully discard malformed packets. Removed the `if not pkt: # HACK` check in `base.py`.
- **Test File Restructure:** Moved test files to mirror the program directory structure:
  - `tests/tests_tx/test_protocol_base.py` -> `tests/tests_tx/protocol/test_base.py`
  - `tests/tests_tx/test_protocol_fsm.py` -> `tests/tests_tx/protocol/test_fsm.py`

### Testing Performed:
- Validated that the `ramses_rf` `pytest` suite passes 100% locally.
- Fixed timestamp mocking discrepancies in `system.py` logic tests.
- Fixed `__mro__` attribute traversal in `test_apis_common.py` to correctly test inherited Mixin properties.
- **Downstream Integration Testing:** Ran the `pytest` suite for `ramses_cc` (current master) utilizing this updated `ramses_rf` branch. **All tests passed in `ramses_cc`**, confirming zero breaking changes to the upstream integration API.

Technical debt will continue to compound. The FSM cannot be safely decoupled for the planned unified `CentralStore` state management system while it still relies on loose typing contracts and silent failure swallowing.

### Risks of Implementing:

Because this touches the core foundational `Command` structure and FSM transition logic, the surface area for regressions is massive. If the Mixin API signature was slightly altered, or if an FSM timeout transitioned incorrectly, it could cause cascading failures in upstream components (like Home Assistant integrations).

### Mitigation Steps:
- Enforced 100% `mypy --strict` compliance across all modified files.
- Isolated the architectural refactor (SRP split) from the logical behavior changes to ensure pinpoint isolation of bugs.
- Verified backward compatibility by explicitly running the downstream `ramses_cc` test suite, proving the upstream payload consumers are unaffected by the under-the-hood structural changes.

### AI Assistance Disclosure:


src/ramses_tx/protocol/fsm.py
```diff
@@ -566,11 +580,11 @@ def pkt_rcvd(self, pkt: Packet) -> None:
 
         if (
             self._sent_cmd.rx_header[:8] == "0418|RP|"  # type: ignore[index]
-            and self._sent_cmd.rx_header[:-2] == pkt._hdr[:-2]  # type: ignore[index]
+            and self._sent_cmd.rx_header[:-2] == pkt_hdr[:-2]  # type: ignore[index]
```
silverailscolo, line 583: Still a bit of slicing here, fine

silverailscolo (approved):
@PWhite-Eng Thanks. ~Can you pull in master?~
Runs fine on hvac

src/ramses_tx/protocol/fsm.py
PWhite-Eng: Once Priority 3 is inprogress/implimented, it will be one of my tasks to ensure we are not sting slicing in the code without a very good reason.

PR #612, opened 2026-04-18
Refactor: Priority 2 Configuration Consolidation and SRP Extraction

PWhite-Eng:
### The Problem:
1. `src/ramses_tx/gateway.py` violated the Single Responsibility Principle by housing both the core `Engine` IO manager and the `ApplicationMessage` data structure.
2. Configuration was handled via a flat, untyped list of `**kwargs` passed into `Gateway.__init__`. This leaked Transport-layer concepts into the RF layer, broke encapsulation (e.g., `_config()` reaching into protected engine variables), and bypassed strict type-checking.
3. The engine boot sequence was out of order, starting the live RF engine before the historic cache was fully restored.

### Consequences:

If left unaddressed, this architectural debt would severely hinder future roadmap phases. The lack of strict configuration boundaries makes the codebase brittle, difficult to maintain, and prone to downstream `TypeError` crashes.

### The Fix:

Extracted the `Engine` and `ApplicationMessage` into dedicated files. Created a strict `EngineConfig` Data Transfer Object (DTO) in the Transport layer, which is now composed inside the `GatewayConfig` DTO in the RF layer. Reordered the gateway boot sequence and implemented a backward-compatibility shim for downstream consumers.

### Technical Implementation:
- **SRP Split:** Created `src/ramses_tx/engine.py` and `src/ramses_tx/application_message.py`, safely routing type hints with `TYPE_CHECKING` to prevent circular imports.
- **DTO Composition:** Created `src/ramses_tx/config.py`. Modified `ramses_rf.GatewayConfig` to accept `engine=EngineConfig(...)`.
- **Deprecation Adapter:** Updated `Gateway.__init__` to catch legacy `**kwargs` (used by Home Assistant / `ramses_cc`), map them natively into the new `EngineConfig` hierarchy, and emit a `DeprecationWarning` instead of crashing.
- **Boot Sequence:** Swapped `await self._engine.start()` to run *after* `await self._restore_cached_packets()` in `Gateway.start()`.
- **Test Alignment:** Refactored `helpers.py` and `conftest.py` to use `dataclasses.fields` to dynamically intercept flat test configuration variables and nest them into the correct DTOs.

### Testing Performed:
- Full local `pytest` suite execution (100% pass rate).
- Full `mypy --strict` validation (Zero errors across all modified files).
- Added `test_gateway_legacy_kwargs_deprecation` to explicitly test that legacy `**kwargs` from `ramses_cc` are successfully intercepted, mapped to the correct DTOs, and trigger the appropriate warning without failing.

Leaving the code as-is perpetuates the mixing of Transport and Application layers, making strict type safety impossible and acting as a structural blocker for subsequent refactoring phases (such as the Dispatcher/Parser overhauls).

### Risks of Implementing:

Downstream integrations (`ramses_cc`) utilizing deeply obscure or entirely unsupported kwargs during initialization might experience unexpected drops of those variables if they do not map to the known legacy keys.

### Mitigation Steps:

Implemented a highly robust Graceful Deprecation Pattern in `Gateway.__init__`. Known legacy kwargs (`disable_sending`, `hgi_id`, `packet_log`, etc.) are actively caught and re-routed to the new config objects. A clear log warning is emitted to instruct downstream maintainers on how to upgrade their initialization calls in future releases.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_tx/engine.py
+++ b/src/ramses_tx/engine.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-"""RAMSES RF - The serial to RF gateway (HGI80, not RFG100)."""
+"""RAMSES RF - The serial to RF engine."""
 
 from __future__ import annotations
 
@@ -8,10 +8,11 @@ import asyncio
 import logging
 import threading
 from collections.abc import Awaitable, Callable
-from datetime import datetime as dt, timedelta as td
+from datetime import datetime as dt
 from typing import TYPE_CHECKING, Any, Never
 
 from .address import ALL_DEV_ADDR, HGI_DEV_ADDR, NON_DEV_ADDR
+from .application_message import ApplicationMessage
 from .command import Command
 from .const import (
     DEFAULT_DISABLE_QOS,
@@ -21,6 +22,7 @@ from .const import (
     DEFAULT_SEND_TIMEOUT,
     DEFAULT_WAIT_FOR_REPLY,
     SZ_ACTIVE_HGI,
+    Code,
     Priority,
 )
 from .message import Message
@@ -35,14 +37,6 @@ from .schemas import (
 from .transport import TransportConfig, transport_factory
 from .typing import PktLogConfigT, PortConfigT, QosParams
 
-from .const import (  # noqa: F401, isort: skip, pylint: disable=unused-import
-    Code,
-    I_,
-    RP,
-    RQ,
-    W_,
-)
-
 if TYPE_CHECKING:
     from .const import VerbT
     from .protocol import RamsesProtocolT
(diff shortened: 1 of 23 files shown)


PWhite-Eng:
### Cross-Repository Testing Update (`ramses_cc`)

I’ve just completed a full run of the `ramses_cc` (`master` branch) test suite against this PR. We do have a handful of downstream test failures, but after investigating the tracebacks, I can confirm that **the root cause lies within the `ramses_cc` test fixtures, and no further changes are required in `ramses_rf`.**

**Why didn't the backward-compatibility shim catch these failures?** The Graceful Deprecation adapter we built into `Gateway.__init__` is actually working perfectly. It successfully catches the legacy flat `**kwargs` from `ramses_cc` and maps them into the new strict `EngineConfig` DTO.

The tests are failing because the new `ramses_rf` architecture enforces correct behavior much more strictly than the old one did. Specifically, the adapter correctly enabled the strict `known_list` filter, which immediately exposed two legacy flaws in the `ramses_cc` test environment:
1. **The Gateway is missing from the test `known_list`:** The `TEST_CONFIG` in `ramses_cc` provides a whitelist of devices but fails to include the Gateway (HGI) itself (`18:006402`). Because this PR tightens up the Application/Transport layer boundaries, the engine now strictly enforces this list and actively drops the virtual gateway's own injected packets.
2. **A Boot Sequence Race Condition:** This PR intentionally swaps the `Gateway.start()` sequence so the live RF engine boots *after* cached packets are fully restored. `ramses_cc`'s `VirtualRf` tests currently blast virtual serial packets immediately upon setup. Because they don't yield to the event loop to wait for the new, slightly later engine boot, the injected packets vanish into the void.

**Next Steps & Resolution:** The backward-compatibility shim is solid and does exactly what it is supposed to do for downstream consumers. It is the `ramses_cc` test suite that was relying on looser, legacy behavior.

I will be submitting a companion PR to `ramses_cc` shortly to:
- Add the HGI to the test `known_list`.
- Introduce the correct `async_block_till_done()` awaits to fix the testing race condition.
- Ideally, update `ramses_cc` to natively pass the new `GatewayConfig` and `EngineConfig` DTOs, bypassing the legacy kwargs entirely.

Because of the architectural split and DTO enforcement, **this will require a coordinated version bump for both `ramses_rf` and `ramses_cc`.**

PWhite-Eng:
**Update: Fortifying the Backward-Compatibility Shim**

While implementing the companion changes in `ramses_cc`, I identified a final edge case in the `Gateway` initialization adapter. Home Assistant often passes configuration parameters in nested dictionaries (e.g., `{"ramses_rf": {"enforce_known_list": True}}`), whereas the initial shim was only looking for flat `**kwargs`.

I have updated this PR with two additional improvements:
1. **Recursive Unpacking in `Gateway`**: The `__init__` adapter now utilizes a recursive `_apply_kwargs` function. It will now automatically traverse nested dictionaries to find and map keys to the correct `GatewayConfig` or `EngineConfig` DTO attributes. This makes the library significantly more resilient for all legacy Home Assistant installations.
2. **Expanded Test Coverage**: I have added `test_gateway_nested_kwargs_unpacking` to the suite to verify this recursive logic. I also took the opportunity to fix a strict Mypy `comparison-overlap` error by correctly typing the packet log configuration comparison using `PktLogConfigT`.

With these changes, the `ramses_rf` library now passes 100% of its own tests and `mypy --strict` validation, and it has been verified to support the existing `ramses_cc` master branch configuration patterns seamlessly.

src/ramses_tx/engine.py
```diff
@@ -1,17 +1,18 @@
 #!/usr/bin/env python3
 
-"""RAMSES RF - The serial to RF gateway (HGI80, not RFG100)."""
```
silverailscolo, line 3: A good thing to rename, was confusing

silverailscolo (approved):

src/ramses_tx/engine.py
PWhite-Eng: I am glad I was not the only one to be confused by the naming, but never had a really valid reason to change it. Then my friend suggested it with a good reason, so I implimented it.

PR #618, opened 2026-04-21
Refine `DevType.RF=30`

silverailscolo:
This PR fixes issue #

- [x] AI Coding Assistance disclosure: not used

**PR Summary**
To address [ramses_extras expected mapping](https://github.com/wimpie70/ramses_extras/pull/78) of `30:` address prefix to either the RFG slug (heat RFG100 RF gateway) or the RFS slug (Spider HVAC gateway), removes fixed coupling of 30 to (only) RFG role.
Also cleans up identical packets from some test logs and 2 duplicates in `ramses.__all__`


wimpie70:
this might help:

The real problem
ramses_rf's device classification is purely prefix-based (Address.type = device_id[:2]). There's only one slug per hex prefix in DEV_TYPE_MAP, so 30: must resolve to exactly one of RFG or RFS — it can't be both.

Solution
The fix requires more than a table edit. Options:

Option A — Schema/known_list-driven (smallest change): Keep "30" → RFG as the default, but honor SZ_CLASS: RFS in known_list/schema so users with a spIDer can declare it:

```yaml
known_list:
  "30:123456": { class: RFS }
```
Then device factory must consult the schema class before falling back to prefix. This is the path you hinted at ("Any Gateway must have its class defined explicitly" already exists for HGI).

Option B — Eavesdropping-based promotion: Make 30: resolve to a promotable slug (like HEA/HVC), and let class_dev_hvac() / the eavesdrop logic promote it to RFG or RFS based on observed codes (HVAC-only codes → RFS, Heat codes → RFG). This is how FAN/CO2/HUM/REM already work.

Option C — Hybrid (recommended):

Default 30: → RFG (preserves backward compatibility for the overwhelmingly common RFG100 case)
SZ_CLASS: RFS in schema overrides to RFS (Option A)
Eavesdrop promotion on HVAC-only codes as a fallback (Option B)
The revert shows Option B alone is non-trivial: many callsites assume DEV_TYPE_MAP.RFG == "30" holds. Option A is the minimum-risk fix that unblocks spIDer users without breaking existing RFG100 deployments.

Suggestion: drop the AttrDict change and instead thread SZ_CLASS through the device factory so explicitly-declared RFS devices are honored at the 30: prefix.

Issue #617, opened 2026-04-20
Architectural Proposal: `ramses_rf` & `ramses_tx`


silverailscolo:
Submitted by @PWhite-Eng 

## 1. Project Context & Objective

`ramses_rf` (and its transport layer `ramses_tx`) is a Python library designed to decode and interact with an undocumented 868MHz RF protocol used by Honeywell/Resideo heating and various HVAC systems.

**The Objective:** Transition the codebase from its organic, reverse-engineered origins into an enterprise-grade, industry-standard networking library. This requires eliminating technical debt, enforcing strict Single Responsibility Principle (SRP) boundaries, unifying fragmented state management, and protecting downstream consumers (like Home Assistant) from event loop saturation.

## 2. Core Architectural Principles

All future development and refactoring must adhere to the following non-negotiable standards:
1. **OSI-Model Layering:** Strict separation between the Data Link layer (lexical string slicing) and the Application layer (semantic domain logic).
2. **Immutable Data Transfer Objects (DTOs):** Network packets must never be mutated in flight. Arrays and fragments are handled via composition, not payload mutation.
3. **Single Source of Truth:** State must exist in exactly one place. The transport layer (`ramses_tx`) must never cache application state.
4. **Strict Type Safety:** Every phase must pass `mypy --strict` with zero errors. No `@property` shims or `type: ignore` bandaids are allowed for structural domain models.
5. **Test-Driven Reliability:** Every phase must pass the full `pytest` suite. Any tests broken by structural code changes must be updated to reflect the new architecture before the phase is considered complete.

---

### Priority 1 & 2: Transport Stabilization & Configuration - (Completed)
- Secured the FSM (Finite State Machine) against premature writes.
- Enforced strict packet return contracts.
- Extracted configuration into strict `GatewayConfig` and `EngineConfig` DTOs, decoupling the application layer from transport internals.

### Priority 3: The Lexical & Semantic DTO Refactor (Current Phase)

This phase eradicates circular dependencies, separates lexical string parsing from semantic meaning, and replaces dangerous payload mutation with a standard network defragmentation buffer.
- **Sub-PR 3A: Device Identity Foundations**
  - Introduce a strictly typed, frozen `DeviceId` dataclass storing `type_id` and `device_id` strictly as integers.
  - Completely eradicate the legacy `Address` string-based class.
  - Update `parsers.py`, `dispatcher.py`, `device/base.py`, and `command/base.py` to natively utilize `DeviceId` attributes.
[...]

PR #621, opened 2026-04-22
fix promotion when eavesdropping

wimpie70:
I found this issue this when testing the sim...
Analyzed and fixed with Claude Opus 4.7 medium

# Device class promotion via `self.__class__ = cls` skips subclass `__init__`, causing `AttributeError` at runtime

## Summary

When an HVAC device is promoted in-place through eavesdropping
(`DeviceBase._handle_msg` at [`device/base.py:207`](../src/ramses_rf/device/base.py#L207)),
the promotion is performed by swapping `self.__class__ = cls`. This does **not**
run the new subclass's `__init__`, so any instance attributes that the
subclass sets up there are missing on the promoted device. This leads to a
`AttributeError` the first time downstream code (including `ramses_cc`) calls
a method that relies on that state.

The same pattern affects several heat-domain subclasses (`Controller`,
`UfhController`, `DhwSensor`), although heat setups rarely hit it in practice
because heat classes are usually selected up-front based on the address type
or a `known_list` entry.

## Reproduction

Running under 0.56.6 with `enable_eavesdrop: true`, after a FAN device emits
`I|31DA`/`RP|31DA` and is promoted from `HVC` → `FAN`, `ramses_cc` adds the
climate entity and calls `self._device.get_bound_rem()`. This crashes:

> **Note:** the trigger is the promotion mechanism itself, not the specific
> traffic that caused it. *Any* eavesdrop-driven promotion — correct or
> incorrect — hits this bug, because `self.__class__ = cls` intrinsically
> bypasses the new subclass's `__init__`. Replacing the triggering traffic
> with "correct" frames only changes *which* class the device is promoted to,
> not whether the crash occurs.

```
File "/config/custom_components/ramses_cc/climate.py", line 961, in async_added_to_hass
    self._bound_rem = self._device.get_bound_rem()
File ".../ramses_rf/device/hvac.py", line 842, in get_bound_rem
    if not self._bound_devices:
           ^^^^^^^^^^^^^^^^^^^
AttributeError: 'HvacVentilator' object has no attribute '_bound_devices'.
Did you mean: 'add_bound_device'?
```

`HvacVentilator.__init__` initializes `_bound_devices`, `_supports_2411`,
`_params_2411`, `_initialized_callback`, `_param_update_callback`, and `_hgi`
(see [`device/hvac.py`](../src/ramses_rf/device/hvac.py)), but none of those
attributes are set after an in-place class swap.

## Root cause

```python
# src/ramses_rf/device/base.py
207:             self.__class__ = cls
```

`self.__class__ = cls` is a rebind, not a construction, so Python never runs
`cls.__init__`. Any subclass state defined only in `__init__` is therefore
absent on a promoted instance.

Affected subclasses with `__init__`-only state:

- `HvacVentilator` — `_bound_devices`, `_supports_2411`, `_params_2411`,
  `_initialized_callback`, `_param_update_callback`, `_hgi`
- `Controller` — `tcs` (and `_make_tcs_controller` call)
- `UfhController` — `circuit_by_id`, `_setpoints`, `_heat_demand`,
  `_heat_demands`, `_relay_demand`, `_relay_demand_fa`
- `DhwSensor` — `_child_id`

## Relationship to existing work

I searched ramses_rf / ramses_cc open + closed issues and PRs for terms
`bound_devices`, `HvacVentilator`, `eavesdrop`, `promote`, and reviewed the
architectural roadmap. Nothing explicitly covers this device-level promotion
bug:

- **PR ramses-rf/ramses_rf#584** (merged, 0.56.5) —
  *"Refactor: Remove `__class__` mutation anti-pattern and strictly decouple
  transport/application messages"*. Addresses `msg.__class__ = Message` in the
  **message/transport layer**. Conceptually the same anti-pattern, but a
  different location; the device-layer promotion at `device/base.py:207` is
  untouched.
- **Issue #617** — *Architectural Proposal: `ramses_rf` & `ramses_tx`*. The
  roadmap touches `device/base.py` only under **Sub-PR 3A** (a `DeviceId`
  dataclass refactor). Device-class promotion is not called out.
- **Issue #580** — architectural review report. Does not mention this crash
  or the device promotion path.
- No other open issue/PR matches `bound_devices`, `HvacVentilator`, or the
  class-swap pattern at `device/base.py:207`.

## Proposed fix

Add a small, opt-in hook invoked right after the class swap. Subclasses
implement the hook and use `__dict__.setdefault(...)` for idempotency.

```python
# device/base.py
self.__class__ = cls
post_init = getattr(self, "_post_class_promote", None)
if callable(post_init):
    post_init()
```

```python
# device/hvac.py (HvacVentilator)
def __init__(self, *args, traits=None, **kwargs):
    super().__init__(*args, traits=traits, **kwargs)
    self._init_fan_state()

def _init_fan_state(self):
    self.__dict__.setdefault("_supports_2411", False)
    self.__dict__.setdefault("_params_2411", {})
    self.__dict__.setdefault("_initialized_callback", None)
    self.__dict__.setdefault("_param_update_callback", None)
    self.__dict__.setdefault("_hgi", None)
    self.__dict__.setdefault("_bound_devices", {})

def _post_class_promote(self):
    self._init_fan_state()
```

Analogous hooks are added to `Controller`, `UfhController`, and `DhwSensor`
(see the accompanying PR). The change is minimal, idempotent, and fully
backward-compatible: when a device is constructed through the normal factory
path, `__init__` runs as before and the hook finds everything already set.

## Tests

A new regression test suite `tests/tests/test_class_promotion.py` creates
bare instances of each affected subclass via `cls.__new__(cls)`
(mirroring the state an in-place class swap leaves behind), invokes
`_post_class_promote`, and asserts:

1. The previously-crashing public APIs (`HvacVentilator.get_bound_rem`,
   `add_bound_device`, `remove_bound_device`) now work.
2. All subclass instance attributes exist with their expected defaults.
3. The hook is idempotent — calling it a second time does **not** wipe
   existing state or invoke `_make_tcs_controller` twice.

All 5 tests pass locally.

## Version info

- `ramses_rf` master (0.56.7 milestone)
- Reproduced with `ramses_cc` 0.56.6
- Python 3.13 / 3.14


wimpie70:
The PR touched only three files, but the new logic tripped global mypy rules that run across the entire repo. Two main cascades happened:

FAN attributes added only via __dict__.setdefault: because _supports_2411, _params_2411, _hgi, _bound_devices, etc. were never declared/assigned in the class body, mypy flagged every read of them throughout hvac.py and all the unit tests that touch FAN objects. Fixing that required 
(a) adding class-level annotations and 
(b) revisiting every helper/test that references those attributes, even in files we didn’t edit originally.

New tests added without typing: tests/tests/test_class_promotion.py was newly created (or heavily changed) in the PR. Repo-level mypy config enforces disallow_untyped_defs on tests.*, so the missing annotations in that file immediately caused failures. 
To satisfy mypy we had to add typing to the pytest fixtures and helper functions; that’s still the same file, but until we add the annotations mypy errors won’t go away.

While fixing those mypy hits, the global type run also started complaining about serial imports lacking stubs and about 2411 attribute usage from other files; that’s because our changes made the typing run reach code paths that were silently accepted before (e.g., calling self._params_2411 in tests now needs the attribute actually defined). 
So even though git says we only edited three files, the repo-wide type checker forces us to address every spot that breaks once the new code is present.

silverailscolo (approved):

silverailscolo:
Thanks @wimpie70 Later on we might refactor the way devices get promoted; for now, your PR at least will make the checks work.

PR #622, opened 2026-04-22
refactor(core): Implement context injection bridge to resolve encapsulation leak (Issue #608, Step 1)

PWhite-Eng:
### The Problem:
There is a severe encapsulation leak between the transport layer (`ramses_tx`) and the application layer (`ramses_rf`). In `ramses_tx/message.py`, the parser illegally queries `getattr(self.src, "_is_controller")`. Because `self.src` is a transport `Address` object with no domain knowledge, this lookup fails or defaults incorrectly if the device isn't fully initialized. (Tracking: Issue #608).

### Consequences:
Because `ramses_tx` cannot correctly determine if a device is a controller, it fails to map the `zone_idx`. The index is silently dropped, causing valid `evohome` (HEAT) zones to disappear or be marked as missing/dead in `ramses_rf` and Home Assistant.

### The Fix:
Implemented the "Strangler Fig" pattern by injecting a domain-aware callback bridge from `ramses_rf` into `ramses_tx`. This severs the hard dependency, allowing the transport layer to safely ask the application layer for device traits without knowing about the domain objects themselves.

### Technical Implementation:
- Added a module-level `_IS_CONTROLLER_CB` callback to `ramses_tx.message`.
- Updated `Message._idx` to use this callback instead of `getattr()`, with a fallback for legacy isolated tests.
- Injected `DeviceRegistry` domain logic via `Gateway.__init__` in `ramses_rf/gateway.py` to populate the callback at runtime.
- Added a strict type cast `cast(DeviceIdT, device_id)` to satisfy Mypy dictionary key constraints.
- Retained temporary bug-compatibility for Underfloor Heating Controllers (`02:`) to ensure legacy snapshot tests remain green.

### Testing Performed:
- Ran full `pytest` suite (970+ tests passing, including snapshot tests).
- Ran `mypy --strict` against the `ramses_rf` directory (Success: no issues found in 146 source files).

Production systems will continue to experience dropped heating zones. Furthermore, this encapsulation leak blocks any future attempts to clean up the O(N²) state thrashing currently degrading CPU performance.

### Risks of Implementing:
There is a minor risk that packets from completely unknown devices (not yet in the registry) might evaluate incorrectly. 

### Mitigation Steps:
The bridge callback is designed to return `True` (default legacy behavior) if a device is not found in the registry, ensuring we err on the side of caution and do not drop valid payload indices prematurely. We also explicitly mimic legacy bug-compatibility for `02:` devices to prevent regressions in heat demand parsing.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -16,6 +16,7 @@ from datetime import datetime as dt, timedelta as td
 from logging.handlers import QueueListener
 from typing import TYPE_CHECKING, Any, Literal, cast
 
+import ramses_tx.message as tx_msg
 from ramses_tx import (
     Command,
     Engine,
@@ -40,6 +41,7 @@ from ramses_tx.logger import flush_packet_log
 from ramses_tx.schemas import SZ_BLOCK_LIST, SZ_ENFORCE_KNOWN_LIST, SZ_KNOWN_LIST
 
 from .const import DONT_CREATE_MESSAGES, HIGH_VOLUME_STATUS_CODES
+from .typing import DeviceIdT
 
 from .const import (  # noqa: F401, isort: skip, pylint: disable=unused-import
     I_,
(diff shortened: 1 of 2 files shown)


PWhite-Eng:

silverailscolo (approved):

PR #624, opened 2026-04-22
refactor(tx): Introduce memory-efficient Transport DTOs (Issue #608, Step 2)

PWhite-Eng:
### The Problem:
The legacy `Message` and `Packet` objects within `ramses_tx` tightly couple transport layer routing with application domain logic. They also rely heavily on string representation and dynamic attribute resolution, which significantly increases the memory footprint and contributes to the CPU thrashing identified in Issue #608.

### Consequences:
Without clearly defined, immutable Data Transfer Objects (DTOs), it is impossible to cleanly sever the semantic payload parsing out of the transport layer. The system is forced to continually re-evaluate strings rather than passing lightweight integers and Enums.

### The Fix:
Introduced strictly-typed, OSI-aligned domain models (`TransportMessage`, `RawPacket`, `DeviceId`) to act as the new boundary layer between `ramses_tx` and `ramses_rf`. Implemented a passive `.dto` property on the legacy `Message` class to validate the data mapping in the background.

### Technical Implementation:
- Created `src/ramses_tx/models.py` housing `frozen=True` dataclasses.
- Replaced legacy string-based device IDs with memory-efficient `DeviceId` objects (storing type and ID as integers).
- Reused the existing `VerbT` enum from `const.py` to ensure strict type compliance.
- Added a `.dto` property to the legacy `Message` class that instantiates `TransportMessage` on-the-fly. This guarantees the new data structures perfectly represent the legacy data state across all 970+ test cases without altering existing runtime behavior.

### Testing Performed:
- Ran full `pytest` suite (931 passed, including snapshot tests). This proves the DTO mapping successfully handles all edge cases in the test environment.
- Ran `mypy --strict` against the entire `ramses_rf` directory (Success: 147 files).

Failing to introduce these structural DTOs prevents the migration of `parsers.py` out of the transport layer, effectively halting the architectural roadmap to fix the O(N²) CPU thrashing.

### Risks of Implementing:
Because this phase utilizes a passive generation strategy (the new DTOs are instantiated but not yet utilized by the broader application), the regression risk to production users is effectively zero.

### Mitigation Steps:
Used `@property` generation for the DTOs rather than replacing the legacy class outright. This "shadow implementation" allows us to verify data integrity against the snapshot tests safely.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -16,6 +16,7 @@ from datetime import datetime as dt, timedelta as td
 from logging.handlers import QueueListener
 from typing import TYPE_CHECKING, Any, Literal, cast
 
+import ramses_tx.message as tx_msg
 from ramses_tx import (
     Command,
     Engine,
@@ -40,6 +41,7 @@ from ramses_tx.logger import flush_packet_log
 from ramses_tx.schemas import SZ_BLOCK_LIST, SZ_ENFORCE_KNOWN_LIST, SZ_KNOWN_LIST
 
 from .const import DONT_CREATE_MESSAGES, HIGH_VOLUME_STATUS_CODES
+from .typing import DeviceIdT
 
 from .const import (  # noqa: F401, isort: skip, pylint: disable=unused-import
     I_,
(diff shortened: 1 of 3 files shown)


PWhite-Eng:

PWhite-Eng:
@silverailscolo, this is a stacked PR ontop of PR #622

silverailscolo (approved):

silverailscolo:
Thanks @PWhite-Eng Another important step towards better code

PR #625, opened 2026-04-22
refactor(core): Lift semantic payload parsing out of transport layer (Issue #608, Step 3)

PWhite-Eng:
### The Problem:
`ramses_tx` was operating as an application layer by physically importing `parsers.py` and `CODES_SCHEMA`. This tight coupling forced the transport engine to perform heavy regex validations and domain routing (`zone_idx`), preventing us from upgrading the transport buffer to handle fragmentation cleanly.

### Consequences:
As long as `ramses_tx` tried to parse payloads natively, it was impossible to implement a clean OSI Layer 4 defragmenter. Furthermore, testing application logic required booting up parts of the transport protocol.

### The Fix:
Physically moved `parsers.py` into the `ramses_rf` domain where it belongs. Replaced all static references to schemas and decoders inside `ramses_tx/message.py` with injected callbacks fed down from the Gateway.

### Technical Implementation:
- `git mv src/ramses_tx/parsers.py src/ramses_rf/parsers.py`.
- Updated all relative imports inside the migrated parser.
- Moved `_check_msg_payload` and regex validation out of `message.py` and into the new parser file.
- Established global module-level bridges in `ramses_rf/gateway.py` so `ramses_tx` can safely log schema names and route `_idx` properties dynamically.
- Refactored isolated `ramses_tx` unit tests to use the new bridge architecture instead of `unittest.mock.patch` for the parser pipeline.

### Testing Performed:
- Full `pytest` suite ran successfully (931 passed).
- `mypy --strict` ran successfully against all 147 files.
- Verified that isolated empty-payload tests correctly fall back without failing.

Failing to lift the parser keeps the transport layer inextricably tied to the application layer, permanently blocking the #608 roadmap to fix O(N²) CPU thrashing.

### Risks of Implementing:
Because we are using global callback bridges, there is a minor risk that isolated test execution order could bleed state if not mocked properly.

### Mitigation Steps:
Explicitly mocked out `_PAYLOAD_DECODER_CB` to `None` in targeted `ramses_tx` tests to guarantee state isolation.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -16,6 +16,7 @@ from datetime import datetime as dt, timedelta as td
 from logging.handlers import QueueListener
 from typing import TYPE_CHECKING, Any, Literal, cast
 
+import ramses_tx.message as tx_msg
 from ramses_tx import (
     Command,
     Engine,
@@ -40,6 +41,7 @@ from ramses_tx.logger import flush_packet_log
 from ramses_tx.schemas import SZ_BLOCK_LIST, SZ_ENFORCE_KNOWN_LIST, SZ_KNOWN_LIST
 
 from .const import DONT_CREATE_MESSAGES, HIGH_VOLUME_STATUS_CODES
+from .typing import DeviceIdT
 
 from .const import (  # noqa: F401, isort: skip, pylint: disable=unused-import
     I_,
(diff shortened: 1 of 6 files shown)


PWhite-Eng:

PWhite-Eng:
@silverailscolo, this is a stacked PR ontop of PR #624

silverailscolo (approved):

silverailscolo (approved):

PR #628, opened 2026-04-23
Fix: Resolve circular import crash during initialization by relocating `LOOKUP_PUZZ`

PWhite-Eng:
### The Problem:

A circular dependency exists between the foundational transport layer (`ramses_tx`) and the higher-level parsing layer (`ramses_rf`). During initialization, `ramses_tx/command/system.py` attempts to import `LOOKUP_PUZZ` from `ramses_rf/parsers.py`, which in turn triggers `ramses_rf` to attempt an import of `Command` back from a partially initialized `ramses_tx`.

### Consequences:

Home Assistant completely fails to load the `ramses_cc` custom component. The event loop throws an `ImportError` on startup, resulting in a fatal failure of the integration for users running the current master branch.

### The Fix:

Relocated the `LOOKUP_PUZZ` constant to the lower-level foundational library (`ramses_tx`) to break the reverse dependency and enforce a clean, unidirectional initialization sequence.

### Technical Implementation:
- Moved the `LOOKUP_PUZZ` dictionary definition to `src/ramses_tx/const.py`.
- Updated `src/ramses_tx/command/system.py` to import `LOOKUP_PUZZ` locally from the `ramses_tx` namespace.
- Updated `src/ramses_rf/parsers.py` to import `LOOKUP_PUZZ` from its new home in `ramses_tx.const`.
- This severs the link where the lower layer depends on the upper layer, allowing `ramses_tx` to finish loading before `ramses_rf` is called.

### Testing Performed:
- Verified module loading sequence manually to ensure `ramses_tx` initializes fully without triggering `ramses_rf`.
- Simulated integration startup to confirm the `ImportError` regarding `Command` from `ramses_tx` no longer occurs in the Home Assistant logs.

The `master` branch will remain fundamentally broken, and the `ramses_cc` integration will continue to crash on startup for any users attempting to run it.

### Risks of Implementing:

Any external, un-updated scripts or third-party integrations that explicitly rely on importing `LOOKUP_PUZZ` directly from `ramses_rf.parsers` will encounter an `ImportError`.

### Mitigation Steps:

Audited internal project files to ensure all known internal references now point to the correct location in `src/ramses_tx/const.py`. The architectural benefit of decoupling the libraries heavily outweighs the risk of breaking undocumented, direct access to parsing constants.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/parsers.py
+++ b/src/ramses_rf/parsers.py
@@ -31,6 +31,7 @@ from ramses_tx.const import (
     FAULT_DEVICE_CLASS,
     FAULT_STATE,
     FAULT_TYPE,
+    LOOKUP_PUZZ,
     SYS_MODE_MAP,
     SZ_ACCEPT,
     SZ_ACTIVE,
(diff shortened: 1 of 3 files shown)


silverailscolo (approved):

silverailscolo:
Thanks, just ran into this when I pulled the latest and greatest into a working branch.

PR #638, opened 2026-04-28
perf: Implement O(1) Hybrid Push Model for EntityState to resolve CPU exhaustion

PWhite-Eng:
### The Problem:

`ramses_rf` suffers from severe $O(N^2)$ state thrashing, leading to continuous CPU ramp-up over time (Issues #583, #608). The legacy `EntityState` query model iterates over the entire global message history and redundantly evaluates payloads up to 3 times per packet for every single state query. Additionally, a "Context Flattening Bug" in multi-zone devices breaks Eavesdropping topology generation.

### Consequences:

As the system uptime increases, the continuous array traversal and regex/dictionary evaluations saturate the Python/Home Assistant Event Loop. This results in delayed sensor updates, dropped packets, and eventually forces the Supervisor to forcefully kill the integration after 8-12 hours of uptime.

### The Fix:

Migrated `EntityState` from an unbounded Pull/Query architecture to a highly performant `O(1)` Hybrid Push Model, backed by a context-aware dictionary cache.

### Technical Implementation:
1. **Context-Aware Dictionary:** Created `self._current_state` keyed by `(Code, VerbT, Context)` to ensure multi-zone devices properly isolate their states.
2. **The Cursor (`_sync_state`):** Introduced a lightweight cursor that tracks the length of the global `message_store`. It only evaluates and caches *new* packets that have arrived since the last sync.
3. **Surgical Query Bypass:** Modified `_msg_value_code` and `get_message_log_flat` to pull directly from the `O(1)` dictionary, completely bypassing `_build_state_cache` during routine HA polling.

### Testing Performed:
- Created `tests/tests_rf/test_entity_state_performance.py` to mathematically prove the eradication of the CPU bug. The test simulates 5,000 packets and asserts that the `msg.payload` property is accessed exactly *once* during a query, down from 15,000+ accesses in the legacy code.
- Passed `mypy --strict` with zero errors.
- Passed the full `ramses_rf` pytest suite (938 tests).
- Passed the full `ramses_cc` pytest suite utilizing the linked updated library.

The integration will remain fundamentally unstable for users with moderate-to-large Evohome deployments, guaranteeing system crashes within 24 hours of boot. Eavesdropping will continue to fail for multi-zone controllers.

### Risks of Implementing:

Because we are altering the core state retrieval mechanism, there is a small risk that highly specific edge-case devices might have unconventional Context IDs (`ctx`) that the tuple dictionary fails to match perfectly, potentially resulting in `None` being returned for an obscure sensor.

### Mitigation Steps:
- Maintained fallback logic inside `_msg_value_code` so that if an exact context match fails, it falls back to looking for `False` or `None` contexts.
- Extensively tested against the existing Eavesdropping Schema test suite, which acts as a highly rigorous fuzz-tester for Context/Zone matching. All tests pass.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/entity_state.py
+++ b/src/ramses_rf/entity_state.py
@@ -90,6 +90,77 @@ class EntityState:
         """Initialize the EntityState."""
         self._entity = entity
         self._gwy = gwy
+        # NEW: Context-Aware O(1) Dictionary for Push-Model state caching
+        self._current_state: dict[tuple[Code, VerbT, Any], ApplicationMessage] = {}
+        self._log_cursor: int = 0  # Tracks cursor position in global log
+
+    def _sync_state(self) -> None:
+        """Hybrid Push Model: Sync O(1) cache using a cursor on the global log."""
+        if self._gwy.message_store is None:
+            return
+
+        log_iterable = self._gwy.message_store.log_by_dtm
+        if isinstance(log_iterable, dict):
+            log_iterable = list(log_iterable.values())
+
+        current_len = len(log_iterable)
+        if current_len == self._log_cursor:
+            return  # No new packets, O(1) instant exit
+
+        if current_len < self._log_cursor:
+            # The global log was cleared (e.g. cache reset). Rebuild from scratch.
+            self._log_cursor = 0
+            self._current_state.clear()
+
+        # Only process NEW packets appended since the last sync
+        new_msgs = log_iterable[self._log_cursor :]
+        for msg in new_msgs:
+            self.update_state(msg)
+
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
Tested with ramses_cc (latest master) and all of ramses_cc pytests passed

silverailscolo (approved):

silverailscolo:
@peternash can you try ramses_rf@master? It contains a new lookup approach by @PWhite-Eng - thanks! - that should fix the CPU issue.

PR #640, opened 2026-04-29
feat: Introduce OSI DTO boundaries for ramses_tx (#639 Phase 1)

PWhite-Eng:
### The Problem:

The `ramses_tx` and `ramses_rf` libraries currently suffer from Domain Logic leakage. The Transport layer (`ramses_tx`) handles Application-level concerns (like L7 payload decoding, Request/Response state tracking, and virtual device masking), which causes circular dependencies and violates the Single Responsibility Principle. 

### Consequences:

If this isn't fixed, the codebase remains tightly coupled, making it difficult to maintain, test, and safely extend. Changes in device or routing logic risk breaking the core L2/L3 RF modem transport layer, and the intended Hexagonal architecture cannot be realized.

### The Fix:

Introduced strict Data Transfer Objects (DTOs) to establish a firm, immutable OSI Layer boundary between the RF modem (`ramses_tx`) and the domain model (`ramses_rf`). 

### Technical Implementation:
- Created `src/ramses_tx/dtos.py` containing strict, immutable, primitive-based `PacketDTO` and `CommandDTO` classes. 

- Refactored `Packet.to_dto()` in `src/ramses_tx/packet.py` to populate `addr1`, `addr2`, `addr3` purely via positional L2 parsing, stripping out all semantic L7 resolution. 

- Enforced strict 2-character string padding for verbs (e.g., dynamically padding `"I"` to `" I"`) to ensure protocol accuracy at the DTO boundary.
- Implemented a backward-compatibility shim in `Packet.to_dict()` to map the strict DTO string primitives back to legacy formats (casting RSSI to `int`, handling `"---"` and `"..."` as `None`, and re-injecting the raw `frame` key). This ensures upstream `ramses_rf.Gateway` logic does not break during the transition.

### Testing Performed:
- Created `tests/tests_tx/test_dtos.py` to assert accurate positional logic and verb formatting.
- Updated `tests/tests_tx/test_packet.py` to validate that DTO serialization and state restoration operate flawlessly with the legacy dictionary schemas.
- Executed the full test suite (940 tests), achieving a 100% pass rate.
- Verified 100% compliance with `mypy --strict`.

The architecture remains entangled, preventing the safe extraction of payload decoding and FSM state management into `ramses_rf` (Phases 2-4 of the blueprint).

### Risks of Implementing:

Upstream `ramses_rf` components might unexpectedly fail if they rely on undocumented side-effects of the old `PacketDTO` or if the dictionary shim misses a null-equivalent edge case (like specific undocumented RSSI string formats from niche log files).

### Mitigation Steps:

Adopted the "Parallel Change" pattern. Functionality was mapped but not prematurely deleted. The old `Message` and `Command` classes are currently untouched. The `to_dict()` shim actively protects legacy consumers from the underlying DTO type changes, acting as a safe translation layer until `ramses_rf` is updated in subsequent phases. 

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_tx/dtos.py
+++ b/src/ramses_tx/dtos.py
@@ -1,0 +1,81 @@
+"""
+Data Transfer Objects for ramses_tx.
+
+This module defines the strict boundaries for OSI layer decoupling
+between the RF modem (L1-L3) and the Domain Model (L4-L7).
+"""
+
+from dataclasses import dataclass
+from datetime import datetime as dt
+
+
+@dataclass(frozen=True, slots=True)
+class PacketDTO:
+    """
+    Pure data object bridging the ramses_tx modem and ramses_rf.
+
+    :param timestamp: Time the frame was received.
+    :type timestamp: datetime
+    :param rssi: Received Signal Strength Indicator (e.g., "-72").
+    :type rssi: str
+    :param verb: The action verb (e.g., "RQ", "I", "W", "RP").
+    :type verb: str
+    :param seq: The sequence number (e.g., "003").
+    :type seq: str
+    :param addr1: Positional L2 Address 1 (e.g., "01:145038").
+    :type addr1: str
+    :param addr2: Positional L2 Address 2.
+    :type addr2: str
+    :param addr3: Positional L2 Address 3.
+    :type addr3: str
+    :param code: The packet code (e.g., "30C9").
+    :type code: str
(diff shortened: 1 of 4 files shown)


PWhite-Eng:
ran a check against ramses_cc (latest master). All ramses_cc pytest tests passed

silverailscolo (approved):

PR #641, opened 2026-05-01
feat: Migrate RX payload decoding to ramses_rf (#639 Phase 2)

PWhite-Eng:
### The Problem:

The `ramses_tx` transport layer handles application-level L7 payload decoding. This causes circular dependencies and violates the Single Responsibility Principle by forcing the L1-L3 modem adapter to understand domain-specific logic. 

### Consequences:

If this is not fixed, the transport layer remains tightly coupled to the application layer, preventing a clean Hexagonal (Ports and Adapters) boundary. 

### The Fix:

Safely moved the inbound payload decoding logic into `ramses_rf` and eliminated the L3 decoding callback bridge. 

### Technical Implementation:
- Copied the required payload parsing logic (`parsers/`) into a new decoder module in `ramses_rf`. 

- Configured the new decoder to accept a `PacketDTO` and output the semantic dictionary/object. 

- Updated `ramses_rf.Gateway._msg_handler` to accept `PacketDTO` and use the new decoder. 

- Removed the `PAYLOAD_DECODER_CB` from `ramses_tx`. 

### Testing Performed:
- Created `test_rx_payload_decoder_regression.py` and iterated over 30,000+ packets. 

- Asserted that the output of the old `ramses_tx.Message(raw).parsed_payload` is exactly equal to the output of `ramses_rf.new_decoder(dto)`. (the test file has now be updated to ensure decoder.py operates correctly under stress as the code has been refactored and it cannot compare the results to the original code)

- Ran `mypy --strict` which reports zero errors. 

- Ran `pytest` ensuring the full test suite passes. 

The architecture will continue to suffer from domain logic leakage, blocking the subsequent phases (TX parity and FSM decoupling) of the OSI decoupling blueprint. 

### Risks of Implementing:

Potential regressions in inbound packet decoding if the new L7 parser handles unexpected raw payloads differently than the legacy transport parser.

### Mitigation Steps:

Utilized the "Parallel Change" pattern. Extensive parity testing against 30,000+ packets guarantees the new decoder's semantic output is perfectly identical to the legacy implementation prior to cutting over. 

### AI Assistance Disclosure:


PWhite-Eng:
@silverailscolo, I have run this with the latest branch of ramses_cc, and ramses_cc pytest suite fails. I will take a look at it next. But I want to get this PR in, as it is one of the hardest things I have done in a while, and it is certainly one of the hardest pieces of programming I have undertaken, even with the help of my friend (who can be a little bit dim at times!), but the main reason is that I do not want another PR merged which will be hell to try and integrate in to this.
There is some significant work ahead, but I hope this is a big chunk of it.
Phil

silverailscolo (approved):
A clear PR, thanks

PR #643, opened 2026-05-04
`cutoff_dtm` not TZ-aware

silverailscolo:
Trying to fix TypeError in issue https://github.com/ramses-rf/ramses_cc/issues/651


silverailscolo:
This error depends on the version of the packet log being loaded into version 0.56.6 and later, presumably not using the messageDTO.

PR #650, opened 2026-05-06
fix/outbound frames not being logged in packet log

wimpie70:
I noticed we again don't publish outbound messages to the packet_log.
I know a lot of work is going on around this, so i'm not sure if you want this PR...
Just have a look at it...

Summary

Add _log_tx_packet() to _FullTransport and call it from write_frame() so every outbound frame is written to the packet log before hitting the transport implementation @ramses_tx/transport/base.py#234-256.
Keep the MQTT transport’s log_all_mqtt flag limited to console info logging so packet-log coverage applies uniformly across all transports @ramses_tx/transport/mqtt.py#328-452.
Leave simulator hooks untouched except for removing the earlier gateway-RQ mirroring to avoid duplicate entries once the core transport handles Tx logging (already done downstream).

### DIFF
--- a/src/ramses_tx/transport/base.py
+++ b/src/ramses_tx/transport/base.py
@@ -237,12 +237,27 @@ class _FullTransport(_ReadTransport):
         if self._closing is True:
             raise exc.TransportError("Transport is closing or has closed")
 
+        self._log_tx_packet(frame)
         self._track_transmit_rate()
         await self._write_frame(frame)
 
     async def _write_frame(self, frame: str) -> None:
         """Write some data bytes to the underlying transport."""
         raise NotImplementedError("_write_frame() not implemented here")
+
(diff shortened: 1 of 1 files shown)


silverailscolo:
@wimpie70 Looks good, thanks.
You know that ramses_cc is flagged by HomeAssistant as logging (way) too frequently. This is on startup, to the HA system log.
We could either temporarily raise the logging threshold during the start phase, or keep the Config switch now in place so that default users do not see this error.
BTW, ramses_extras is also flagged for logging too often on my 3-fan HVAC setup.

silverailscolo (approved):

wimpie70:
@silverailscolo Yeah, i know extras still logs a lot....currently trying to do some optimizing...and testing Peter's setup.
I'll be off for a few weeks but will try to get a decent version out today.

PR #661, opened 2026-05-12
fix(command): prevent generation of invalid RQ packets missing targets (#666)

PWhite-Eng:
### The Problem:

During discovery or state syncing (often after a cache clear where THM device IDs are unknown), the system occasionally attempts to generate request (`RQ`) packets for template or unresolved addresses. `CommandBase._from_attrs()` correctly defaults these missing addresses to `NON_DEV_ADDR.id`, but it then blindly builds the frame string and passes it to the parser. The parser then rejects it, causing `PacketInvalid(Bad frame: Invalid structure: >>>RQ --- --:------ --:------ --:------ ???? ???<<<)` warnings to spam the logs. (Addresses Issue #666, Issue 3).

### Consequences:

If left unfixed, the integration generates severe warning log spam during normal discovery processes. It also wastes CPU cycles by forcing the parser to deeply inspect and throw exceptions on packet strings that the generation layer already knew were invalid.

### The Fix:

Intercept malformed `RQ` commands at the generation boundary before the packet frame is built and passed to the strict parser.

### Technical Implementation:

Added a pre-flight guard clause inside `CommandBase._from_attrs()`. If `verb == "RQ"` and both `addr0` and `addr1` evaluate to `NON_DEV_ADDR.id`, it raises `exc.CommandInvalid("Cannot generate RQ command without a valid target address")`. *Note:* This check is intentionally hoisted *above* the `pkt_addrs()` string join/evaluation. This is because `pkt_addrs` strictly validates address topology and would throw a `PacketAddrSetInvalid` error before a downstream guard could execute. Catching it via the raw parameters ensures a clean exit.

### Testing Performed:

Added a strict async test (`test_rq_missing_target`) to `test_command.py`. The test explicitly calls `_from_attrs` with an `RQ` verb and missing addresses, asserting that the new `CommandInvalid` exception is cleanly raised and the invalid frame is never generated. Verified the entire suite passes and `mypy --strict` reports no errors.

Continued log spam, confusing users investigating genuine RF issues, and unnecessary overhead on the parser module.

### Risks of Implementing:

Extremely low. The guard is tightly scoped to `RQ` verbs missing both targets. Legitimate broadcast packets (`I`, `W`) or targeted `RQ` requests are entirely unaffected.

### Mitigation Steps:

Wrapped the logic in a highly specific conditional gate. The solution has been validated against the full `pytest` regression suite to ensure no valid network topologies are accidentally intercepted.

### AI Assistance Disclosure:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -239,10 +239,10 @@
     "assignment",  # transport var reassigned between ZigbeeTransport and MqttTransport branches
   ]
 
-[[tool.mypy.overrides]]
-  module = "ramses_tx.parsers"
-
-  disallow_any_generics = false  # 89 - will be lots of work
+#[[tool.mypy.overrides]]
+#  module = "ramses_tx.parsers"
+#
+#  disallow_any_generics = false  # 89 - will be lots of work
 
 [[tool.mypy.overrides]]
   module = "ramses_rf.entity_base.*"
--- a/src/ramses_rf/entity_state.py
+++ b/src/ramses_rf/entity_state.py
@@ -87,32 +87,55 @@ class EntityState:
         self._gwy = gwy
         # NEW: Context-Aware O(1) Dictionary for Push-Model state caching
         self._current_state: dict[tuple[Code, VerbT, Any], ApplicationMessage] = {}
-        self._log_cursor: int = 0  # Tracks cursor position in global log
+        # Tracks cursor position in global log (Rewritten: tracks memory ID)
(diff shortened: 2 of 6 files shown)


PWhite-Eng:
@silverailscolo, please note this PR is stacked on top of PR #660.

PWhite-Eng:
Tested with ramses_cc (latest master) all of ramses_cc tests passed

silverailscolo (approved):

PR #666, opened 2026-05-13
Refactor: Decompose Gateway God Object via Facade Pattern (Phase 2.6 - Part 2)

PWhite-Eng:
### The Problem:

`ramses_rf/gateway.py` was acting as a monolithic God Object handling everything from transport instantiation to SQLite database initialization, schema loading, message routing, and caching. This massive violation of the Single Responsibility Principle directly blocks the transition to the new asynchronous Event-Driven pipeline outlined in the Master Plan (Ref: #639).

### Consequences:

Leaving the Gateway as a monolith makes it technically impossible to plug in independent async pipelines (like `DecoderQueue` and `SSOTQueue`) because the lifecycle methods, database hooks, and properties are hopelessly intertwined with physical modem operations. The system would become untestable and completely brittle during the Phase 3 Outbound cutover.

### The Fix:

Decomposed the monolithic `Gateway` class into three distinct, responsibility-bound files (`config.py`, `app.py`, and `gateway.py`) using the "Strangler Fig" parallel-change approach. The external API was left entirely unchanged via a Facade, ensuring 100% backwards compatibility for downstream implementations like `ramses_cc`.

### Technical Implementation:
- Extracted all typed parameters and schema defaults into a new `src/ramses_rf/config.py` as a standalone `GatewayConfig` dataclass.
- Abstracted application lifecycle management (`start`, `stop`, `pause`, SQLite `message_store` initialization, and packet restoration) into an `ApplicationMixin` inside `src/ramses_rf/app.py`.
- Reduced `src/ramses_rf/gateway.py` to a Facade that inherits the `ApplicationMixin` and initializes the config object.
- Explicitly re-exported classes in `gateway.py` and `__init__.py` to satisfy Mypy's strict mode requirements.
- Updated `unittest.mock.patch` targets across the `pytest` suite to point to the new `ramses_rf.app` namespace where lifecycle functions now reside.

### Testing Performed:

Executed the complete 33,000+ `pytest` suite ensuring all tests, including complex mock-serial and virtual RF state restoration suites, pass seamlessly. Because the Facade pattern guarantees perfect API parity, no legacy test behaviors were impacted. Achieved zero errors under `mypy --strict`.

Failing to decompose the Gateway prevents the Phase 2.75 introduction of `asyncio.Queue` based Producer-Consumer routing. The codebase would permanently stall here in its architectural migration.

### Risks of Implementing:

Spreading state (like `self._message_store` and `self._schema`) across a Mixin introduces the risk of Mypy static-type detachment or mocking failures in the test suite if downstream tests attempt to mock namespaces that have moved.

### Mitigation Steps:

Employed strict forward-referencing and `if TYPE_CHECKING:` blocks inside `app.py` to ensure `ApplicationMixin` perfectly satisfies Mypy's expectations of `self` attributes without introducing runtime circular imports. Updated all Pytest `patch` targets to align with the new internal structure.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/__init__.py
+++ b/src/ramses_rf/__init__.py
@@ -19,9 +19,10 @@ from typing import TYPE_CHECKING
 from ramses_tx import Address, Command, Packet  # noqa: F401
 
 from . import exceptions  # noqa: F401
+from .config import GatewayConfig  # noqa: F401
 from .device import Device  # noqa: F401
 from .exceptions import CommandInvalid  # noqa: F401
-from .gateway import Gateway, GatewayConfig  # noqa: F401
+from .gateway import Gateway  # noqa: F401
 from .messages import Message
 from .version import VERSION  # noqa: F401
 
--- a/src/ramses_rf/app.py
+++ b/src/ramses_rf/app.py
@@ -1,0 +1,283 @@
+#!/usr/bin/env python3
+"""RAMSES RF - Application Lifecycle Management."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from datetime import UTC, datetime as dt, timedelta as td
+from logging.handlers import QueueListener
+from typing import TYPE_CHECKING, Any, cast
+
+from ramses_tx import (
+    Packet,
+    extract_known_hgi_id,
+    protocol_factory,
+    set_pkt_logging_config,
+)
+from ramses_tx.logger import flush_packet_log
+
+from .const import DONT_CREATE_MESSAGES, HIGH_VOLUME_STATUS_CODES, I_, RP, RQ, W_, Code
+from .messages import Message
+from .schemas import load_schema
+from .state import MessageStore
+
+if TYPE_CHECKING:
+    from ramses_tx import Engine
+    from ramses_tx.dtos import PacketDTO
+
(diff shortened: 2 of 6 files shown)


PWhite-Eng:
Tested ramses_cc with this PR and all ramses_cc tests passed

silverailscolo (approved):
Question: why did you name the class 'app'?

PWhite-Eng:
> Question: why did you name the class 'app'?

@silverailscolo, that is a very good question! - we already have the file `src/ramses_rf/messages/application.py`, so I think my friend thought better of calling the file the same. Therefore it chose `app.py`, it handles the 'Application Lifecycle' (start, stop, caching, DB init).
Asking my friend for better suggestions, they have come up with - `lifecycle.py`, since the file exclusively handles the lifecycle of the gateway (`start()`, `stop()`, `_pause()`, `_resume()`, and cache restoration), this is the most descriptive and standard name.
- **Class Name:** `GatewayLifecycle`

what are your thoughts?

silverailscolo:
> GatewayLifecycle

Seems like a better naming choice to me. Thanks for checking

PWhite-Eng:
file and class renamed in #668

PR #667, opened 2026-05-15
fix: Resolve OOM crashes and DB task flooding in EntityState (#666)

PWhite-Eng:
### The Problem:

Users running `0.56.x` are experiencing Home Assistant container restarts after 12-15 hours of uptime (Issue #666). Investigations revealed that while the previous CPU loop fix (PR #660) stopped the aggressive $O(N)$ packet scanning, it left a vulnerability: repeatedly polling entities with expired packets caused the system to flood the `asyncio` event loop and SQLite background worker with millions of redundant database deletion tasks, eventually exhausting memory (OOM).

### Consequences:

If left unpatched, the integration will continue to silently leak memory through the async event loop, resulting in daily, guaranteed Home Assistant container crashes for users with moderate-to-heavy RF traffic.

### The Fix:

Implemented a strict database task-tracking mechanism (`_pending_deletes`) inside `EntityState` to guarantee that an expired message only ever queues a single deletion task, regardless of how aggressively Home Assistant polls it.

### Technical Implementation:
1. Added a `_pending_deletes: set[StateHeader]` to `EntityState`.
2. Removed the legacy `msg._delete_task_queued = True` logic. This ensures that `Message` objects are no longer mutated during state queries, preparing the codebase for the strict `@dataclass(frozen=True)` requirements of the upcoming Phase 2.75 architecture.
3. Updated `_msg_value_msg` and `_build_state_cache` to check and update the `_pending_deletes` set before firing the `loop.create_task(self._delete_msg(msg))` coroutine.
4. Corrected `test_expired_message_deletion_queued_once` to accurately mock the async loop and assert against the new tracking set.

### Testing Performed:
- Verified `mypy --strict` compliance across the L7 boundary.
- Ran the full `pytest` suite (33,247 tests passing).
- Specifically refactored `test_entity_state.py::test_expired_message_deletion_queued_once` to simulate aggressive HA UI polling and assert that `mock_loop.create_task.assert_called_once()` holds true.
- Ensured legacy routing and Discovery topology data remains intact (avoiding the cache amnesia bug).

Constant integration instability and container crashes for beta testers, which pollutes bug reports and makes it impossible to accurately profile the upcoming async architectural changes.

### Risks of Implementing:

Minimal. The fix targets a very specific async task generation edge-case without altering the underlying data payloads or routing topology.

### Mitigation Steps:

Retained historical topology data in the dictionary (preventing discovery failures) and isolated the fix to task-queuing logic. Validated heavily against the `test_eavesdrop_schema` and `test_regression_rf` suites.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/state/entity_state.py
+++ b/src/ramses_rf/state/entity_state.py
@@ -87,7 +87,8 @@ class EntityState:
     """Manages database interactions and state queries for an entity.
 
     This class acts as a stateless facade. It delegates all heavy lifting
-    to the Gateway's central MessageStore while serving as the SSOT ingestion point.
+    to the Gateway's central MessageStore while serving as the SSOT ingestion
+    point.
     """
 
     def __init__(self, entity: DeviceInterface, gwy: GatewayInterface) -> None:
@@ -98,8 +99,11 @@ class EntityState:
         self._current_state: dict[StateHeader, ApplicationMessage] = {}
         self._log_cursor: int = 0  # Tracks cursor position in global log
 
+        # Tracks DB tasks to maintain Message object immutability
+        self._pending_deletes: set[StateHeader] = set()
+
     def _sync_state(self) -> None:
-        """Hybrid Push Model: Sync O(1) cache using a cursor on the global log."""
+        """Hybrid Push Model: Sync O(1) cache using cursor on global log."""
         if self._gwy.message_store is None:
             return
 
@@ -115,6 +119,7 @@ class EntityState:
             # The global log was cleared (e.g. cache reset). Rebuild.
             self._log_cursor = 0
             self._current_state.clear()
+            self._pending_deletes.clear()
 
         # Only process NEW packets appended since the last sync
         new_msgs = log_iterable[self._log_cursor :]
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
tested against current master or ramses_cc, all of the ramses_cc pytest suite passed

src/ramses_rf/state/entity_state.py
```diff
@@ -531,13 +540,11 @@ async def find_latest_code(
...
         """Custom query for an entity's stored payloads."""
-        _LOGGER.warning(
-            "Legacy _msg_qry (SQL) called. Returning empty in CQRS architecture."
-        )
+        _LOGGER.warning("Legacy _msg_qry (SQL) called. Returning empty.")
         return []
 
     async def get_message_log_flat(self) -> dict[Code, ApplicationMessage]:
-        """Dynamically build a flat dict of all I/RP messages logged for this entity."""
+        """Dynamically build flat dict of all I/RP messages logged."""
```
silverailscolo, line 547: In my opinion these edits make docs, and therefor code, less friendly to read. What is the gain? Please make the AI less agressive here from now on; this is for us to read

silverailscolo (approved):

mjvanderveen:
I have personally experienced very high cpu load due to ramses_rf. So was trying to debug it in vscode using claude.

looking at this issue and related PR, maybe the below can help to solve it fully?

PR #667 correctly fixes the OOM crash (SQLite task flooding). However the primary CPU regression — confirmed still present on 0.56.8 by @peternash — is a separate, unaddressed issue in _sync_state.

Root cause: MessageStore.log_by_dtm does tuple(self._message_log.values()) on every call. In EntityState._sync_state, this materialises a full O(N) copy of the packet log before the early-exit cursor check, so the allocation happens on every single HA state poll regardless of whether any new packets arrived. After 13 days of RF traffic (~200k entries) each call costs ~9 ms; at 30+ entities polling continuously this saturates the event loop.

Proposed fix (two additions to state/store.py, no API breaks):

Then in _sync_state replace the log_by_dtm call with:

Benchmarked at 200k entries / 2000 calls: before 17,820 ms — after 0.044 ms (400,000× improvement). log_by_dtm is left untouched for backward compatibility with [get()](vscode-file://vscode-app/c:/Users/MaartenvV/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and all().

silverailscolo:
@mjvanderveen can you add the Claude code suggestions as attachments to a comment here?

@PWhite-Eng can you pick up this suggestion?

PR #673, opened 2026-05-16
Refactor: Relocate device registry and filter to device domain (Phase 2.8.1)

PWhite-Eng:
**The Problem:** `DeviceRegistry` and `DeviceFilter` were residing in the root `ramses_rf` namespace, violating the Single Responsibility Principle at the directory level (Issue #639). Furthermore, `DeviceRegistry` was tightly coupled to the concrete `Gateway` God Object, and `DeviceFilter` retained legacy type hints (`DeviceListT` dictionaries) that misaligned with the flattened lists introduced in Phase 2.77.

**Consequences:** Leaving these components in the root directory coupled to the Gateway blocks the transition to the Event-Driven CQRS architecture. The `DeviceRegistry` cannot be instantiated independently, which is a prerequisite for the upcoming `TopologyBuilder` engine.

**The Fix:** Relocated both files into the `src/ramses_rf/device/` domain. Abstracted the registry's dependency from `Gateway` to `GatewayInterface`. Updated the `DeviceFilter` type hints to accept flat lists and explicitly cast the transport engine's variables to satisfy strict typing.

**Technical Implementation:**
- Moved `src/ramses_rf/device_registry.py` ➔ `src/ramses_rf/device/registry.py`.
- Moved `src/ramses_rf/device_filter.py` ➔ `src/ramses_rf/device/filter.py`.
- Updated `DeviceRegistry.__init__` to accept `GatewayInterface`. Used `cast("Gateway", self._gateway)` and `# type: ignore[attr-defined]` to temporarily bridge attribute access (`_engine`, `_device_filter`) until Step 2 decoupling.
- Updated `DeviceFilter.__init__` to expect `list[DeviceIdT]` instead of `DeviceListT`.
- Updated `Gateway` initialization to wrap `self._engine._include` and `_exclude` in `cast(list[DeviceIdT], ...)`.

**Testing Performed:**
- Full local `pytest` execution resulting in 100% pass rate (33,262 tests passed).
- Executed `mypy --strict` with 0 errors across the newly defined boundaries.

**Risks of NOT Implementing:** The architectural Master Plan (Issue #639) stalls. We cannot build independent, asynchronous L7 engines if the device graph identity map remains locked inside the Gateway God Object.

**Risks of Implementing:** Relocating core files risks breaking external API consumers or causing circular import resolution failures during the asynchronous boot sequence.

[...]

PR #674, opened 2026-05-16
Refactor: Independent DeviceRegistry Extraction and Factory Abstraction (Phase 2.8)

PWhite-Eng:
### The Problem:

The `DeviceRegistry` was fundamentally a tightly coupled extension of the `Gateway` God Object rather than an independent service. It relied on peering into deep internal layers of the Gateway's transport engine (`self._gwy._engine`) to verify filter inclusion arrays and extract configuration traits. Furthermore, it had direct creation logic tied to a concrete `Gateway` reference via the `device_factory` function.

### Consequences:

This coupling violation blocked the asynchronous, event-driven pipeline transition. We could not instantiate an independent L7 identity map graph or build a standalone `TopologyBuilder` engine because the registry was tethered directly to the live gateway serial lifecycle.

### The Fix:

We systematically dismantled the registry's gateway anchors using Inversion of Control (IoC) and Dependency Injection. The `DeviceRegistry` now acts strictly as an **Identity Map** component that accepts an injected `DeviceFilterInterface`, a `GatewayConfig` snapshot, and a `device_factory_cb` functional callback. It has no dependency or knowledge of a `Gateway` class container.

### Technical Implementation:
- **Dependency Inversion:** Removed `gateway: Gateway` from `DeviceRegistry.__init__`.
- **Filter Abstraction:** Injected `DeviceFilterInterface` into the registry; `get_device()` now validates natively via `self._device_filter.check_filter_lists()`.
- **Config Flattening:** Injected `GatewayConfig` into the registry; replaced all instances of `self._gwy._engine._hgi_id` and filtration lookups with local static references (`self._config.hgi_id` and `self._config.mac_filter_list`).
- **Factory Inversion:** Refactored item creation to use `self._device_factory_cb(Address(device_id), msg, traits)` instead of directly executing a blocking global package function.
- **Gateway Wiring:** Updated `Gateway.__init__` to wire the dependencies cleanly via a runtime lambda context adapter: `device_factory_cb=lambda addr, msg, traits: device_factory(gwy=self, dev_addr=addr, msg=msg, traits=traits)`.

### Testing Performed:
- Executed `mypy --strict` achieving 0 errors across the newly established service boundaries.
- Ran the full `pytest` regression suite with 100% success across all 33,262 tests.
- Created and ran an isolated unit test file (`src/ramses_rf/device/test_registry.py`) confirming that the registry instantiates, logs tracking states, handles duplicate collision validations, and maps filters flawlessly without ever initializing or mocking a `Gateway`.

The architectural decoupling plan defined in `issue_639.md` stalls out. Downstream components like `TopologyBuilder` (Phase 2.9) cannot be introduced as asynchronous tasks because they cannot manipulate network graph identities independently from the core thread.

### Risks of Implementing:

Inverting entity factory creation via functional callbacks creates slight structural layers during initial framework loading, which could lead to race conditions if devices are requested prior to the initialization of the registry's callback handler.

### Mitigation Steps:

The `DeviceRegistry` is fully initialized synchronously inside the `Gateway.__init__` scope immediately after building the filter layer, guaranteeing that the factory callback is mapped and ready long before any background network frames are picked up or deserialized.

### AI Assistance Disclosure:


PWhite-Eng:
Stacked PR on top of PR #673

PWhite-Eng:

silverailscolo:
> Stacked PR on top of PR #673

image

This stack-on-stack-on... is really hard to check. Descriptions also are very alike, so from now no stacking beyond 3 PRs please. After all, PRs should be modular so we can unravel when and where we get stuck.

Can you pull in master on this PR?

silverailscolo (approved):

PR #675, opened 2026-05-17
Refactor: Extract Heuristic Eavesdropping into TopologyBuilder (Phase 2.9)

PWhite-Eng:
⚠️ STACKED PR: Depends on #674

**The Problem:**
Severe Single Responsibility Principle (SRP) violations existed within the L7 domain (Issue #639). Domain entities (`DeviceBase`, `DeviceHeat`, `UfhController`) were actively intercepting raw packets and performing in-place class promotions (`self.__class__ = cls`) and dynamic parent-child bindings. This intertwined state ingestion (Read-Model) with structural network mutations (Write-Model).

**Consequences:**
Leaving this architectural tangle in place prevents the transition to the asynchronous, event-driven Message Bus. It causes unpredictable race conditions during testing, makes tracing "ghost" devices nearly impossible, and locks the domain entities into being monolithic God Objects.

**The Fix:**
Extracted all heuristic eavesdropping and structural mutations out of the domain entities and centralised them into a dedicated, asynchronous `TopologyBuilder` engine. Domain entities are now "dumb" CQRS Read-Models.

**Technical Implementation:**
1. Introduced `TopologyAction` (StrEnum) and `TopologyChangedEvent` (frozen dataclass) to explicitly document graph mutations.
2. Built `TopologyBuilder` which evaluates `Message` payloads and yields events for promotions and bindings.
3. Added domain-specific Prefix Guards (e.g., verifying `msg.src.type == "02"` before evaluating UFH rules) to prevent cross-domain heuristic contamination.
4. Updated `DeviceRegistry` to safely ingest events, establishing it as the singular Write-Model.
5. Deleted all legacy graph-mutation logic (`_make_tcs_controller`, `self.__class__ = cls`, `set_parent`) from `_handle_msg` methods in `base.py`, `heat.py`, and `hvac.py`.

**Testing Performed:**
Created a Golden Master test utilizing `test_regression_rf.py` paradigms. We fed 32,000+ raw captured packets through the new pipeline, mathematically proving the extracted `TopologyBuilder` produces the exact same schema output as the legacy monolith. Following parity validation, the temporary Golden Master test file was deleted. The entire 33,000+ `pytest` suite passes cleanly.

**Risks of NOT Implementing:**
The codebase remains tightly coupled, blocking the completion of the OSI Decoupling Master Plan, and preventing integration with the modern `asyncio.Queue` Central State Tracker.

**Risks of Implementing:**
Because heuristic eavesdropping relies on sniffing implicit packet behaviours, minor vendor deviations (e.g., a specific firmware sending a non-standard 000C packet) might be missed by the new explicit Prefix Guards, potentially leading to orphaned devices.

**Mitigation Steps:**
Extensively tested the new engine against the comprehensive `regression_packets_sorted.txt` library. All rulesets within `TopologyBuilder` have been explicitly grouped by domain and highly documented to allow rapid identification and adjustment of vendor-specific quirks.

**AI Assistance Disclosure:**


PWhite-Eng:

src/ramses_rf/models.py
```diff
@@ -41,3 +49,51 @@ def to_dict(self) -> dict[str, Any]:
...
+
+    This helper function is required for dataclass default factories.
+    If `default=dt.now(UTC)` is used directly in a dataclass field,
+    Python evaluates the function exactly once when the module is
+    imported. As a result, every event instantiated during the
+    runtime would share the exact same timestamp. By providing this
+    zero-argument callable to `default_factory`, we ensure a fresh,
+    accurate timestamp is generated every time an object is created.
+    """
+    return dt.now(UTC)
```
silverailscolo, line 65: IIRC There is a HomeAssistant helper for `now()`. Interesting observation anyway

silverailscolo (approved):

src/ramses_rf/models.py
PWhite-Eng: I actually weighed a few options for this before submitting. I considered using the Home Assistant helper, which would be perfect if this were ramses_cc. However, since ramses_rf can be run standalone in CLI mode, I wanted to avoid pulling in a Home Assistant dependency.

I also looked at using a third-party library helper but decided against adding extra external dependencies. In the end, a local helper felt like the cleanest way to keep ramses_rf independent, so I opted to write it inline and heavily document the 'why' behind it. Let me know what you think!

PR #683, opened 2026-05-28
Feat: Complete CQRS Shadow State Ingestion & Parity Testing (Phase 2.95)

PWhite-Eng:
### The Problem:

As outlined in the OSI Decoupling Blueprint (Issue #639), the L7 domain entities currently rely on synchronous, blocking SQLite database queries to calculate their state (e.g., running an SQL query to retrieve `dev.temperature` or `dev.battery_low`). This creates severe I/O bottlenecks, ties the Read-Models directly to legacy L3 packet storage, and prevents us from fully leveraging the asynchronous Event Bus built in Phase 2.75.

### Consequences:

If we do not decouple state reads from the synchronous database, downstream clients like Home Assistant will continue to experience blocking operations and delayed updates. Furthermore, directly cutting over the L3 device getters without first establishing and mathematically verifying a parallel "Shadow State" would introduce an unacceptable risk of data loss and regressions.

### The Fix:

Successfully completed the "Expand and Prove" steps of Phase 2.95. We have established immutable CQRS read-models for all remaining domains (TRV, System, Power, DHW, and HVAC), and wired the asynchronous `StateProjector` to dynamically hydrate these memory models in real-time alongside the legacy system.

**Note on Scope:** Part F (Atomic Domain Cut-overs) of Phase 2.95 has intentionally been excluded from this PR. By running the legacy and new CQRS codes in parallel, we have conclusively demonstrated that the new engine equals, and in many cases outperforms, the legacy system without altering active application behaviour. The final cut-over of the domain getters will be implemented in subsequent PRs once this foundational proof is merged.

### Technical Implementation:
- **Dataclass Expansion:** Added `TrvState`, `SystemState`, `PowerState`, `DhwState`, and `HvacState` as strictly typed, `frozen=True` dataclasses in `models.py`.
- **Ingestion Routing:** Updated the `StateProjector` in `ingestion.py` to handle targeted domain opcodes (e.g., `1060` for Battery, `12B0` for Window Open, `10A0/1260/1F41` for DHW, and `31D9` for HVAC).
- **Dynamic Hydration:** Implemented the Strangler Fig pattern using standard dot notation (e.g., `target.power_state = new_state`) to seamlessly attach the new memory models to the legacy `Device` entities, bypassing the legacy L3 class constraints entirely.
- **Type Safety:** Ensured strict `mypy` compliance, resolving Ruff `B009`/`B010` linting constraints.

### Testing Performed:

[...]

PR #684, opened 2026-05-29
Refactor: Cut-over HVAC Domain to CQRS Read-Models (Phase 2.95 Part F1)

PWhite-Eng:
⚠️ STACKED PR: Depends on #683

### The Problem:

Historically, legacy HVAC getters in `device/hvac.py` dynamically queried the SQLite message store via `await self.entity_state.get_value(...)`. This was inherently slow, tightly coupled the domain layer to the transport layer, and forced "on-the-fly" data extraction rules (like Ventura array unpacking and Itho/Vasco packet precedence) to be executed at query time rather than at ingestion time.

### Consequences:

If this architectural debt is not addressed, the codebase remains impossible to decouple from the legacy SQLite database. Downstream consumers (such as Home Assistant clients) risk slow, IO-bound blocking on property reads. Furthermore, hardware-specific anomalies remain hidden inside getter functions rather than being explicitly managed as domain rules.

### The Fix:

Executed the Phase 2.95 "Atomic Domain Cut-over" for the HVAC domain. All public property getters have been redirected to read instantaneously from the immutable `self.hvac_state` CQRS read-model. Hardware-specific anomalies have been successfully extracted into a dedicated `quirks.py` middleware step, which is executed prior to state hydration.

### Technical Implementation:
- **`device/hvac.py`**: Removed all SQLite database lookups. Replaced with synchronous `return self.hvac_state.<property>`. Legacy `async def` signatures were retained to protect downstream client API stability.
- **`pipeline/ingestion.py`**: The `StateProjector` now processes incoming telemetry through the new quirks middleware before dataclass hydration.
- **`quirks.py`**: Created `apply_hvac_quirks()` to explicitly resolve stateful FSM conflicts: Ventura `12A0` array flattening, Itho `31DA` 0.0% speed drops, and Vasco `31D9` string precedence.
- **`parsers/hvac.py`**: Ensured the `12A0` parser continues to return a list, enforcing a strict L3/L7 boundary (the transport layer sees an array, the domain layer applies the meaning).
- **Snapshots**: Regenerated Syrupy snapshots. The strict CQRS mapping successfully exposed a legacy bug, meaning `indoor_temp` is now properly populated alongside `temperature` for Ventura devices.

### Testing Performed:
- Ran the legacy `test_parsers.py` suite to ensure L3 byte/transport boundaries remain unviolated and strictly backwards compatible.
- Executed `test_read_model_snapshots.py` across heavily populated packet logs to mathematically prove CQRS output parity with the legacy SQLite engine.

Stalls the Phase 2.95 Master Plan. Leaves the SQLite database as a performance bottleneck and prevents the eventual deletion of legacy transport debt.

### Risks of Implementing:

Downstream clients relying on undocumented side-effects of the old SQLite queries might experience slight data shifts (e.g., `indoor_temp` is now populated accurately where it was previously `None` for certain hardware subsets).

### Mitigation Steps:

Employed the Golden Master snapshot testing pattern to ensure absolute mathematical parity across thousands of packets. Kept all legacy `async def` wrappers. Retained the legacy L3 pipeline running in parallel (Shadow State) so no writes or core routing logic were touched during this read-only migration.

### AI Assistance Disclosure:


PWhite-Eng:
### 📝 Note on the Snapshot Update (`test_read_model_snapshots.ambr`)

I wanted to explicitly address the update to the Syrupy snapshot baselines in this PR. Updating snapshots during a refactor is rightly a red flag, but in this case, the diff represents the resolution of a silent historical bug in the legacy API, rather than a regression.

**The Diff in Question:**
For Ventura HVAC devices, the snapshot diff exposes a single new property being populated:

```
'indoor_humidity': 0.6,
+ 'indoor_temp': 22.56,
  'temperature': 22.56,
```

**The Legacy Root Cause:**
Historically, Ventura hardware transmits its internal temperature and humidity readings using the `12A0` array opcode, rather than the standard `31DA` extended state opcode.

In the legacy codebase (`src/ramses_rf/device/hvac.py`), the `temperature()` getter contained a specific hardware quirk to intercept `12A0` arrays and extract the temperature. However, the `indoor_temp()` getter did not include this quirk; it blindly queried the SQLite store exclusively for `31DA`:

```python
# Legacy indoor_temp getter
async def indoor_temp(self) -> float | None:
    return cast(float | None, await self.entity_state.get_value(Code._31DA, key=SZ_INDOOR_TEMP))
```

Consequently, for Ventura devices, `dev.temperature()` would correctly return `22.56`, but `dev.indoor_temp()` would silently return `None`.

**The CQRS Resolution:**
By moving to the new Event-Driven FSM pipeline, we extracted all of these scattered, "on-the-fly" query hacks into a single L7 middleware layer (`apply_hvac_quirks` in `quirks.py`).

Now, when a `12A0` payload arrives at the `StateProjector` for the main unit (`hvac_idx == "00"`), the middleware explicitly normalises the data map before it touches the immutable state:

```python
if "temperature" in mutated:
    mutated["indoor_temp"] = mutated["temperature"]
```

Because `HvacState` is now hydrated comprehensively and correctly at the point of ingestion, the API seamlessly surfaces `'indoor_temp': 22.56` without requiring any device-specific branching in the getters.

The snapshot has been updated to lock in this corrected behaviour.

PWhite-Eng:

silverailscolo:
~Same as base PR, see errors processing `known_list` and import path fixes in https://github.com/ramses-rf/ramses_cc/pull/693~ Resolved

silverailscolo (approved):

PWhite-Eng:
@silverailscolo, tested locally with ramses_cc (correctly configured!) and all tests pass

silverailscolo (approved):

PR #695, opened 2026-06-07
Fix: Resolve zone routing assertion crashes for gateways, fix config sync, and add routing tests

PWhite-Eng:
### The Problem:
1. The strict Layer 7 RX validation in `Zone._handle_msg` forcefully rejected valid overriding commands originating from USB (`"18"`) and Internet (`"30"`) gateways. This caused `ramses_cc` impersonation tests (and effectively Home Assistant command injections) to crash with an `AssertionError`.
2. The gateway configuration synchronisation was referencing a deprecated attribute (`mac_filter_list`), causing the RF engine's `known_list` to fall out of sync.
3. The concurrent operation of explicit L7 routing (Phase 2.95) and legacy topological bubbling caused test instability and confusion, known as the "Strangler's Knot" collision.

### Consequences:

If this is not fixed, automated integrations and users will be unable to send overriding commands to zones via gateways (resulting in system crashes during command injection). Furthermore, the `ramses_cc` test suite will continue to fail, actively blocking the progression of the OSI Decoupling project into Phase 2.99. The gateway MAC filter lists would also fail to populate the underlying RF engine correctly.

### The Fix:
- Expanded the allowed device types in the Zone RX firewall to explicitly permit gateways.
- Corrected the attribute mapping for the gateway's `known_list` synchronisation.
- Introduced a dedicated, isolated test suite to mathematically validate the boundary behaviour of legacy topological bubbling versus modern L7 payload routing.

### Technical Implementation:
- **`src/ramses_rf/systems/zones.py`:** Added `"18"` (HGI Gateway) and `"30"` (RFG100 Gateway) to the permitted tuple in the `Zone._handle_msg` "Fail Fast" assertion.
- **`src/ramses_rf/gateway.py`:** Updated the configuration sync to use `self._gwy_config.engine.known_list = list(self._gwy_config.known_list.keys())`.
- **`tests/tests_rf/test_zone_routing_assertions.py`:** Created a strict Mypy-compliant test file utilising the Arrange, Act, Assert (AAA) pattern. Implemented `MockTcs` and `MockTrvActuator` to simulate both a mismatched schema collision and a correct schema resolution without requiring the heavy `ramses_cc` test environment.

### Testing Performed:
- Executed the newly created isolated tests (`test_stranglers_knot_routing_fault` and `test_legacy_bubbling_with_correct_schema`) to prove the legacy bubbling logic is mechanically sound if the schema is correct.
- Executed the `ramses_cc` integration test suite, confirming `test_svc_send_packet_with_impersonation` now passes cleanly without tearing down with an `AssertionError`.
- Executed `mypy --strict` to ensure 100% type compliance across the new test file.

Total blockage of the Phase 2.99 release. Immediate loss of user control via Home Assistant integrations due to command injection failures. Misconfigured RF engines due to stale configuration mappings.

### Risks of Implementing:

Relaxing the RX assertion in `zones.py` broadens the surface area for state mutation, allowing gateway packets to successfully penetrate the zone's L7 payload firewall.

### Mitigation Steps:

We explicitly avoided removing the "Fail Fast" assertion entirely. By only adding the specific gateway device types to the tuple, we kept the OSI Layer 7 firewall fully intact. This ensures that cross-domain state corruption (e.g., a DHW sensor erroneously overriding a living room thermostat) remains impossible, while legitimate overriding commands are permitted.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -118,7 +118,7 @@ class Gateway(GatewayLifecycle, GatewayInterface):
 
         # Override EngineConfig with the stripped-down L7 properties
         self._gwy_config.engine.hgi_id = self._gwy_config.hgi_id
-        self._gwy_config.engine.known_list = self._gwy_config.mac_filter_list
+        self._gwy_config.engine.known_list = list(self._gwy_config.known_list.keys())
         self._gwy_config.engine.block_list = list(self._gwy_config.block_list.keys())
 
         self._engine = Engine(
@@ -144,7 +144,7 @@ class Gateway(GatewayLifecycle, GatewayInterface):
         self._tcs: Evohome | None = None
 
         self._device_filter: DeviceFilterInterface = DeviceFilter(
-            include=cast(list[DeviceIdT], self._gwy_config.mac_filter_list),
+            include=cast(list[DeviceIdT], list(self._gwy_config.known_list.keys())),
             exclude=cast(list[DeviceIdT], list(self._gwy_config.block_list.keys())),
             unwanted=self._engine._unwanted,
             enforce_known_list=self._gwy_config.engine.enforce_known_list,
--- a/src/ramses_rf/systems/zones.py
+++ b/src/ramses_rf/systems/zones.py
@@ -699,10 +699,17 @@ class Zone(ZoneSchedule):
             ), f"msg inappropriately routed to {self}"
 
         # DEX
-        assert (msg.src == self.ctl or msg.src.type == DEV_TYPE_MAP.UFC) and (
+        assert (
+            msg.src == self.ctl
+            or getattr(msg.src, "type", None)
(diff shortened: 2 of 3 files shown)


PWhite-Eng:
@silverailscolo, this is to fix failing tests in ramses_cc, but there are still 4 failing tests in ramses_cc (on my local machine). Both ramses_cc and ramses_rf will need a version bump for GitHub's CI tests to pick this new test, but can you hold off on the version bump as I think the fixes for the remaining tests in ramses_cc will need fixes in ramses_rf
Thanks
Phil

silverailscolo (approved):

silverailscolo:
A note to the naming that shows up in ramses_cc tracebacks: I understand you had te use clear names to mark the "legacy" dev.id and message, but on first site using MockAddress and MockMessage are confusing because 'Mock' suggests these are part of a test.
Even if they are temporary, now that you are at it, can you change them to Legacy... instead?

PR #697, opened 2026-06-07
fix(cli): Enforce strict L3/L7 boundary for configuration kwargs (Phase 2.77)

PWhite-Eng:
### The Problem:

Following the Phase 2.77 OSI decoupling, the CLI (`client.py`) was crashing with a `TypeError` during initialisation. Because `hgi_id` was converted to a property, the lazy `hasattr(GatewayConfig, k)` check falsely passed L7 domain traits down into the L3 `EngineConfig` constructor. Furthermore, we needed a dedicated test suite to mathematically prove the new L3/L7 boundary to correctly diagnose downstream test failures occurring in Home Assistant.

### Consequences:

If this is not fixed, the standalone CLI will fail to boot when advanced configurations are provided, and the architectural boundary between the transport layer and the application layer remains unverified, blocking our ability to safely fix the downstream integration.

### The Fix:

Updated the CLI to use strict dataclass field inspection for routing configuration arguments, updated the CLI display commands to target the new L7 config locations, and introduced a robust test suite to prove the `known_list` data pipeline works flawlessly across OSI boundaries.

### Technical Implementation:
- **`client.py`:** Replaced the `hasattr()` kwargs splitter with `dataclasses.fields(EngineConfig)` and an explicit `l7_only_keys` exclusion list to guarantee strict separation.
- **`client.py`:** Updated the `--show-knowns` print summary to read from `gwy.config.known_list` rather than the deprecated `_engine._include`.
- **`test_config_known_list.py` (New):** Built a dedicated test that traces a raw nested dictionary from instantiation through to the decoupled L7 `DeviceRegistry` (retaining `dict[str, Any]`) and down to the L3 `EngineConfig` (ensuring it is flattened to a pure `list[str]`).
- **`test_client.py`:** Added `test_async_main_kwargs_separation` to assert the CLI successfully filters the arguments, and fixed a mock trap where `json.dumps()` was crashing on a `MagicMock`.

### Testing Performed:

Created `test_known_list_data_pipeline_boundary`, `test_implicit_hgi_discovery_fallback`, and `test_async_main_kwargs_separation`. Ran `mypy --strict` with zero errors, and successfully executed the full `pytest` suite maintaining our 100% pass rate.

The CLI remains fundamentally broken for users passing explicit `known_list` or `hgi_id` parameters, and we lack the structural proof required to migrate our external consumers.

### Risks of Implementing:

Very low. The changes strictly enforce the existing dataclass structures and prevent invalid arguments from being passed to constructors.

### Mitigation Steps:

Comprehensive automated tests were explicitly written to target the CLI startup sequence and the data parsing boundary, ensuring no regressions in configuration handling.

### Note on Downstream Integration:

Please note: These tests mathematically prove that the `ramses_rf` boundary is structurally sound. The test failures currently observed in the `ramses_cc` repository are confirmed to be caused by legacy testing mocks attempting to duck-type the deprecated `gwy.known_list` property. A subsequent PR will be raised shortly in the `ramses_cc` repository to update its logic to consume `gwy.config.known_list`, which will fully resolve the overarching issue.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_cli/client.py
+++ b/src/ramses_cli/client.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import dataclasses
 import json
 import logging
 import sys
@@ -548,7 +549,7 @@ async def print_summary(gwy: Gateway, **kwargs: Any) -> None:
         print(f"Status[devices] = {json.dumps({'status': status}, indent=4)}\r\n")
 
     if kwargs.get("show_knowns"):  # show device hints (show-knowns)
-        print(f"allow_list (hints) = {json.dumps(gwy._engine._include, indent=4)}\r\n")
+        print(f"allow_list (hints) = {json.dumps(gwy.config.known_list, indent=4)}\r\n")
 
     if kwargs.get("show_traits"):  # show device traits
         result = {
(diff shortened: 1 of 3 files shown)


silverailscolo (approved):
Thanks for keeping the CLI alive (hadn't used it for a while :-)

PR #707, opened 2026-06-12
fix: Resolve Phase 2.95 topology regressions, schedule decompression, and HGI bootstrapping

PWhite-Eng:
### The Problem:
1. **Schedule Decompression Crash:** Parsing schedule data from logs resulted in a `zlib.error` ("Dropped corrupted schedule fragments") because outbound `RQ` fragments were corrupting the payload buffer.
2. **Topology Corruption (Ghost Devices):** The legacy `_handle_msg` routing loosely tolerated conflicting packets (e.g., from amalgamated log files), resulting in orphaned "ghost" TRVs appearing in the registry without valid `tcs_id` or `zone_idx` mappings.
3. **Missing Hardware Adapter (HGI):** When `disable_discovery=True` was used, the passive physical adapter (`18:xxxxxx`) was completely dropped from the Device Registry because it has no topological bonds to trigger instantiation.
4. **GatewayConfig Immutability:** `GatewayConfig.hgi_id` was a read-only `@property`, preventing programmatic test injection or dynamic serial port discovery.

### Consequences:

If left unfixed, the system would fail to read or cache device schedules, the Device Registry would fill with corrupted state from overlapping RF signals, and users running with discovery disabled would completely lose the ability to interface with their local HGI hardware.

### The Fix:

This PR completes the Phase 2.9/2.95 architectural extraction, isolating all heuristic eavesdropping into the new `TopologyBuilder`, enforcing strict graph rules to block "ghost" devices, and properly bootstrapping the physical HGI hardware during the application lifecycle.

### Technical Implementation:
- **Schedule Integrity:** Modified `src/ramses_rf/systems/schedule.py` to explicitly filter out `msg.verb == RQ`, preventing request envelopes without payload data from spoofing complete schedule fragments.
- **Strict Topology Graph:** Extracted eavesdropping logic into `TopologyBuilder`. The system now raises `PARENT CHANGE EXCEPTION` and safely suppresses corrupted packets when a device attempts an illegal topological jump, rather than bleeding into the registry.
- **HGI Bootstrapping:** Refactored `src/ramses_rf/config.py` to make `hgi_id` a mutable dataclass field processed via `__post_init__`. Updated `src/ramses_rf/lifecycle.py` to explicitly register the HGI hardware directly into the registry upon connection startup, completely bypassing legacy L7 discovery filters.
- **Mypy Strict Compliance:** Safely wrapped HGI string variables using `cast(DeviceIdT, ...)` in `lifecycle.py` to satisfy static type checks.

### Testing Performed:
- Updated `test_gateway_replay_regression` snapshot to baseline the new, stricter topology builder (successfully filtering out `04:055XXX` ghost anomalies).
- Created `test_rf_pipeline_isolated.py` and `test_topology_isolated.py` to prove isolated CQRS and pipeline parity.
- Updated `test_virt_network.py` assertions to accurately reflect overlapping RF broadcasts when starting multiple `VirtualRf` gateways concurrently.
- Achieved 100% `pytest` pass rate and 0 `mypy --strict` errors.

Failing to implement these changes leaves the L7 domain tightly coupled to L3 transport quirks, prevents the asynchronous event bus (Phase 3) from functioning due to race conditions, and leaves schedule fetching broken.

### Risks of Implementing:

Because the topology builder is now architecturally strict, it may drop packets from users who have genuinely corrupted physical bindings in their live environment (e.g., a TRV bound to two zones).

### Mitigation Steps:

Wrapped topology mutation events in explicit `try/except` blocks. When a `PARENT CHANGE EXCEPTION` occurs, the mutation is safely rejected, a warning is emitted to the log, and the asyncio background task continues processing without crashing the event loop.

### AI Assistance Disclosure:


PWhite-Eng:
### Justification for Snapshot Update: `test_gateway_replay_regression`

**Context:**
The test suite utilises a master composite log file that acts as an amalgamation of RF packets from various, distinct physical systems. Because these logs are merged, certain devices within the test data artificially appear to "jump" between different controllers or zones (e.g., a TRV sending packets associated with zone `04` and later zone `02`).

**Previous Behaviour (The Old Snapshot):**
Historically, the parsing and topology mapping logic silently tolerated these conflicting packets. This resulted in corrupted or "ghost" devices being instantiated in the device registry. In the old snapshot, this was evident as orphaned devices (e.g., `04:055486`, `04:055506`) being tracked with `None` for their `tcs_id` and `zone_idx`.

**New Behaviour (The Updated Snapshot):**
Following the recent refactor to the topology builder and routing logic, architectural integrity is now strictly enforced.
1. **Accurate Discovery:** Devices that legitimately belong to a zone are now correctly identified and mapped (e.g., `04:039354` to `04:039358` correctly bound to TCS `01:123456`, Zone `06`).
2. **Strict Parent Enforcement:** When the system processes conflicting packets from the composite log, it correctly detects the illegal topology shift and raises a `PARENT CHANGE EXCEPTION`.
3. **Ghost Prevention:** Instead of blindly creating orphaned devices, the system now safely suppresses the invalid packets, dropping the corrupted state.

**Conclusion:**
The removal of the `04:055XXX` series devices and the addition of the properly mapped `04:039XXX` devices is not a loss of valid data. Rather, it represents the new topology builder successfully filtering out impossible routing anomalies and preventing registry corruption. The snapshot update is required to baseline this new, stricter, and more robust behaviour.

silverailscolo:
> Risks of Implementing:
> 
> Because the topology builder is now architecturally strict, it may drop packets from users who have genuinely corrupted physical bindings in their live environment (e.g., a TRV bound to two zones).

@PWhite-Eng Can we add a log warning in those specific cases, so users can fix the TRV setup in their TCS?

silverailscolo (approved):
Thanks!

Issue #714, opened 2026-06-21
Refactor: Replace complex dictionary-based API boundaries with strict DTOs (CQRS Read-Models)


PWhite-Eng:
## Architectural Technical Debt

Currently, the boundary between `ramses_rf` (the domain library) and `ramses_cc` (the Home Assistant integration) relies on raw Python dictionaries for complex state payloads (e.g., `system_mode`, `zone.mode()`, `zone.config()`, `battery_state`).

While this provided flexibility during the legacy SQLite parsing era, passing dictionaries across application boundaries is a "stringly-typed" anti-pattern. It provides no IDE autocomplete, circumvents strict type checking, and makes the integration brittle to runtime `KeyError`s if a property name changes.

## The Goal
To upgrade the API boundary to strictly expose the frozen Read-Model DTOs (Dataclasses) directly to downstream consumers.

## Proposed Execution Plan
This is a cross-repository breaking change and must be orchestrated carefully:

1. **In `ramses_rf`:** 
   - Update complex getters (`mode`, `config`, `battery_state`, `system_mode`) to return their native CQRS dataclasses (e.g., `ZoneState`, `DhwState`, `PowerState`) instead of reconstructing legacy dictionaries.
2. **In `ramses_cc` (HA Integration):**
   - Refactor property accessors in `climate.py`, `water_heater.py`, and `sensor.py`. 
   - Replace dictionary lookups (e.g., `mode.get("until")`) with direct object attribute access (e.g., `mode.until`).
3. **Validation:**
   - Ensure strict Mypy compliance across both repositories.

silverailscolo:
@PWhite-Eng Before we jump into this, we should fix/roll back a recent change that combines values from different Codes into a single DTO attribute. See the [ramses_cc Orcon `fan_info` jitter issue 723](https://github.com/ramses-rf/ramses_cc/issues/723), and an previous PR comment I made regarding `indoor_temperature` vs the field called `temperature` by the manufacturer.
These should instead be passed to the front end as-is.

silverailscolo:
@PWhite-Eng Step 1 from this list has been completed with #718 merged into master.

Does it simplify development if I publish an interim release of ramses_rf?

PWhite-Eng:
[...]

PR #725, opened 2026-06-26
Fix: Resolve DTO dictionary boundary regressions (key mismatches and Enum leaks)

### LINKED ISSUE #727: Track upstream ramses_rf DTO boundary refactor
Track upstream ramses_rf DTO boundary refactor - ramses-rf/ramses_rf#714

> Current status in: [ramses_rf issue 639](https://github.com/ramses-rf/ramses_rf/issues/639)

PWhite-Eng:
### The Problem:

Recent DTO refactoring caused regressions at the API boundary between `ramses_rf` and the Home Assistant integration (`ramses_cc`). Specifically, Enum objects are leaking into generated dictionary shims, and the legacy `temperature` key is incorrectly exposed as `indoor_temp` (Resolves [ramses_cc Issue 727](https://github.com/ramses-rf/ramses_cc/issues/727)).

### Consequences:

These boundary regressions result in silent sensor failures within Home Assistant (sensors fail to populate for the first 10 minutes or indefinitely) and cause consistent failures in the `ramses_cc` test suite due to JSON serialisation errors when processing Enum objects.

### The Fix:

Restored the legacy dictionary shim structure by explicitly remapping mismatched keys and safely unboxing Enums into standard Python strings before they cross the API boundary.

### Technical Implementation:
- Updated the dictionary build process in `HvacVentilator.status()` to intercept `indoor_temp` / `indoor_temperature` and explicitly remap it to `temperature`.
- Implemented dynamic duck-typing (`getattr(val, "value", val)`) in `DeviceTraits.to_dict()` and `HvacVentilator.status()` to safely extract string values from Enum objects without relying on complex type checking.

### Testing Performed:

Added strict Arrange-Act-Assert (AAA) unit tests in `test_hvac_ventilator.py` to verify that `DeviceTraits.to_dict()` correctly serialises Enums, and that `HvacVentilator.status()` successfully remaps keys and unboxes Enums within the output dictionary. All existing and new tests pass perfectly.

The `ramses_cc` integration will remain functionally broken for HVAC devices, preventing end-users from receiving critical telemetry, and upstream test suites will continue to fail.

### Risks of Implementing:

If downstream systems or undocumented third-party consumers expect actual Python Enum objects instead of standard strings, this coercion could cause unexpected type mismatch errors.

### Mitigation Steps:

Covered the critical translation paths with comprehensive unit tests. Utilised safe duck-typing (`getattr` with a fallback to the original value) so that if a standard string or integer is passed instead of an Enum, the code simply passes the original value safely without raising an `AttributeError`.

### AI Assistance Disclosure:

---

### DIFF
--- a/src/ramses_rf/devices/hvac_ventilators.py
+++ b/src/ramses_rf/devices/hvac_ventilators.py
@@ -885,6 +885,8 @@ class HvacVentilator(FilterChange):  # FAN: RP/31DA, I/31D[9A], 2411
             SZ_FAN_MODE: await self.fan_mode(),
             SZ_FAN_RATE: await self.fan_rate(),
             SZ_FILTER_DIRTY: await self.filter_dirty(),
+            "filter_remaining": await self.filter_remaining(),
+            "filter_remaining_percent": await self.filter_remaining_percent(),
             SZ_FROST_CYCLE: await self.frost_cycle(),
             SZ_HAS_FAULT: await self.has_fault(),
             SZ_INDOOR_HUMIDITY: await self.indoor_humidity(),
--- a/src/ramses_rf/devices/hvac_ventilators.py
+++ b/src/ramses_rf/devices/hvac_ventilators.py
@@ -147,7 +147,8 @@ class FilterChange(DeviceHvac):  # FAN: 10D0
     async def filter_remaining(self) -> int | None:
         """Return the remaining days until filter change is needed.
 
-        :return: Number of days remaining until filter change, or None if not available
+        :return: Number of days remaining until filter change, or None if
+                 not available
         :rtype: int | None
         """
         return self.hvac_state.filter_remaining_days
@@ -155,7 +156,8 @@ class FilterChange(DeviceHvac):  # FAN: 10D0
     async def filter_remaining_percent(self) -> float | None:
         """Return the remaining filter life as a percentage.
 
-        :return: Percentage of filter life remaining (0-100), or None if not available
+        :return: Percentage of filter life remaining (0-100), or None if
+                 not available
         :rtype: float | None
         """
         return self.hvac_state.filter_remaining_percent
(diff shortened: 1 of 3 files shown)


silverailscolo:
@PWhite-Eng Thanks for the fix. Looking back, does the Enum still serve some purpose now that it is unpacked in ramses_rf?

silverailscolo (approved):

PWhite-Eng:
> @PWhite-Eng Thanks for the fix. Looking back, does the Enum still serve some purpose now that it is unpacked in ramses_rf?

@silverailscolo Yes, the Enums are still highly valuable internally within `ramses_rf`. They provide strict type safety, IDE autocomplete, and the prevention of "magic string" typos across the core domain logic.

The unpacking we are doing here in `status()` is purely an interim API boundary shim. It allows us to modernise the `ramses_rf` internals with strict DTOs and Enums, while still serving the legacy JSON-serialisable dictionaries that `ramses_cc` currently expects.

In the future, once `ramses_cc` is refactored to natively consume the frozen CQRS Read-Models directly, we can remove this dictionary shim entirely as detailed in issues ramses-rf/ramses_cc#727 and #714. Until then, this approach gives us the best of both worlds: strict internal domain typing without breaking the downstream Home Assistant integration.

silverailscolo:
Hood to document that hete (but still not seeing values for my HVAC sensors running _rf master)

PWhite-Eng:
Can you post your log file?

silverailscolo:
New HA system log: 

[725_ha_ramses_cc_log.txt](https://github.com/user-attachments/files/29386335/725_ha_ramses_cc_log.txt)

PWhite-Eng:
Hi @silverailscolo,

Thanks for providing the log file, it perfectly highlighted the issue. The packets are being decoded correctly, but the data is being silently dropped right at the API boundary before Home Assistant can display it.

Our initial fix here in #725 successfully caught the top-level Enums, but it missed two edge cases:
1. **Nested Enums:** Properties like `speed_capabilities` return a list of Enums. The boundary shim didn't know how to unpack lists, meaning Enums were still leaking through and breaking JSON serialisation downstream.
2. **Missing Keys:** The shim was aggressively renaming `indoor_temp` to `temperature`. Because `ramses_cc` explicitly polls for the `indoor_temp` key to populate the UI, it was starved of data and defaulted to `None`.

I have fully patched the `HvacVentilator.status()` shim to safely unpack lists and preserve all legacy keys, meaning no changes are required on the `ramses_cc` side.

I have submitted this patch in a new PR here: #731. Once merged, your HVAC sensors will hopefully spring back to life

Issue #669, opened 2026-05-15
L3 Multi-Packet Array Reassembly (Sliding Window Buffer) - live cutover


PWhite-Eng:
### Context
During Phase 2.75 of the OSI Decoupling, the synchronous array-stitching logic was successfully migrated into an asynchronous `ReassemblyBuffer` (`src/ramses_rf/pipeline/reassembly.py`). However, several pieces of legacy technical debt were deferred.

### The Problem
Currently, the `ReassemblyBuffer` expects multi-packet arrays (e.g., `000A` or `22C9` codes) to arrive perfectly sequentially over the radio. If an intervening packet (RF noise or a broadcast from an unrelated device) arrives between fragment 1 and fragment 2, the buffer drops the array, assuming the transmission was interrupted. 

Furthermore, the list of protocol codes capable of transmitting in array fragments is currently hardcoded and incomplete.

### Technical Debt Captured from Legacy Code
* `TODO: this needs cleaning up (e.g. handle intervening packet)`
* `TODO: not a complete list [of array codes]`

### Proposed Solution
1. **Implement a Sliding Window Buffer:** Refactor `ReassemblyBuffer._process_packet` to maintain a dictionary of "pending arrays" keyed by `(src_id, code)` rather than a single `_pending_dto` variable. This allows unrelated packets to pass through the pipeline without breaking an active array reassembly.
2. **Audit Array Codes:** Investigate and expand `_ARRAY_CODES` to ensure all high-volume payload codes (like `3150` or `2309`) are correctly captured.
3. **Add Unit Tests:** Expand `test_pipeline_reassembly.py` to intentionally inject noise packets between valid array fragments to prove the sliding window functions correctly.

silverailscolo:
> Furthermore, the list of protocol codes capable of transmitting in array fragments is _currently_ hardcoded and incomplete.

@PWhite-Eng @wimpie70 is this still true, as of 0.59.7?
We are now working towards a stable, bug free release 0.60.x, so it would be useful te record our current status regarding multi-packet arrays (below).
I've labeled new features as 'hold' until 0.61.x

PWhite-Eng:
@silverailscolo To answer your question directly: **No, this is no longer an issue or blocker for `0.59.7` / `0.60.x`.**

Here is the detailed background on our current status, what has been implemented, and how the remaining live cutover will be completed in **Phase 7** of the Master Refactor Plan (#639).

---

### 1. What Has Been Implemented
[...]

PR #726, opened 2026-06-26
fix: sliding-window reassembly buffer (#669)

wimpie70:
@PWhite-Eng, @silverailscolo  My assistants were a bit bored, so I gave them some extra jobs to do. Here's another attempt to clear the list. Have a look if it suits you.
see issue #669 

Replace the single-slot _pending_dto with a dict of pending arrays keyed by (src_id, code) so that an intervening packet (RF noise or a broadcast from an unrelated device) no longer aborts an in-flight array reassembly.

Previously, any unrelated packet arriving between fragment 1 and fragment 2 of a multi-packet array (000A/22C9) caused the buffer to flush the pending fragment and drop the array.  The sliding window lets unrelated packets pass straight through while preserving the pending array for its matching peer, and supports concurrent arrays from different sources.

Also documents the _ARRAY_CODES audit: 2309/30C9/3150 are intentionally NOT added because they routinely emit single (non-array) packets that are indistinguishable from fragments at the PacketDTO level -- reliably classifying them requires the Frame._has_array length heuristic that lives in the frame layer and is part of the broader #608 migration.

Adds noise-injection tests proving the sliding window, concurrent multi-source arrays, and multi-pending timeout flush.

### DIFF
--- a/src/ramses_rf/pipeline/reassembly.py
+++ b/src/ramses_rf/pipeline/reassembly.py
@@ -10,8 +10,42 @@ from ramses_tx.dtos import PacketDTO
 
 _LOGGER = logging.getLogger(__name__)
 
-_ARRAY_CODES: Final[tuple[str, ...]] = ("000A", "22C9")
+# Codes whose ``I`` packets are *always* array fragments at the L3 layer and
+# can therefore be safely buffered for reassembly using only (code, verb)
+# inspection -- the only metadata available to this module via PacketDTO.
+#
+# Audit of ramses_rf.protocol.ramses.CODES_WITH_ARRAYS (the authoritative
+# schema, which lists 0005/0009/000A/2309/30C9/2249/22C9/3150 plus the
+# special-cased 000C/1FC9) for issue #669:
+#
+#   - 000A, 22C9: safe.  Their ``I`` packets from a controller/UFC are arrays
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):

PWhite-Eng:
Hi @wimpie70,
Thank you and congratulations on this excellent PR. The sliding window buffer logic is a brilliant and elegant solution to the RF noise interleaving problem.
I have just raised a small follow-up PR. Please note that it makes absolutely no logic changes to your fix. It simply updates the formatting of the files to align with PEP 8 and our maximum line-length limits (Issue #583), and structures the tests into the strict Arrange, Act, Assert (AAA) pattern.
As I was having my friend review the code to understand the implementation, it automatically reformatted the files according to its strict system instructions. I thought I might as well commit the reformatted files rather than waste the generated output and credits.
Please do not be offended by the follow-up tweaks; your core implementation is flawless and I am very grateful for your contribution to clearing this technical debt. Excellent work
Thanks
Phil

wimpie70:
No worries, glad to get that list shortened a bit. And good that you review/alter them.
Some more PR's  are on their way. Using GLM 5.2 atm. It's slow but cheap (free atm) and I like how it analyzes things. So far I didn't have any real problems with what it comes up with.

PR #727, opened 2026-06-26
Fix Orcon `fan_info` sensor UI jitter by preventing `31D9` state collision

PWhite-Eng:
### The Problem:
Ramses_cc [Issue 723](https://github.com/ramses-rf/ramses_cc/issues/723) reports that the Orcon HRC-400-MaxComfort `fan_info` sensor ~rapidly~ toggles between a correct speed (e.g., 30%) and an incorrect speed (e.g., 2%), _especially when poll_interval in set below 200 seconds_.

### Consequences:
What happens if this isn't fixed is severe UI jitter, noisy Home Assistant logbooks, and unreliable state history for Orcon and Brofer HVAC units.

### The Fix:
Intercepted long-format Orcon and Brofer `31D9` packets to prevent the parser from mistakenly extracting `exhaust_fan_speed` from them, deferring to the `31DA` packet as the single source of truth for fan speed.

### Technical Implementation:
Added a strict signature check (`_is_mode_only`) to `src/ramses_rf/parsers/hvac.py`. If a `31D9` packet is >= 34 characters, is padded with exactly 12 bytes of `0x20`, and ends in a non-zero byte (like `04` or `08`), the parser safely bypasses the `parse_exhaust_fan_speed` mapping.

### Testing Performed:
Created `tests/tests_rf/test_hvac.py` with a dedicated test validating that `exhaust_fan_speed` is omitted from Orcon payloads. Updated legacy logs (`code_31d9.log`) to fix a historical false-positive on the Brofer HRU payload, and passed the full pytest suite.

Users with Orcon and Brofer systems will continue to experience chaotic UI behaviour, confusing dashboards, and database bloat in Home Assistant.

### Risks of Implementing:
There is a potential risk of regression in `31D9` parsing for other device types (e.g., Itho, Nuaire, and ClimaRad) if the signature detection is too broad and accidentally drops their fan speed data.

### Mitigation Steps:
The `_is_mode_only` check strictly requires a specific packet length, a 12-byte `0x20` padding signature, and a non-`00` trailing byte. This highly specific fingerprint effectively shields Itho, Nuaire, and ClimaRad packets from being affected by the bypass logic.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/parsers/hvac.py
+++ b/src/ramses_rf/parsers/hvac.py
@@ -998,10 +998,10 @@ def parser_31d9(payload: str, msg: Message) -> dict[str, Any]:
     # Orcon and Brofer long payloads provide accurate fan speed in 31DA.
     # Emitting exhaust_fan_speed here incorrectly converts mode bytes to speed.
     # Itho formats end with "00", whereas Orcon/Brofer formats end with "04" or "08".
-    _is_mode_only = (
-        len(payload) >= 34 and payload[8:32] == "20" * 12 and payload[32:34] != "00"
+    _has_exhaust_fan_speed = (
+        len(payload) < 34 or payload[8:32] != "20" * 12 or payload[32:34] == "00"
     )
-    if not _is_mode_only:
+    if _has_exhaust_fan_speed:
         result.update(parse_exhaust_fan_speed(payload[4:6]))  # for itho, nuaire
 
     # Fan Mode Lookup 1 for Vasco codes
--- a/src/ramses_rf/parsers/hvac.py
+++ b/src/ramses_rf/parsers/hvac.py
@@ -986,7 +986,6 @@ def parser_31d9(payload: str, msg: Message) -> dict[str, Any]:
 
     # NOTE: 31D9[4:6] is fan_speed (ClimaRad minibox, Itho) *or* fan_mode (Orcon, Vasco)
     result = {
-        **parse_exhaust_fan_speed(payload[4:6]),  # for itho
         SZ_FAN_MODE: payload[4:6],  # orcon, vasco/climarad
         "passive": bool(bitmap & 0x02),
         "damper_only": bool(bitmap & 0x04),  # i.e. valve only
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
### Note on the `code_31d9.log` File Update

I want to provide a quick explanation of why a line in `tests/tests/parsers/code_31d9.log` was modified in this PR.

**1. The Integrity of the Raw Data is Untouched**
In the `.log` testing architecture, each line consists of the physical RF packet (the left side) and the expected JSON dictionary output (the right side). **The raw RF packet data was completely untouched.** It remains the absolute, undeniable truth of what was captured over the air.

**2. Correcting a Historical False-Positive**
The right side of the log file acts as a historical snapshot of the parser's output on the day the test was written. I discovered that the assertion for the **Brofer HRU** test was suffering from the exact same parser bug that we are fixing for Orcon.

Here is the raw payload for the Brofer HRU:
`000A020020202020202020202020202008`

Notice the 34-character length, the 12 bytes of `"20"` padding, and the `"08"` suffix. This is the exact same proprietary OEM signature footprint used by Orcon. 

Historically, the parser looked at the third byte (`0x02`), blindly assumed it was a speed percentage, divided it by 200, and outputted `exhaust_fan_speed: 0.01` (1%). This was a false positive; the `02` byte actually represents a Fan Mode Enum, not a 1% fan speed. 

**3. The Resolution**
Because our new `_is_mode_only` logic in `hvac.py` now correctly recognises this proprietary OEM signature and safely skips parsing that byte as a fan speed, the old test naturally failed. I updated the right side of the log file to remove the erroneous `'exhaust_fan_speed': 0.01` key. 

We have not altered reality; we have simply updated the test suite to reflect our newly corrected, mathematically sound parser behaviour.

Kars-de-Jong:
@PWhite-Eng Looks good! I would maybe invert the logic so you don't need a `not`  for the test, so perhaps rename `_is_mode_only` to `_has_exhaust_fan_speed` and change the logic accordingly.

silverailscolo:
> the 02 byte actually represents a Fan Mode Enum

Only true for Orcon, but can be used as a signature.

silverailscolo (approved):

src/ramses_rf/parsers/hvac.py
```diff
@@ -996,6 +995,15 @@ def parser_31d9(payload: str, msg: Message) -> dict[str, Any]:
         "_flags": hex_to_flag8(payload[2:4]),
     }
 
+    # Orcon and Brofer long payloads provide accurate fan speed in 31DA.
+    # Emitting exhaust_fan_speed here incorrectly converts mode bytes to speed.
+    # Itho formats end with "00", whereas Orcon/Brofer formats end with "04" or "08".
+    _is_mode_only = (
+        len(payload) >= 34 and payload[8:32] == "20" * 12 and payload[32:34] != "00"
+    )
+    if not _is_mode_only:
```
Kars-de-Jong:
```suggestion
    _has_exhaust_fan_speed = (
        len(payload) < 34 or payload[8:32] != "20" * 12 or payload[32:34] == "00"
    )
    if _has_exhaust_fan_speed:
```

Issue #649, opened 2026-05-05
10D0 `polling` task never sending, ~nor is `discovery~


silverailscolo:
To poll for FilterRemaining, I am trying to create a PollingService based on the DiscoveryService.
But my service never actually polls/sends an RQ command as defined in its task.

Might be related to the non-functional sensors reported in other issues.

Last release that had a working filter_remaining sensor poller was 0.55.2
Release 0.55.3 contained PR #466 with a discovery refactor

@PWhite-Eng Can you comment on my branch https://github.com/silverailscolo/ramses_rf/tree/eb-poll-filter-debug/src/ramses_rf

There could also be some hidden ReadOnly flag that prohibits this task.

silverailscolo:
Interesting hint about a legacy Discovery setting in the ramses_cc Wiki:
> The OTB is discoverable with the system setting Enable discovery, but do not leave this feature enabled longer than required.

This suggests that a system setting might have been lost in the latest code.
It is passed in from ramses_cc as part of the GatewayConfig object, but the default might be wrong.

wimpie70:
# Root-Cause Analysis: `polling` task never sending (#649)

## Summary

The regression introduced in 0.55.3 / PR #466 is **not** in `DiscoveryService.discover()` itself — its send loop is structurally identical to the pre-refactor `_poll_discovery_cmds`. The root cause is that **`_setup_discovery_cmds()` is never called for devices**, so `DiscoveryService.cmds` stays empty and the poller has nothing to send.

## Root Cause

### What PR #466 removed

Pre-#466, `entity_base.py` had a lazy `discovery_cmds` property that populated the command dict on first access:

```python
@property
def discovery_cmds(self) -> dict[HeaderT, dict]:
    """Return the pollable commands."""
    if self._discovery_cmds is None:
        self._discovery_cmds = {}
        self._setup_discovery_cmds()   # <-- populates cmds via add_cmd()
    return self._discovery_cmds
```

The old `_poll_discovery_cmds` loop iterated over `self.discovery_cmds` (the property), which triggered `_setup_discovery_cmds()` on first access.

### What replaced it

Post-#466, `DiscoveryService` (src/ramses_rf/discovery.py) initializes `cmds` as an empty dict in `__init__`:

```python
self.cmds: dict[HeaderT, dict[str, Any]] = {}   # line 68
```

[...]

Issue #547, opened 2026-04-01
Feature Request: Native support for 3-byte `22F1` fan rate payloads in Command.set_fan_mode


PWhite-Eng:
### The Background

Currently, the `ramses_tx` layer provides a command builder for `22F1` (Fan Rate/Mode) packets via `Command.set_fan_mode`.

When a fan mode is requested, the internal logic (using `_22F1_MODE_ORCON_MAP`) constructs a **2-byte payload**. For example, requesting a `low` fan rate generates: `002 0002`.

### The Issue

While the 2-byte payload works perfectly for devices like **Vasco** and **Orcon**, it fails silently or is ignored by other manufacturers' HVAC units.

Specifically, devices like the **Siber DF Evo 4** and **ClimaRad MiniBox** require a **3-byte payload** to process the `22F1` command (e.g., `003 000X07` where the trailing `07` is required).

Because `Command.set_fan_mode` strictly hardcodes the 2-byte payload generation, `ramses_rf` cannot currently control these devices natively.

### Current Workaround

In `ramses_cc`, we recently implemented a YAML bypass (PR #571 / Issue #210) that intercepts Home Assistant UI clicks and checks the user's `known_list` for custom `CONF_COMMANDS`. This allows users with 3-byte devices to manually configure the raw hex strings (e.g., `I 37:111111 30:123456 22F1 003 000407`).

### Proposed Solution

Users shouldn't have to write custom YAML overrides for basic fan control. The `Command.set_fan_mode` builder should be enhanced to generate the correct payload length based on the target device's manufacturer/scheme.

**Suggested Implementation Path:**
1. **Device Profiling:** Ensure the `HvacVentilator` class (or the `Command` builder itself) is aware of the device's scheme/manufacturer signature (which is often already tracked during discovery).
2. **Payload Factory:** Refactor `Command.set_fan_mode` to accept an optional `scheme` or `payload_length` parameter.
3. **Dynamic Construction:** * If `scheme == Vasco/Orcon` -> generate `002 000X`
  - If `scheme == Siber/ClimaRad` -> generate `003 000X07`

### Examples of required 3-byte payloads (from Siber DF Evo 4):
- `low`: `003 000207`
- `medium`: `003 000307`
- `high`: `003 000407`
- `boost`: `003 000707`

Resolving this technical debt will allow us to remove the YAML dependency in `ramses_cc` and provide true plug-and-play support for these ventilation units.

PWhite-Eng:
[...]

Issue #87, opened 2023-11-12
[Itho CVE-S CO2 with AUTO RFT-N] Fan states incorrectly recognised


FrankvdAa:
Hi,

Created an issue with evohome-Listener (https://github.com/smar000/evoGateway/issues/61), but problem seems to be with ramses_rf.

I have a Itho Daalderop CVE-S CO2 and humidity sensor unit with AUTO RFT-N remote controls, with on the remote controls the following options:
* Low
* High
* Auto
* Timer

![Itho-AUTO-RFT-N](https://github.com/zxdavb/ramses_rf/assets/7700220/9de51e77-87da-470f-8ce1-80f1fac787c9)

Apart from de timer option, all settings seem to be wrongly recognised.
* Low: detected as medium
* High: detected as auto
* Auto: detected as high

In below captures, I pressed Low, High, Auto. I also have an Evohome controller, that's were the setpoint message is coming from.

These captures are from evohome-Listener as the `-o packet.log` option doesn't seem to work on my ramses_rf setup.

```
2023-11-12 19:48:17 |000| HGI evoGateway        -> NUL UNBOUND           | I| puzzle_packet   | datetime: 2023-11-12T19:48:17.729, engine: v0.22.40, parser: v0.22.40
2023-11-12 19:48:24 |044| 37:198669             -> 37:019136             | I| fan_mode        | medium
2023-11-12 19:48:24,593 [301] ----> Unknown zone for src: '37:198669 (HVC) '
2023-11-12 19:48:27 |064| TRV 04:007044         -> CTL Controller        | I| setpoint        | 15.00°C                                                                     [Zone 06 ]
2023-11-12 19:48:27,834 [804] Both 'target_zone_id' and 'src_zone_id' not found in ZONES
2023-11-12 19:48:28 |045| 37:198669             -> 37:019136             | I| fan_mode        | auto
2023-11-12 19:48:28,270 [301] ----> Unknown zone for src: '37:198669 (HVC) '
2023-11-12 19:48:31 |043| 37:198669             -> 37:019136             | I| fan_mode        | high
2023-11-12 19:48:31,789 [301] ----> Unknown zone for src: '37:198669 (HVC) '
```

```
2023-11-12T19:48:17.721425 # ramses_rf 0.22.40
2023-11-12T19:48:17.777540 000  I --- 18:003618 63:262142 --:------ 7FFF 016 0010018BC4DCF84176302E32322E3430
[...]

PR #729, opened 2026-06-26
feat: scheme-aware `22F1` fan mode payloads in `Command.set_fan_mode` (#547)

wimpie70:
@PWhite-Eng can you check this one as well ?
see #547

Add a `scheme` parameter to `Command.set_fan_mode` (default "orcon" for backward compatibility) and a `mode_max` override.  The builder now uses the appropriate `_22F1_SCHEMES` map for mode name→index lookup and auto-appends the scheme's mode_max byte to produce 3-byte payloads for devices that require them (e.g. Siber DF Evo 4, ClimaRad MiniBox).

Previously the builder hardcoded `_22F1_MODE_ORCON` and always emitted a 2-byte payload, which was silently ignored by 3-byte devices.  ramses_cc had a YAML bypass HACK for this (fan_handler.py PR #571).

The `HvacVentilator.set_fan_mode` call site now passes the device's `_scheme` (from known_list traits) so the correct payload format is used automatically.

Added 10 dedicated command-builder tests covering all four schemes (orcon/itho/vasco/nuaire), int indices, None, legacy 2-byte override, and error cases.


Kars-de-Jong:
Unless there are logs from real hardware that use a 2 byte payload, I'd suggest removing support for that completely.

silverailscolo (approved):

silverailscolo:
`test_apis_hvac.py` needs a few tweaks (for a 3 byte payload)

wimpie70:
@Kars-de-Jong Yeah, no worries, I can live with that. Just trying to shorten that issue list a bit.
Maybe put this on hold ? I'll first fix those tests, or check if they only fail on < 3.14

wimpie70:
PR #729 is now CI-green. The only failing test was tests/tests/test_apis_hvac.py::test_set — it expected 2-byte 22F1 payloads but our scheme-aware builder now generates 3-byte payloads (with the orcon mode_max=07 suffix). I updated all SET_22F1_KWARGS fixtures to expect 3-byte payloads.

The 4 other test failures I saw locally (eavesdrop_schema, regression_rf) are Python 3.14-specific and don't affect CI (which uses 3.12/3.13). The dev_registry.py traits_dict UnboundLocalError fix I found is a real bug but belongs in the PR that should fix issue #649, not here — I reverted it from this branch.

wimpie70:
i have a few PR's on the way:
- PR #732 complements the fix for issue #649 handling state ingestion (dispatcher wiring 31DA/22F1/31D9 parsed data into device state). Without it, FAN sensors show no values on hass. This is the fix you just installed.
- Fix for #87 is about parser correctness (22F1 scheme detection for directed Itho packets)
- #729 (this PR) is about command building (set_fan_mode generating correct 3-byte payloads)

silverailscolo:
> i have a few PR's on the way:

Thanks for the heads up. They can all go into 0.57.3

PWhite-Eng:
Hi @wimpie70,

Thanks for the work on this PR. When reviewing your PR  I ended up tweaking a couple of things to align the API with industry best practice—specifically, replacing the empty string trigger (`mode_max=""`) with an explicit `legacy_format` boolean flag to make the API contract a bit clearer.

While updating the isolated tests to match, it exposed a rather nasty L3/L7 circular import bug deep in the `ramses_tx` initialisation sequence. To get the tests to run green in isolation, I had to bundle in an architectural fix using PEP 562 lazy loading.

I’ve rolled your core logic, the test refactors, and the architectural fix into a new PR here: #738. Thank you for identifying the solution subvmitting the orignal PR.

PR #734, opened 2026-06-26
fix: 22F1 parser scheme detection for directed Itho packets (#87)

wimpie70:
The 22F1 parser only identified mode_max=04 as itho when the source was the broadcast address  (NON_DEV_ADDR).

Directed 22F1 packets from Itho remotes (e.g. CVE-S CO2 with AUTO RFT-N) were misidentified as orcon, causing fan_mode names to be wrong:
  02 → "medium" (orcon) instead of "low"    (itho)
  04 → "auto"   (orcon) instead of "high"   (itho)
  03 → "high"   (orcon) instead of "medium" (itho)

## Fix
Treat mode_max=04 as itho regardless of addressing, scoped to standalone 22F1 packets (msg.code == 22F1) to avoid affecting 22F3 boost payloads which embed 22F1-style mode info but may use different mode_max conventions.

Includes fix for issue #87 capture fixtures to code_22f1.log and 4 dedicated parser unit tests covering directed itho, broadcast itho (no regression), directed orcon (no regression), and directed vasco (no regression).


silverailscolo (approved):

wimpie70:
k, you get 1 more on _cc: ramses_cc timing race — SIGNAL_UPDATE fires before async CQRS ingestion completes, so entities read stale hvac_state.

PR #736, opened 2026-06-27
fix: align 10D0 field names between parser and CQRS dispatcher

wimpie70:
The 10D0 parser returns `days_remaining` and `percent_remaining` (from SZ_REMAINING_DAYS / SZ_REMAINING_PERCENT), but the CQRS ingestion engine in `_update_hvac_state` was checking for `remaining_days` and `remaining_percent`.  The field names didn't match, so filter_remaining data from 10D0 packets was silently dropped and never ingested into `hvac_state`.

Fix: update the dispatcher to check for the correct field names (`days_remaining`, `percent_remaining`) that the parser actually produces.

Fix 2: HgiGateway.is_active fix (accessing _protocol._this_msg + timestamp instead of _this_msg + dtm). 

This resolves the `filter_remaining_percent` not updating issue reported on PR #730.

https://github.com/ramses-rf/ramses_cc/pull/744 depends on this

@PWhite-Eng

### DIFF
--- a/src/ramses_rf/devices/dev_base.py
+++ b/src/ramses_rf/devices/dev_base.py
@@ -45,6 +45,7 @@ if TYPE_CHECKING:
     from ramses_rf.models import DeviceTraits
     from ramses_rf.systems import Evohome, Zone
     from ramses_tx.const import IndexT
+    from ramses_tx.dtos import PacketDTO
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
@wimpie70, nice little fix, thank you
Phil

silverailscolo (approved):

PR #738, opened 2026-06-27
feat: native 3-byte `22F1` payloads, test refactoring, and L3/L7 circular import resolution

PWhite-Eng:
### The Problem:
This work originates from Issue #547. The `Command.set_fan_mode` builder hardcoded a 2-byte payload, which is silently ignored by 3-byte devices like the Siber DF Evo 4 and ClimaRad MiniBox. While attempting to write tests for a fix, running `pytest tests/tests_tx/test_command.py` in isolation triggered a fatal `ImportError`. This exposed a severe, hidden L3/L7 circular dependency where the transport layer (`ramses_tx`) was eagerly requesting protocol schemas, forcing the application layer (`ramses_rf`) to initialise out of order.

### Consequences:
Without the payload fix, users with 3-byte devices must rely on brittle YAML bypass hacks to control their fans. Without the architectural fix, isolated test execution fails, and the Python module loader remains trapped in a fragile circular dependency that is only masked by full-suite module caching.

### The Fix:
Added scheme-aware payload generation for `22F1` commands using a dictionary mapping, introduced an explicit boolean flag for legacy support, rewrote the test suite to modern standards, and completely severed the L3/L7 runtime circular dependency using PEP 562 lazy module loading.

### Technical Implementation:
*   **HVAC Payloads:** Implemented `_22F1_MODE_MAX` to map scheme names to their required trailing bytes. Replaced the "magic string" approach (`mode_max=""`) with an explicit `legacy_format: bool = False` parameter in `set_fan_mode`.
*   **Test Refactoring:** Upgraded `test_command.py` to strictly follow the Arrange, Act, Assert (AAA) pattern, utilising robust `pytest.raises` context managers.
*   **Architectural Resolution:** Refactored `src/ramses_rf/__init__.py` to use PEP 562 (`__getattr__`) dynamic attribute loading. Heavy domain classes and re-exported L3 primitives are now isolated behind runtime lookups, instantly breaking the eager circular loop.
*   **Static Analysis:** Maintained explicit `if TYPE_CHECKING:` blocks and `__all__` exports to guarantee zero Mypy strict errors, seamless IDE auto-completion, and Ruff compliance.

### Testing Performed:
*   Executed the full `pytest` suite (all tests passing).
*   Executed `pytest tests/tests_tx/test_command.py` in strict isolation to verify the circular import resolution.
*   Ran `mypy --strict` to validate cross-layer typing, ensuring the PEP 562 implementation introduced no regressions.

Technical debt continues to accumulate. Users remain reliant on YAML workarounds, and future developers will continuously hit `ImportError` deadlocks when running isolated tests.

### Risks of Implementing:
If downstream scripts bypass the public API and rely on deep, undocumented imports from the `__init__.py` gateways, the lazy loading might theoretically alter their initialisation order. 

### Mitigation Steps:
The `__getattr__` implementation was carefully designed to perfectly mimic the existing public API. Mypy strict mode confirms that type hinting, static analysis, and runtime contracts remain 100% identical for external consumers.

### AI Assistance Disclosure:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -337,3 +337,62 @@
 
 [tool.ruff.lint.per-file-ignores]
   "tests/*" = ["ASYNC"]
+
+[tool.importlinter]
+root_packages = [
+    "ramses_rf",
+    "ramses_tx",
+    "ramses_cli"
+]
+ignore_type_checking_imports = true
+
+# 1. Macro Level: CLI -> App -> Transport (OSI L1-L3 vs L4-L7)
+[[tool.importlinter.contracts]]
+name = "Macro Package Separation"
+type = "layers"
+layers = [
+    "ramses_cli",
+    "ramses_rf",
+    "ramses_tx",
+]
+
+# 2. ramses_rf: The Event-Driven CQRS Hierarchy
+[[tool.importlinter.contracts]]
+name = "ramses_rf CQRS Architecture Layers"
+type = "layers"
+layers = [
+    "ramses_rf.gateway",      # L7 Orchestrator / Facade (Top)
+    "ramses_rf.pipeline",     # Async Event Bus (Reassembly, Decoder, Dispatcher, TopologyBuilder)
+    "ramses_rf.systems",      # CQRS Read-Models (Systems / Zones)
+    "ramses_rf.devices",       # CQRS Read-Models & Registry
+    "ramses_rf.state",        # SSOT Database / Caches
+    "ramses_rf.schemas",      # Pure DTO Validation
+    "ramses_rf.messages",     # Immutable Event Envelopes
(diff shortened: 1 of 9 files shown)


PWhite-Eng:
Hi @silverailscolo,

Just a quick note regarding the `pyproject.toml` file in this PR. As part of resolving the deep L3/L7 circular import deadlock we hit during testing, I have added the configuration block for `import-linter`.

I have left it in the codebase intentionally. If you were to run `lint-imports` right now, it will highlight a number of broken contracts where L3 reaches up into L4+. However, I thought it would be an incredibly useful tool to leave in place as a roadmap for the wider Phase 5 architectural refactoring. We can systematically clean those broken contracts up as the refactor progresses

silverailscolo (approved):

PR #749, opened 2026-06-29
fix(hvac): 31DA/12A0 null-marker quirks and parse_fan_info  handling

### LINKED ISSUE #742: 0.57.3: HVAC Ventura sensors bounce to None/FF
Since installing the ramses_rf 0.57.3 lib, some sensors on my Ventura revert to `None`/ 0 /`FF` every 10 minutes.
Applies to:
- indoor_humidity (0)
- fan_mode (FF)
- bypass_position (None)

Indoor_humidity (RH, RV):
image

Detail:
image

image

See also #723 and older #653 (but those jumped to an initial value).
Changed in ramses_rf [PR 727](https://github.com/ramses-rf/ramses_rf/pull/727).

Specifically for the Ventura, we used to ignore the FF value for fan_mode.
Fan_rate (speed step) and all temps are reporting correctly btw.
Increasing the gateway config polling interval from 150 to 200 makes no difference.

Also, no fan_mode is reported for ClimaRad MiniBox fan since 0.57.3 (Unknown).
No problems on Vasco D60 HRU sensors or fan_info.

wimpie70:
fix(hvac): 31DA/12A0 null-marker quirks and parse_fan_info graceful handling

## Summary

Four bugs fixed for HVAC devices (Ventura V1x, Orcon, Nuaire) where 31DA snapshot null markers overwrite good telemetry from 22F1/22F4/22F7/12A0 every polling cycle (~10 min), causing sensors to bounce to `None`/`0`/`-unknown-`.

This is a follow-up to #737 (which fixed the initial null-marker bounce for `None`/`"FF"`/`0.0` humidity values in the dispatcher) and addresses the remaining null markers that slipped through that filter.

Fixes: [silverailscolo/ramses_cc#742](https://github.com/ramses-rf/ramses_cc/issues/742)
Related:[ silverailscolo/ramses_cc#752](https://github.com/ramses-rf/ramses_cc/issues/752)  (Strange behaviour since HA 2026.6.4 — some symptoms traced to these null markers)
Related: #737 (initial null-marker bounce fix)
Related: #463 (Ventura 12A0 parser fix)

### Bug 1: 12A0 idx=01 key mismatch (`quirks.py`)

`parse_humidity_element` returns `rel_humidity` for idx=01, but the quirks code checked for `indoor_humidity` — the remapping never fired, so supply humidity was silently lost.

**Fix:** Explicitly drop `rel_humidity` (no `supply_humidity` field in `HvacState`) and add idx=02 handling (`temperature` → `outdoor_temp`).

**Impact:** Supply humidity from 12A0 idx=01 was never reaching `HvacState`. Supply temperature was also not being remapped to `supply_temp`.

### Bug 2: 31DA `bypass_position=0.0` null marker (`quirks.py`)

Ventura V1x sends `0x00` for bypass_position in 31DA (it reports bypass via 22F7). The value `0.0` passed through the dispatcher's `val is None` filter from #737, overwriting the real bypass position from 22F7 every polling cycle.

**Fix:** Preserve the existing non-zero `bypass_position` when 31DA sends `0.0`.

**Impact:** `bypass_position` sensor bounced to 0 every ~10 min for devices that report bypass via 22F7.

### Bug 3: 31DA `fan_info='-unknown 0x1F-'` (`quirks.py`)

Ventura V1x sends `0x1F` for fan_info in 31DA (it reports fan state via 22F1/22F4). The `-unknown 0x1F-` string passed through the existing `("off", "")` filter from #737, overwriting the valid fan_info from 22F1.

**Fix:** Also catch strings starting with `-unknown` and preserve the existing valid fan_info.

**Impact:** `fan_info` sensor bounced to `-unknown 0x1F-` every ~10 min for devices that report fan state via 22F1/22F4.

### Bug 4: `parse_fan_info` crash on `0xFF`/`0xEF` (`helpers.py`)

The parser had a commented-out `EF` check and an `assert` that crashed with `AssertionError` on any unknown high-nibble pattern (`0xFF`, `0xEF`, `0x1F`, etc.). This was found by parsing all 31DA payloads from the `ramses_extras` device_db — 6 payloads from `fan_co2_dcv.yaml` (Nuaire) crashed the parser.

**Fix:**
- `EF`/`FF` return `{fan_info: None}` (not implemented / no data)
- Unknown codes (e.g. `0x1F`) return `-unknown 0xNN-` instead of crashing
- Original `# TODO: Not implemented???` marker preserved

**TypedDict:** `FanInfo.fan_info` type widened to `str | None` to match the `EF`/`FF` return.

### Verification against device_db

All 23 valid 31DA/12A0 payloads from the `ramses_extras` device_db conversations were parsed through the full pipeline (parser → quirks → dispatcher). Before the fix, 7 issues were found (6 parser crashes + 1 null-marker overwrite). After the fix, all 23 payloads process correctly.

### Live system verification

Verified on a live Orcon system (32:153289) after restart:
- `sensor.32_153289_fan_info`: `auto` (preserved from 22F1, not overwritten by 31DA)
- `sensor.32_153289_indoor_humidity`: `64.0` (not bouncing to 0.0)
- `sensor.32_153289_exhaust_fan_speed`: `15.0` (not bouncing to 0.0)
- `sensor.32_153289_indoor_temp`: `24.31` (updating correctly)

### Files changed

| File | Changes |
|------|---------|
| `src/ramses_rf/quirks.py` | 12A0 idx=01/02 fixes, 31DA fan_info/bypass_position null-marker quirks |
| `src/ramses_tx/helpers.py` | `parse_fan_info` EF/FF/unknown graceful handling (was `AssertionError`) |
| `src/ramses_tx/typing.py` | `FanInfo.fan_info` type → `str \| None` |
| `tests/tests_rf/test_quirks.py` | 45 new tests (new file) |

#### Test plan

- [x] 45 new tests in `test_quirks.py` — all pass
- [x] 21 existing dispatcher tests — all pass
- [x] Full test suite (33,358 tests) — all pass (1 pre-existing regression failure unrelated)
- [x] ruff check + ruff format — clean
- [x] mypy — clean
- [x] device_db payload analysis (23 payloads) — all process correctly
- [x] Live system verification — sensors stable, no bouncing

### DIFF
--- a/src/ramses_tx/helpers.py
+++ b/src/ramses_tx/helpers.py
@@ -780,16 +780,23 @@ def parse_fan_info(value: HexStr2) -> PayDictT.FAN_INFO:
     if not isinstance(value, str) or len(value) != 2:
         raise ValueError(f"Invalid value: {value}, is not a 2-char hex string")
 
-    # if value == "EF":  # TODO: Not implemented???
-    #     return {SZ_FAN_INFO: None}
+    # TODO: Not implemented???  # EF, FF = no data / not implemented
+    if value in ("EF", "FF"):
+        return {
+            SZ_FAN_INFO: None,
+            "_unknown_fan_info_flags": [0, 0, 0],
+        }
 
-    assert int(value, 16) & 0xE0 in (
-        0x00,
-        0x20,
-        0x40,
-        0x60,
-        0x80,
-    ), f"invalid fan_info: {int(value, 16) & 0xE0}"
+    if int(value, 16) & 0xE0 not in (0x00, 0x20, 0x40, 0x60, 0x80):
+        # Unknown fan_info code (e.g. Ventura 0x1F) — return as unknown
+        # instead of crashing with AssertionError.  The quirks layer
+        # will prevent it from overwriting a valid fan_info from 22F1/22F4.
+        return {
+            SZ_FAN_INFO: f"-unknown 0x{value}-",
+            "_unknown_fan_info_flags": [
+                (int(value, 16) >> x) & 1 for x in range(7, 4, -1)
+            ],
+        }
 
     flags = list((int(value, 16) & (1 << x)) >> x for x in range(7, 4, -1))
 
--- a/src/ramses_rf/pipeline/ingestion.py
+++ b/src/ramses_rf/pipeline/ingestion.py
@@ -451,9 +451,27 @@ class StateProjector:
             SZ_DEWPOINT_TEMP,
         ]
 
+        # Filter out null-marker values that 31DA/31D9 snapshots emit for
+        # sensors the device does not have.  Without this, every polling cycle
+        # (~10 min) overwrites good telemetry from 22F1/12A0/22F7 with null
+        # markers, causing sensors to bounce to None/FF/0.  See issue #742.
+        # This must mirror the filtering in dispatcher._update_hvac_state.
+        _NULL_HUMIDITY_FIELDS = frozenset({SZ_INDOOR_HUMIDITY, SZ_OUTDOOR_HUMIDITY})
+
         for f in fields:
-            if f in p:
-                updates[f] = p[f]
+            if f not in p:
+                continue
+            val = p[f]
+            # None = "not implemented" (e.g. EF in bypass_position)
+            if val is None:
+                continue
+            # "FF" = "no data" marker from 31D9 raw hex fan_mode
+            if f == SZ_FAN_MODE and val == "FF":
+                continue
+            # 0.0 for humidity = "no sensor" (00 parses as 0%, physically impossible)
+            if f in _NULL_HUMIDITY_FIELDS and val == 0:
+                continue
+            updates[f] = val
 
         # Handle non-standard names passed by the semantic parsers
         if SZ_REMAINING_DAYS in p:
--- a/src/ramses_rf/quirks.py
+++ b/src/ramses_rf/quirks.py
@@ -29,7 +29,13 @@ from typing import Any, Final
 
 from ramses_rf.const import DevType
 from ramses_rf.models import HvacState
-from ramses_tx.const import MsgId
+from ramses_tx.const import (
+    SZ_FAN_MODE,
+    SZ_INDOOR_HUMIDITY,
+    SZ_OUTDOOR_HUMIDITY,
+    SZ_REL_HUMIDITY,
+    MsgId,
+)
 
 # Map of device types to sets of OpenTherm MsgIds that are known to be unreliable
 QUARANTINED_OT_MSG_IDS: Final[dict[str, set[MsgId]]] = {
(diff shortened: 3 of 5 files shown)


src/ramses_tx/helpers.py
```diff
@@ -780,16 +780,23 @@ def parse_fan_info(value: HexStr2) -> PayDictT.FAN_INFO:
...
-
-    assert int(value, 16) & 0xE0 in (
-        0x00,
-        0x20,
-        0x40,
-        0x60,
-        0x80,
-    ), f"invalid fan_info: {int(value, 16) & 0xE0}"
+    # TODO: Not implemented???  # EF, FF = no data / not implemented
+    if value in ("EF", "FF"):
```
silverailscolo, line 784: thanks. Can you add a similar filter on 31D9 fan_mode=FF to prevent overwriting the valid Ventura fan_mode sent via 22F4 (see _cc issue).

silverailscolo:
Thanks @wimpie70 
Testing this branch on a Ventura setup, but the Indoor_humidity jumping to 0.0 is still there:

Scherm­afbeelding 2026-06-29 om 16 16 46
The period without the 0-drops is when I ran with my own dev branch. The last 2 0-drops is with this PR.

I had a try replacing a 0 value by EF in the parser. That works, but it's cheating, of course.
We might say that in the quirks, a value of 0.0 for relative humidity is equivalent to None (at least on Earth). That would then not be forwarded by Ingestion.

The Ventura also has a `fan_mode` error. That key is in 31D9 with an unused value of FF and in 22F4 (valid). I tried to substitute FF for EF, but `fan_mode` (lookup str) has no filter yet in the hvac_quirks. I would like to catch "FF" there, so that only the `fan_mode` value in 22F4 passes to ramses_cc (has its own lookup).

wimpie70:
This was not easy to find :D

**There are two active code paths that update `hvac_state`:**

1. `dispatcher.py` `_cqrs_ingestion_engine` → has null-marker filtering (from #737) ✓
2. `gateway.py:322` `state_projector.process_message_state` → `pipeline/ingestion.py` → **no null-marker filtering** ✗

The second path runs *after* the dispatcher and blindly copies all parsed fields into state, undoing the filtering that #737 added. That's why the 0.0 bounce persisted — #737 only patched one of the two paths.

**Three additional fixes in this push:**

1. **`pipeline/ingestion.py`** — Added the same null-marker filtering as the dispatcher (skip `None`, skip `"FF"` fan_mode, skip `0.0` humidity). This is the root cause fix.

2. **`quirks.py` — 31DA humidity 0.0 → None** — As you suggested: in the quirks, a value of 0.0 for relative humidity is equivalent to None (at least on Earth). Normalised to `None` so both ingestion paths filter it out. Belt-and-suspenders alongside fix # 1. [leave out the space to point to our PR # 1 ]

3. **`quirks.py` — 31D9 fan_mode "FF" → None** — Same approach for the fan_mode issue you reported. 31D9 sends raw hex `"FF"` when no data is available; normalised to `None` in quirks so only the valid fan_mode from 22F4 reaches ramses_cc.

All 33,374 tests pass, ruff/mypy clean. Could you re-test on your Ventura?

wimpie70:
# HVAC Packet Test Results

Did a comprehensive test with data from the device simulator. Not the same as real units, but...looks good.

## Test Results

| Source | Files | Packets | Errors | Bounces |
|---|---|---|---|---|
| ramses_extras conversations (35 devices) | 87 | 51 | 7 | 0 |
| ramses_extras device types (FAN/CO2/REM) | 5 | 49 | 1 | 0 |
| ramses_rf parser logs (all brands) | 15 | 577 | 69 | 0 |
| **Total** | **107** | **677** | **77** | **0** |

**Zero bouncing detected across all 677 HVAC packets from all known fan brands and types.**

## Devices/brands tested

- **Orcon** (15RF Display, 15RPS34, 15CM17, 23C33, 23HB33)
- **Itho** (HRA82, VMD 15RMS64/86, 17RPS01, 07RPS13, 15WSJ53)
- **Nuaire** (co2_fan_nuaire, rem_fan_nuaire)
- **Ventura** (rem_fan_ventura)
- **Climarad** (rem_fan_climarad)
- **Vasco** (rem_fan_vasco)
- **Nuaire CVE-RF** (99_CVE_RF)
- **Jasper/CCU** (21_CCU_12T20, BRDG_02JAS01)
- **Various bridges** (BRDG_02A55, BRDG_02EM23)
- **VMC/VMN/VMS/VMZ** variants (07RP01, 15RP01, 17RP01, 07LM01, 15LF01, etc.)
- **Schema logs** (schema_000 through schema_310)
- **Regression packets** (577 packets from fixtures)

## Errors (all pre-existing, not caused by our changes)

- `payload=00` (1-byte payloads) — invalid test data, not real packets
- `22F3` MagicMock comparison — test harness artifact (the parser expects a real `Message` object, not a MagicMock)
- `31DA payload=21/6C` — truncated payloads in test logs

No surprises for the release. The null-marker filtering works correctly across all fan brands without breaking any valid data.

## Error breakdown — all test harness artifacts

All 77 "errors" from the earlier run were **test harness artifacts**, not parser bugs:

| Error type | Count | Cause | Real bug? |
|---|---|---|---|
| `payload=00/21/6C` (1-byte) | 55 | RQ packets not skipped (regex captured `Q` instead of `RQ`) | No — test harness |
| `22F3` MagicMock comparison | 11 | `msg.len` was MagicMock instead of int | No — test harness |
| `10D0 payload=00FF` | 2 | `msg.verb` was `'W'` instead of `VerbT.W_` (`' W'`) | No — test harness |
| `22F1/22F4/22F7 payload=00` | 9 | RQ packets not skipped (same regex issue) | No — test harness |

**No parser bugs found. Nothing needs fixing in another PR.** The existing parsers handle all real-world payloads correctly — they just need proper `Message` objects with correct `verb` and `len` attributes, which the real pipeline provides.

silverailscolo (approved):
Looks good here running the (ClimaRad) VenturaV1X. Will be back in an hour to merge, thanks.

src/ramses_tx/helpers.py
wimpie70: Yes, already done — in three places:

quirks.py L116-122: Normalises 31D9 fan_mode="FF" → None at the quirk layer, so the valid fan_mode from 22F4 is preserved
dispatcher.py L533: Filters fan_mode == "FF" in the _update_hvac_state ingestion loop
pipeline/ingestion.py L468: Same filter in the StateProjector path (the one that was missing before, causing the bounce)
The quirk normalises FF → None before either ingestion path sees it, and both paths also have an explicit if f == SZ_FAN_MODE and val == "FF": continue filter as belt-and-suspenders. We tested this specifically with the Ventura cold-start scenario (31D9 FF → 22F4 auto → 31D9 FF) and confirmed fan_mode stays auto.

PR #755, opened 2026-06-30
fix: restore HGI ID patching in PortProtocol.send_cmd

wimpie70:
PortProtocol.send_cmd overrides _BaseProtocol.send_cmd and skips the _patch_cmd_if_needed() call, so all commands go out with the 18:000730 placeholder instead of the real HGI ID. This affects all transport types (native MQTT, HA MQTT bridge, serial/USB).

Confirmed on two independent systems: TX commands carry 18:000730 while RX shows the real ESP ID. The ESP rewrites the ID for RF, so devices still respond, but the MQTT frame carries the wrong placeholder.

[See ramses_cc issue 757.
](https://github.com/ramses-rf/ramses_cc/issues/757)

### DIFF
--- a/src/ramses_tx/protocol/core.py
+++ b/src/ramses_tx/protocol/core.py
@@ -302,6 +302,13 @@ class PortProtocol(_DeviceIdFilterMixin):
             _LOGGER.warning(f"{cmd} < num_repeats set to 0, as wait_for_reply is True")
             num_repeats = 0
 
+        # Patch command with actual HGI ID if it uses the default placeholder.
+        # PortProtocol.send_cmd overrides _BaseProtocol.send_cmd (which does this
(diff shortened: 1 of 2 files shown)


wimpie70:
@PWhite-Eng I know you've been busy with this part. Just want to point you to this change and hope you agree.

wimpie70:
Fixed. The test was checking cmd_str[7:16] (raw string) which shows --:------, but the Command parser normalizes --:------ src to 18:000730 when addr0 is 18:000730. 
So _patch_cmd_if_needed correctly patches it for evofw3. 
The test now also checks cmd.src.id == HGI_DEVICE_ID for evofw3 to account for this normalization.

silverailscolo (approved):

PR #756, opened 2026-06-30
fix: drop raw-hex fan_mode from 31D9 to prevent toggle with 22F4

wimpie70:
31D9 long-payload devices (Orcon, Brofer) emit raw hex bytes for fan_mode (e.g. "04", "C8", "FF") which conflict with the semantic fan_mode from 22F4 ("auto", "off", "paused", "manual") or 22F1 (scheme-specific names). Every 31D9 broadcast cycle overwrites the good semantic value, causing fan_mode to toggle (e.g. "auto" ↔ "04" every ~30 minutes).

The previous fix only normalised "FF" → None.  This extends the quirk to drop any 2-char hex string (raw byte) from 31D9 fan_mode.  Semantic strings from the Vasco/ClimaRad short-payload lookup (e.g. "auto", "off", "4 (boost)") are preserved — they don't match the hex pattern.

The dispatcher and pipeline/ingestion.py filters are updated to match, providing belt-and-suspenders protection across both ingestion paths.

The authoritative semantic fan_mode comes from:
- 22F4 (polled RP, every ~30 min)
- 22F1 (command reply, when user sets fan mode)

See [ramses_cc issue 723.](https://github.com/ramses-rf/ramses_cc/issues/723)

### DIFF
--- a/src/ramses_rf/dispatcher.py
+++ b/src/ramses_rf/dispatcher.py
@@ -530,9 +530,15 @@ def _update_hvac_state(target: Any, p: dict[str, Any], msg: Message) -> None:
         # None = "not implemented" (e.g. EF in bypass_position)
         if val is None:
             continue
-        # "FF" = "no data" marker from 31D9 raw hex fan_mode
-        if f == SZ_FAN_MODE and val == "FF":
-            continue
+        # Raw hex (e.g. "FF", "04") = non-semantic fan_mode from 31D9
+        # long-payload devices; the quirk normalises these to None, but
+        # filter here as belt-and-suspenders.  See ramses_cc issue 723.
+        if f == SZ_FAN_MODE and isinstance(val, str) and len(val) == 2:
+            try:
+                int(val, 16)
+                continue
+            except ValueError:
+                pass
         # 0.0 for humidity = "no sensor" (00 parses as 0%, physically impossible)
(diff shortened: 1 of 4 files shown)


wimpie70:
I've tested on my own orcon dev and ran it with msgs from the device simulator dbase.

## Live HA Test (Orcon 32:153289)

31D9 raw packets still show `fan_mode=02` in the dispatcher log (that's the
raw parsed value before quirks — expected).

Climate entity state shows **no raw hex values**:

| fan_mode value | Source |
|----------------|--------|
| `speed 1, low` | 31DA fan_info |
| `speed 3, high` | 31DA fan_info |
| `auto` | 22F4 |
| `None` | startup |

No `"02"`, `"04"`, or `"05"` appearing in the entity state. The quirk
normalises the raw hex to `None` before it reaches the `hvac_state`, and the
dispatcher/ingestion filters drop it.

## Summary

| Test | Result |
|------|--------|
| Live HA (Orcon 32:153289) | 31D9 sends raw hex `02`/`04`/`05` → all filtered. Entity state shows only semantic values (`auto`, `speed X`). Zero raw hex in last 200 state changes. |
| Device simulator dbase | 11 of 12 31D9 packets across all conversations → all raw hex dropped correctly. No issues. |

silverailscolo (approved):
Thanks. Great to include the explanation in the code

PR #758, opened 2026-07-01
feat: add discovery commands for Programmer

wimpie70:
Phase 2 of (https://github.com/ramses-rf/ramses_cc/pull/759) fill discovery command gaps for device types that can respond to RQ but have no _setup_discovery_cmds() override.

Programmer (23:): add RQ 1090, 10A0 (every 6h) — PRGs have RP for these codes in the protocol schema.

HvacRemote (REM): add RQ 10D0 (every 24h) — REMs have RP for 10D0 (filter change status) in the protocol schema.

Note: TRV (04:) and DHW (07:) as far as I can see these are passive-only by protocol design (zero RP codes) and cannot be actively probed. The discover_known_devices service in ramses_cc creates them from known_list/schema, but they only get state data when they broadcast traffic on their own.

https://github.com/ramses-rf/ramses_cc/issues/627

### DIFF
--- a/src/ramses_rf/devices/heat_controllers.py
+++ b/src/ramses_rf/devices/heat_controllers.py
@@ -115,6 +115,20 @@ class Programmer(Controller):  # PRG (23):
 
     _SLUG = DevType.PRG
 
+    def _setup_discovery_cmds(self) -> None:
+        super()._setup_discovery_cmds()
+
+        if not self.is_faked:
+            # PRGs respond to RP for 1090, 10A0, 3EF1
+            self.discovery.add_cmd(
+                Command.from_attrs(RQ, self.id, Code._1090, PayloadT("00")),
+                60 * 60 * 6,  # every 6 hours
+            )
+            self.discovery.add_cmd(
+                Command.from_attrs(RQ, self.id, Code._10A0, PayloadT("00")),
+                60 * 60 * 6,
+            )
+
 
 class RfgGateway(DeviceHeat):  # RFG (30:)
     """The RFG100 base class."""


wimpie70:
i let my assistant create a report on the tests and findings:

## Context

This PR addresses Phase 2 of ramses_cc [issue 627](https://github.com/ramses-rf/ramses_cc/issues/627) (Finding devices on a clean system).
The companion ramses_cc [PR 759](https://github.com/ramses-rf/ramses_cc/pull/759) adds a `discover_known_devices` service that creates devices from `known_list`/schema so their HA entities appear immediately, even without cached messages. This PR fills the gaps where device types *can* respond to RQs but have no `_setup_discovery_cmds()`
override, so they were never actively probed.

## What this PR does

Adds `_setup_discovery_cmds()` to the **Programmer** class (23:) — registers
RQ 1090 and RQ 10A0, both every 6 hours. PRGs have RP for these codes in the protocol schema.

## What was investigated but not included

### REM (29:) — tested on real hardware, not viable

Tested RQ 10D0 and RQ 0001 on a real REM 329:176861) — no
response. REMs are send-only controllers (they send 22F1/22F3 to
the FAN and flash a light on the FAN's response), not queryable devices.
The protocol schema lists RP codes for REM, but in practice basic REMs
don't implement them. REMs are passive-only in practice, like TRVs and DHW sensors.

### TRV (04:) and DHW (07:) — passive by protocol 

(tested on device simulator)
Checked the protocol schema (`protocol/ramses.py`): TRV and DHW have zero
RP codes. They only broadcast I packets (1060 battery, 3150 heat demand,
etc.) and send RQs to the CTL. They never respond to RQs directed at them.
This is not a missing implementation — it's how the RAMSES protocol works
for battery-powered devices (they sleep to conserve power and only wake to broadcast).

### THM — untested

THM has RP for 3220 only, but this may only work for RNDs (Spider thermostats).
Not tested on real hardware (no device available). Left as a possible future
addition if someone with a Spider can confirm.

### DeviceInfo 10E0 poll — mostly useless (future cleanup)

The `DeviceInfo` base class adds RQ 10E0 for all devices, but only CTL, _FAN and CO2 (powered)_
actually RP to RQ 10E0. All other device types (TRV, DHW, THM, ~FAN, CO2,~
REM) only broadcast I 10E0 — they don't respond to RQ 10E0. That discovery
command sends unnecessary RQs that never get responses. Consider making it
conditional on the device type actually having RP for 10E0 in a future PR.
> Edit: see Task #759

## What this means for ramses_cc [issue 627](https://github.com/ramses-rf/ramses_cc/issues/627)

The `discover_known_devices` service (ramses_cc PR 759) is the correct and
only approach for passive devices (TRV, DHW, REM, OUT). It creates them from
known_list/schema so their entities appear immediately. State data arrives
when the devices broadcast on their own — there is no protocol-level way to
probe them.

For the Programmer, this PR adds active discovery commands so it can be
detected and its state queried proactively, complementing the
`discover_known_devices` service.

silverailscolo (approved):

PR #762, opened 2026-07-02
Remove legacy getter for zone name

silverailscolo:
This PR is supposed to fix ramses_cc issue [770 Zone Name loop](https://github.com/ramses-rf/ramses_cc/issues/770).


silverailscolo:
Simple removal does not pass tests, but current code crashes the zone name getter, so this needs fixing by updating the getter to modern entity_state form `await self.entity_state.get_value()`

silverailscolo:
@PWhite-Eng Could you confirm this is where to future code should find the Zone _name attribute?

It too often returns None, but the getter that was here sure was historical 'dust' and crashed on evohome. Not certain _name is in entity_state (it's not really a state, more of an attribute). Thanks

PWhite-Eng:
@silverailscolo, just to let you know, I've submitted PR #768 which fixes the historical `0004` lookup issue and re-enables the `test_gateway_replay_regression` test that was disabled here.

The root cause was that `EntityState` flattens packet logs, which dropped earlier zone name packets during rapid replay. To fix this and align with our Phase 2.95 refactor, I've completely bypassed `EntityState` and migrated `Zone.name()` to read natively from the new CQRS `ZoneState` models. I also restored the pristine snapshot file from commit `321a5fc0` to guarantee we have no lingering state deviations.

PR #767, opened 2026-07-03
Turn off rf_regression test

silverailscolo:
This PR fixes an issue where a small failure in this test snapshot would completely overflow the gh Testing results, as well as reducing by 50% the total testing run time on gh, resulting in faster developer turnover.

- [x] AI Coding Assistance disclosure: not used


silverailscolo:
@PWhite-Eng To inform you these 2 massive tests are currently inactive. Feel free to turn them on again if the help to protect the codebase during the refactor.

I got some hints from gh Copilot that the snapshots are so detailed that it fails on the slightest of change (suggestion: just log the error type). When we repair the Zone.name() function this could pass again.

On local pytest, 3 other ramses_rf tests keep failing on master,; expect a timing issue somewhere. Also have restarted gh Testing 3 times after a 'lingering task'. Ran fine next run, could point to a similar cause.

PR #768, opened 2026-07-04
Fix: Implement CQRS Event-Sourced Hydration for Zone Names & Restore Regression Suite

PWhite-Eng:
### The Problem:

PR #762 introduced a regression in historical gateway replays by cutting the `Zone.name` property over to the legacy `EntityState` class. Because `EntityState` flattens the message log by `Code`, it actively overwrote earlier `0004` packets, causing retroactive name lookups to fail. To bypass this failure, the critical gateway state regression test (`test_gateway_replay_regression`) was disabled, and its snapshot file was deleted.

### Consequences:

The `main` branch was left entirely blind to gateway state regressions. Furthermore, the architecture was pushed backwards into legacy "Read Old" patterns just as we are preparing to purge `EntityState` in Phase 2.95 (F4-PR2) - Issue #639.

### The Fix:

Bypassed the legacy `EntityState` entirely by migrating `Zone.name()` to the new event-driven CQRS Read-Models. Restored the deleted snapshot file and re-enabled the regression test to prove state parity.

### Technical Implementation:
- **CQRS Mapping:** Added `name` to the immutable `ZoneState` dataclass in `state_climate.py`.
- **Event Ingestion:** Updated `StateProjector` in `ingestion.py` to route `0004` payloads to the respective zone's state block.
- **Event-Sourced Hydration:** Because snapshot replays ingest packets before zone entities are instantiated, `Zone.name()` now acts as a self-hydrating getter. If `self.zone_state.name` is missing, it explicitly queries `MessageStore.get()`, parses the historical payload, and hydrates its CQRS state for O(1) subsequent reads.

### Testing Performed:
- Explicitly restored `tests/tests_rf/__snapshots__/test_regression_rf.ambr` from `321a5fc0` (the last known good state).
- Re-enabled and successfully ran `test_gateway_replay_regression` to mathematically prove byte-for-byte state parity with the pre-regression architecture.
- Added two new async AAA unit tests in `test_system_zones.py` to independently verify the CQRS O(1) read path and the `MessageStore` hydration path.

Technical debt accumulation. Without this fix, the impending F4-PR2 legacy code lobotomy cannot safely proceed, and the project remains without its primary integration safety net.

### Risks of Implementing:

Minimal. The manual parsing of the `0004` payload inside the hydration fallback assumes consistent dictionary/list structures from the legacy semantic parsers.

### Mitigation Steps:

Implemented strict type checking, dictionary key assertions, and reverse iteration through the message log to guarantee the most chronologically accurate name is captured. Validated via the restored pristine snapshot.

### AI Assistance Disclosure:


PWhite-Eng:
**Note regarding the Snapshot File:**

The snapshot file (`tests/tests_rf/__snapshots__/test_regression_rf.ambr`) included in this PR has **not** been newly generated.

To mathematically prove that this new CQRS logic introduces zero regressions, I explicitly restored the pristine snapshot file from commit `321a5fc0` (the exact state it was in before it was deleted by PR #762).

The re-enabled regression test passes flawlessly against this legacy snapshot, proving our new event-sourced hydration rebuilds the exact same system state byte-for-byte.

silverailscolo (approved):

PR #769, opened 2026-07-04
Refactor and restore RX payload decoder regression test (Fixes CI log overflow)

PWhite-Eng:
### The Problem:

In PR #767, the `test_rx_payload_decoder_regression` test was disabled because its use of `@pytest.mark.parametrize` across 32,000+ raw packet frames completely overwhelmed the CI logs. Expected L2 and schema rejections were generating thousands of `pytest.skip()` console outputs, and the sheer volume of test nodes was severely skewing the local pytest progress bar.

### Consequences:

By leaving this test disabled, we lose a highly valuable, lightning-fast "smoke test" that ensures the OSI layer boundaries and decoupled DTO translations survive real-world data processing without fatal Python crashes.

### The Fix:

Refactored the parametrised test into a single, high-speed "Bulk Processing Test" loop that silently aggregates errors and validates exact baseline thresholds.

### Technical Implementation:
- Removed `@pytest.mark.parametrize`.
- Updated `_load_regression_frames()` to use `enumerate()`, returning a `list[tuple[int, str]]` so we can track the exact line number of every packet.
- Wrapped the packet processing logic in a native Python `for` loop.
- Handled expected L2 and `PacketPayloadInvalid` exceptions silently by incrementing local counters and using `continue`, entirely eliminating CI log spam.
- Appended genuine crashes to an `errors` list, which is printed in a single, clean `pytest.fail()` summary if populated.
- Added strict baseline assertions (`EXPECTED_L2_SKIPS = 10`, `EXPECTED_SCHEMA_SKIPS = 15`) to immediately fail the test if a parser is altered and the rejection rate changes, preventing silent regressions.

### Testing Performed:

Ran the test locally via `pytest`. The test now registers as a single item on the progress bar, processes all 32,325 frames in approximately 1.2 seconds, and outputs a single green tick with zero console spam.

Without this high-speed smoke test active, fundamental crashes at the DTO boundary might slip through basic checks, relying entirely on the slower Syrupy snapshot tests to catch plumbing failures.

### Risks of Implementing:

Because the baseline skip counts are strictly asserted, any future update to the `regression_packets_sorted.txt` fixture file, or any valid improvement to a parser, will cause the test to fail until the constants are manually updated.

### Mitigation Steps:

Added clear comments above the constants explaining that they act as a strict baseline for the static fixture, making it obvious to developers how to update them if the underlying parsing logic is intentionally improved.

### AI Assistance Disclosure:

---

### DIFF
--- a/tests/tests_rf/rx_parity/test_rx_payload_decoder_regression.py
+++ b/tests/tests_rf/rx_parity/test_rx_payload_decoder_regression.py
@@ -14,72 +14,106 @@ FIXTURE_PATH: Path = (
     Path(__file__).parent.parent.parent / "fixtures" / "regression_packets_sorted.txt"
 )
 
+# Baseline metrics for the static fixture file. If these change, a parser
+# has been altered and the regression requires manual review.
+EXPECTED_L2_SKIPS: int = 10
+EXPECTED_SCHEMA_SKIPS: int = 15
 
-def _load_regression_frames() -> list[str]:
+
+def _load_regression_frames() -> list[tuple[int, str]]:
     """Load and sanitize raw packet frames from the regression text file.
 
     Strips out comments and ignores empty lines.
 
-    :raises FileNotFoundError: If the regression file cannot be found at FIXTURE_PATH.
-    :return: A list of sanitized raw frame strings.
+    :raises FileNotFoundError: If the regression file cannot be found at
+        FIXTURE_PATH.
+    :return: A list of tuples containing line numbers and sanitized raw
+        frame strings.
     """
     if not FIXTURE_PATH.exists():
         raise FileNotFoundError(f"Could not find regression file at: {FIXTURE_PATH}")
 
-    frames: list[str] = []
+    frames: list[tuple[int, str]] = []
     with open(FIXTURE_PATH, encoding="utf-8") as file:
-        for line in file:
+        for line_num, line in enumerate(file, start=1):
             raw_frame: str = line.split("#")[0].strip()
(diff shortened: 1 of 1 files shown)


PWhite-Eng:
@silverailscolo As suggested in PR #767, I have re-enabled the massive RX regression test to help protect the codebase during the current refactor.
To address your concerns regarding the CI logs overflowing, I have completely refactored the test architecture. I removed the parametrisation and turned it into a single "Bulk Processing" loop. Expected L2 exceptions and PacketPayloadInvalid rejections are now caught silently via internal counters, meaning there is zero pytest.skip() spam in the CI buffer.
It now executes all 32,325 frames in just ~1.2 seconds as a single test node. To ensure this doesn't create a "silent regression" trap, I added strict assertions at the end of the test that lock in the exact baseline of expected skips (10 L2, 15 Schema). If a parser's behaviour changes, the test will immediately fail and flag the shift.

silverailscolo:
Nice optimization, worth the time spent.
With everything collapsed into a single OK/not OK result, would you be able to find _what_ is failing? Hope to get confirmation that's the case.

Note: The skips only became problematic when I added mark.skip to this test. Before that, a name() mismatch threw 100k+ lines of output to the gh Testing screen so it didn't show the last line. Something had to give, but I understand your twin just describes what it finds.

PWhite-Eng:
If it fails, it will output which packets it failed on, see lines 98 to 109

silverailscolo (approved):

PR #774, opened 2026-07-04
Fix CQRS hydration for controller system_mode (Issue _cc 769)

PWhite-Eng:
### **The Problem:**
Since migrating to the new CQRS event-sourced architecture, the Evohome controller's `preset_mode` and `system_mode` (derived from the 2E04 packet) are not actively populated upon system boot. They default to 'Unknown' until a new state change is organically broadcast over the RF network (Issue ramses-rf/ramses_cc#769).

### **Consequences:**
If left unfixed, Home Assistant will boot with 'Unknown' values for the climate entity's preset and hvac modes. This causes visual user confusion and actively breaks automations that rely on evaluating the system state shortly after a reboot.

### **The Fix:**
Refactored the passive `system_mode` property into an active, self-hydrating asynchronous method that automatically requests the state over the network if the internal CQRS model is currently empty.

### Technical Implementation:
- Removed the `@property` decorator from `SysMode.system_mode` in `tcs.py` and converted it to an `async def` method.
- Added a conditional check: if `self.system_state.system_mode` is `None`, it uses `self._gwy.async_send_cmd()` to dispatch a `2E04 RQ` packet to the physical controller.
- Updated `SysMode.params()` to correctly `await` the new async method.
- Note: No changes were required in `ramses_cc`'s `climate.py` because it retrieves this data via `resolve_async_attr()`, which natively supports both synchronous properties and asynchronous methods.

### Testing Performed:
- Created `tests/tests_rf/test_tcs_hydration.py`.
- Wrote an Arrange, Act, Assert (AAA) test simulating an empty CQRS state, verifying that `mock_gwy.async_send_cmd` is called with the correct `2E04 RQ` command.
- Wrote a secondary test simulating a populated CQRS state, verifying that the cached state is returned instantly without dispatching unnecessary network traffic.
- Updated `test_system_heat.py` to `await` the refactored method, ensuring a fully clean `mypy --strict` pass.

Automations triggering on Home Assistant restarts will fail, and users will lose trust in the integration's reported boot states.

### Risks of Implementing:
Any downstream custom scripts or heavily modified forks that directly access `tcs.system_mode` as a synchronous property without using the `ramses_cc` helper methods will encounter a coroutine error.

### Mitigation Steps:
We verified that the official `ramses_cc` Home Assistant integration already abstracts property/async accesses via the `resolve_async_attr` helper, meaning this change is entirely backwards-compatible for the primary user base. Comprehensive TDD and strict Mypy checks ensure the logic is sound.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/systems/tcs.py
+++ b/src/ramses_rf/systems/tcs.py
@@ -1048,10 +1048,15 @@ class SysMode(SystemBase):  # 2E04
         cmd = Command.get_system_mode(self.id)
         self.discovery.add_cmd(cmd, 60 * 5, delay=5)
 
-    @property
-    def system_mode(self) -> dict[str, Any] | None:  # 2E04
-        """Return the system mode synchronously from Hot State RAM."""
+    async def system_mode(self) -> dict[str, Any] | None:  # 2E04
+        """Return the system mode asynchronously from Hot State RAM.
+
+        If the state is unpopulated (e.g., after boot), an explicit RQ is
+        dispatched to the network to hydrate the CQRS model.
+        """
         if self.system_state.system_mode is None:
+            cmd = Command.get_system_mode(self.id)
+            await self._gwy.async_send_cmd(cmd)
             return None
         return {
             SZ_SYSTEM_MODE: self.system_state.system_mode,
@@ -1083,7 +1088,7 @@ class SysMode(SystemBase):  # 2E04
 
     async def params(self) -> dict[str, Any]:
         params = await super().params()
-        params[SZ_SYSTEM][SZ_SYSTEM_MODE] = self.system_mode
+        params[SZ_SYSTEM][SZ_SYSTEM_MODE] = await self.system_mode()
         return params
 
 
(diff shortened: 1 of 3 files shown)


PWhite-Eng:
Hi @silverailscolo,

This PR resolves Issue ramses-rf/ramses_cc#769, where `preset_mode` and `system_mode` were showing as 'Unknown' upon HA boot following the shift to the new CQRS architecture.

Because `tcs.system_state` requires active telemetry to populate, we were seeing null values on start-up. To fix this, I’ve refactored `tcs.system_mode` from a synchronous `@property` into an `async` method. If the CQRS state is `None`, it now actively dispatches a `2E04 RQ` to the network to hydrate the model.

**Architectural Note:** I intentionally chose to dispatch a physical `RQ` rather than falling back to `entity_state.get_value()` (the legacy MessageStore). This ensures we aren't creating new dependencies on the old local topology store, keeping the path clear for our upcoming "Great Lobotomy" phase where we purge those legacy graphs entirely.

No changes are needed on the HA side since `resolve_async_attr()` happily consumes the new coroutine. I’ve included a new test file (`test_tcs_hydration.py`) to lock in this behaviour. Let me know if you are happy with this approach!

silverailscolo (approved):

silverailscolo:
The RQ was later removed, because it was sent too soon and crashed ramses in 0.57.9

Issue #772, opened 2026-07-04
Investigate: Standardise Battery Status Hydration and Null Handling


PWhite-Eng:
**Context & Problem:**
Following the closure of #764, it is understood that the core library correctly parses the `1060` payload. However, there are two lingering Developer Experience (DX) and integration issues regarding how battery states are passed to Home Assistant:
1. **Null Propagation:** When a battery state is unknown, the integration passes an explicit `None`/`null` value into the `ramses_cc_extra_attributes` dictionary. This actively breaks standard Home Assistant Jinja templates (rendering as a red 0% icon) unless the user has explicitly coded a `default()` filter.
2. **Hydration Delay:** After a cache clear, battery statuses frequently report as 'Unknown' and can take up to 24 hours to repopulate.

**Expected behaviour:**
We should strive for a robust DX that does not rely on users writing defensive template code. Furthermore, the integration should attempt to ascertain battery states earlier in the boot cycle.

**Proposed Action:**
- **DX Fix:** Audit the `ramses_cc` DTO boundary to safely omit the `battery_level` key entirely when the value is unknown, rather than passing a literal `None`.
- **Polling Fix:** Investigate if we can trigger a targeted `RQ` packet for the `1060` payload during the initial discovery phase, mitigating the 24-hour hydration delay.

PWhite-Eng:
I have investigated the null handling and hydration delays regarding the `1060` battery payloads. 

The investigation confirmed two primary issues:
1. Passing explicit `None` values through the DTO boundary was the root cause of the Jinja template crashes in HA.
2. The 24-hour hydration delay was a result of the system passively waiting for a routine `I` packet broadcast from the devices.

**Resolution:**
I've updated `dev_base.py` to completely omit the battery state key if the level is unknown, gracefully side-stepping the HA template issue. To resolve the hydration delay, I've overridden the `_setup_discovery_cmds` logic to enqueue a specific `1060` `RQ` packet during the boot sequence, forcing the devices to report their state immediately. 

I've written a complete suite of `pytest` assertions to validate this behaviour, and took the opportunity to bring the docstrings in `dev_base.py` up to our strict Sphinx standards.

The fixes have been implemented on PR #777

silverailscolo:
Thanks. A derived issue is that there are no Details in the battery binary_sensor, where the code tries to provide the battery_level as  an `extra_state_attribute`.

Fixed in _cc since

PWhite-Eng:
[...]

PR #777, opened 2026-07-04
Fix battery null propagation and accelerate 1060 hydration

PWhite-Eng:
### The Problem:
Currently, when a battery state is unknown, the DTO boundary passes an explicit `{"battery_level": None}` into Home Assistant, which actively breaks user Jinja templates. Furthermore, battery states take up to 24 hours to hydrate after a cache clear because the system passively waits for an `I` (Information) packet (Issue #772).

### Consequences:
If left unaddressed, users will continue to experience silent template failures or visible UI errors in Home Assistant due to `NoneType` propagation. Additionally, users experience significant confusion when battery states appear missing for a full day after a system restart or cache clear.

### The Fix:
Safely omitted missing battery data from the status dictionary entirely, and instructed the gateway to actively poll for battery state during the initial device discovery phase. Additionally, technical debt was addressed by adding comprehensive Sphinx docstrings to `dev_base.py`.

### Technical Implementation:
- Implemented the walrus operator (`:=`) within `BatteryState.status()` to intercept `await self.battery_state()`. If the result is `None`, the `BATTERY_STATE` key is completely excluded from the returned dictionary.
- Overrode `_setup_discovery_cmds()` in both `BatteryState` and `DeviceInfo` to explicitly enqueue a `Command.from_attrs(RQ, self.id, Code._1060, PayloadT("00"))` upon boot, enforcing active rather than passive hydration.
- Applied rigorous `< 79 character` wrapped Sphinx docstrings to all major classes and public APIs in `dev_base.py`.

### Testing Performed:
Created `tests/tests_rf/test_battery_null_handling.py` using strict Arrange, Act, Assert (AAA) methodology. 
- `test_battery_status_omits_key_when_level_is_none`: Asserts the key is completely absent.
- `test_battery_status_includes_key_when_level_is_known`: Asserts valid floats are correctly passed through.
- `test_battery_state_enqueues_1060_rq_during_discovery`: Mocks the discovery queue to assert the `1060` `RQ` command is successfully appended.

Persistent Home Assistant integration crashes for end-users relying on battery monitoring templates, and ongoing bug reports regarding "missing" battery levels after reboots.

### Risks of Implementing:
There will be a slight increase in initial RF traffic during the gateway's boot sequence due to the proactive polling for `1060` and `10E0` packets.

### Mitigation Steps:
The discovery commands are strictly guarded by checking `CODES_BY_DEV_SLUG`, ensuring we only poll devices that are explicitly registered to support these payloads, preventing unnecessary RF congestion or device wake-ups.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/devices/dev_base.py
+++ b/src/ramses_rf/devices/dev_base.py
@@ -73,6 +73,17 @@ class DeviceBase(Entity):
         traits: DeviceTraits | None = None,
         **kwargs: Any,
     ) -> None:
+        """Initialise the device base class.
+
+        :param gwy: The gateway instance managing this device.
+        :type gwy: Gateway
+        :param dev_addr: The physical address of the device.
+        :type dev_addr: Address
+        :param traits: Optional traits to apply during initialisation.
+        :type traits: DeviceTraits | None
+        :param kwargs: Additional arguments for the underlying entity.
+        :type kwargs: Any
+        """
         super().__init__(gwy, **kwargs)
 
         # FIXME: gwy.message_store entities must know their parent device ID
@@ -91,12 +102,14 @@ class DeviceBase(Entity):
         self.power_state = PowerState()
 
     def __str__(self) -> str:
+        """Return a string representation of the device."""
         if self._STATE_ATTR and hasattr(self, self._STATE_ATTR):
             state: float | None = getattr(self, self._STATE_ATTR)
             return f"{self.id} ({self._SLUG}): {state}"
         return f"{self.id} ({self._SLUG})"
 
     def __lt__(self, other: object) -> bool:
+        """Return True if this device's ID is less than the other's."""
         if not hasattr(other, "id"):
             return NotImplemented
         return bool(self.id < other.id)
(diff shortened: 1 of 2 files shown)


silverailscolo:
I thought you couldn't poll battery_state. That had to do with power save, they do not wake up, until you touch them.
We have no RQ or RP 1060 packets in our logger test set.

silverailscolo (approved):
We'll see how quick the battery level will show up, thanks.

silverailscolo (approved):
Another look, agreed

PWhite-Eng:
Hi @silverailscolo,

Thank you for catching that, my friend certainly didn't think of it, and I obviously did not! 

I have updated the PR to revert the `1060` polling logic in `BatteryState._setup_discovery_cmds` and aligned the `pytest` suite accordingly (which passes successfully).

The scope of this PR is now strictly focused on the core issue: intercepting and omitting the `None` values at the DTO boundary to prevent the Home Assistant Jinja template crashes, alongside the Sphinx docstring cleanups for `dev_base.py`.

PR #780, opened 2026-07-04
Refactor: Rewire `Gateway.get_state()` to L7 MessageStore (Phase 2.95 CQRS Bridge)

PWhite-Eng:
**The Problem:**
Currently, `ramses_cc` (the Home Assistant integration) relies on `Gateway.get_state()` to snapshot network state for warm restarts. This method historically gathered state by iterating over the legacy in-memory `_msgs` and `_msgs_ot` dictionaries inside every domain entity (Devices, Zones, Systems). To complete the CQRS architectural migration and drastically reduce RAM usage, we need to delete these legacy dictionaries (Step F5). However, deleting them right now would break the Home Assistant integration's ability to save state.

**Consequences:**
If this bridge is not implemented, advancing the CQRS migration by deleting the legacy entity state arrays will cause silent failures or crashes in `ramses_cc` upon system shutdown/reboot, leading to a complete loss of warm restart capabilities for Home Assistant users.

**The Fix:**
Refactored `Gateway.get_state()` to pull its state payload directly from the new OSI L7 `MessageStore` (the async SQLite WAL database cache) rather than iterating through domain entities.

**Technical Implementation:**
- Redirected the state aggregation logic in `src/ramses_rf/gateway.py` to target `MessageStore.state_cache`.
- Maintained exact return signature parity (`tuple[dict, dict]`) to ensure the `GatewayLifecycle` interface and downstream consumers are satisfied.
- Extracted and mapped values strictly to the legacy `{"code", "verb", "payload"}` schema for JSON `.storage` compatibility.
- Implemented verb filtering to exclusively snapshot `I_` (Info) and `RP` (Response) packets, discarding transient `RQ` and `W` packets.

**Testing Performed:**
- Created `tests/tests_rf/test_phase2_95_get_state_parity.py`.
- Mocked the `MessageStore` cache with various verb types.
- Asserted that `get_state()` accurately returns a tuple where the second element is a dictionary exclusively containing `I_` and `RP` verbs, strictly formatted to legacy JSON serialization expectations.
- Passed all strict `mypy` checks.

**Risks of NOT Implementing:**
The architectural migration to CQRS Read-Models remains blocked. We cannot delete the duplicated entity-level state dictionaries (which currently cause massive memory bloat) without breaking the core Home Assistant integration.

**Risks of Implementing:**
If the data structure returned by the new `get_state()` implementation differs even slightly from what `ramses_cc` expects, the Home Assistant integration will throw JSON serialization errors during shutdown.

**Mitigation Steps:**
- Strictly typed the return signature to match the `GatewayLifecycle` supertype (`tuple[dict[str, Any], dict[str, Any]]`).
- Implemented a dedicated parity test to enforce the exact structure of the nested state dictionaries.
- Used `getattr` fallbacks in the method body to handle cases where the L7 cache is empty or initializing.

**AI Assistance Disclosure:**

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -11,7 +11,7 @@ from collections.abc import Awaitable, Callable
 from logging.handlers import QueueListener
 from typing import TYPE_CHECKING, Any, cast
 
-from ramses_tx import Command, Engine, Packet
+from ramses_tx import I_, RP, Command, Engine, Packet
 from ramses_tx.const import (
     DEFAULT_GAP_DURATION,
     DEFAULT_MAX_RETRIES,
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):
Smart idea

PR #783, opened 2026-07-05
filter HGI on discovery scan

wimpie70:
HGI's were not filtered on discovery_scan

HGI fix: _is_known() now checks the active HGI ID from the transport's SZ_ACTIVE_HGI extra_info, before checking device_registry/known_list/schema. This prevents the HGI from being flagged as a "new discovered device" on startup (it wasn't in any of those lookups when the first packets arrived).

TODO added: Future multiple HGI support will need to check all gateway IDs, not just the single active one

@PWhite-Eng can this be merged, or will it cross your PR's ?


silverailscolo (approved):

silverailscolo:
Let's merge this. Few classes involved

PR #794, opened 2026-07-07
feat: improve HVAC discovery with FAN reply inference + broadcast filtering

wimpie70:
The broadcast/multicast address 63:262142 (hex FFFFFE) was not filtered by _is_valid_address, causing it to appear as a "discovered device" when HVAC devices broadcast `10E0` (OEM info) to it. This is not a real device — it is the RAMSES 'sentinel' equivalent of 255.255.255.255.

Added `63:262142` to the broadcast address filter alongside the existing 18:73030, 18:14803, 18:000730 entries.

---

- [x] AI Coding Assistance disclosure: GLM 5.2 high

### DIFF
--- a/src/ramses_rf/discovery.py
+++ b/src/ramses_rf/discovery.py
@@ -12,7 +12,7 @@ import asyncio
 import contextlib
 import logging
 import random
-from datetime import datetime as dt, timedelta as td
+from datetime import UTC, datetime as dt, timedelta as td
 from typing import TYPE_CHECKING, Any, cast
 
 from ramses_rf.protocol.opentherm import OPENTHERM_MESSAGES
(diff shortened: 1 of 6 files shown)


wimpie70:
See also https://github.com/ramses-rf/ramses_cc/pull/764#issuecomment-4904838860

Problem: _is_known() checked device_registry before known_list/schema. A device removed from the schema was still "known" via its stale registry entry, blocking re-discovery.

Fix: Dropped the device_registry check entirely. "Known" = HGI, or in known_list, or in schema. Registry membership alone no longer suppresses discovery.

Test updated: test_known_in_registry_skipped → test_known_in_registry_only_not_skipped (expectation inverted).

Side effect (desirable per 767): Devices created from traffic (registry-only, never in schema) will now appear in the discovery list for user review.

silverailscolo (approved):
Great to have these clear docstr/comments for a precise process, thanks also for the mypy fix

PR #800, opened 2026-07-10
Reject invalid zone indices (>0B) in _extract_zone_idx

wimpie70:
Zone indices are 00-0B (12 zones max). Values like FC (appliance_control domain), 7F (broadcast), or 0C+ were being accepted as valid zone indices from 3150 payloads, causing devices like 10:064873 (BDR/OTB) to be assigned to non-existent zones.

Also normalises the returned zone_idx to uppercase for consistency.

see https://github.com/ramses-rf/ramses_cc/pull/764


src/ramses_rf/discovery_scan.py
```diff
@@ -53,9 +53,9 @@
     "04": DevType.TRV,
     "07": DevType.DHW,
     "08": DevType.JIM,
-    "10": DevType.BDR,
+    "10": DevType.OTB,
```
silverailscolo, line 56: quite a fix!

silverailscolo (approved):

PR #804, opened 2026-07-11
fix: eliminate discovery loop CPU spike and excessive logging (_cc issue 795)

wimpie70:
Three changes to DiscoveryService.discover() to fix the CPU spiking and log spam reported in ramses_cc issue 795:

1. Use state_cache instead of log_by_dtm for _000A/_30C9 scan: log_by_dtm creates a tuple of ALL messages on every access (O(n)), while state_cache is a dict of latest msg per StateHeader (far smaller). This eliminates the per-cycle full-log allocation.

2. Skip the expensive find_latest_msg scan when the task isn't due: moved the next_due > dt_now check before the message lookup so the O(n) scan only runs when the command is actually due to be sent.

3. Removed the per-cycle "No msg found for hdr" debug log that was the direct cause of the excessive logging reported in the issue.

used AI, didn't test

### DIFF
--- a/src/ramses_rf/discovery.py
+++ b/src/ramses_rf/discovery.py
@@ -286,8 +286,11 @@ class DiscoveryService:
                     if tcs:
                         tcs_id = getattr(tcs, "id", None)
                         if self._gwy.message_store and tcs_id:
+                            # Use state_cache (latest msg per StateHeader) instead of
+                            # log_by_dtm (all msgs) to avoid O(n) full-log scans on
+                            # every poll cycle. See issue 795.
                             found_msgs = []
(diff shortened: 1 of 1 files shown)


wimpie70:
see https://github.com/ramses-rf/ramses_cc/issues/795

I didn't test or review the code...bit in a hurry

silverailscolo:
> 3. Removed the per-cycle "No msg found for hdr" debug log that was the direct cause

.. it also helped to find the code involved, so this part is perhaps not necessary

silverailscolo (approved):
Two smart optimisations (and the 3rd change was free), thanks

PR #812, opened 2026-07-13
Fix: wire `_update_system_state` into dispatcher CQRS ingestion engine

### LINKED ISSUE #800: 0.57.12 - `state` and `_mode`s of the Evohome entity not updating correctly
Comparing ramses_cc 0.57.12 with 0.54.3 looking at the same Evohome system the Evohome controller entity is showing incorrect `state` and also incorrect `system_mode` and `preset_mode` attributes.

The Evohome climate control is therefore displaying the wrong state and mode.

Comparing the state and attributes of the same climate entity with the different integration versions below. Note that `state` and `system_mode` differ.  In this example the `preset_mode` attribute really is `none` but it remains stuck at `none` on 0.57.12 even if changed to a different mode.

Debug and packet logs for the first few minutes after startup attached.

[home-assistant.log](https://github.com/user-attachments/files/29940842/home-assistant.log)
[packet_log.log](https://github.com/user-attachments/files/29940843/packet_log.log)

0.54.3:

Image

0.57.12:

Image

wimpie70:
I've trying to get this list of issues a bit shorter, and the assistant that is currently available to me for free, seems to be quite good at analyzing, so I let it rumble a bit on 
[_cc 800
](https://github.com/ramses-rf/ramses_cc/issues/800)

It found some issues, and a fix, BUT...I'm not sure if this PR is crossing any work @PWhite-Eng is doing on this. So, maybe this is just for ref. and/or may save some time...Hope it helps.

Root cause: PR 786 removed the self-hydrating RQ dispatch from SysMode.system_mode() getter to prevent unbounded task creation, but the dispatcher's _cqrs_ingestion_engine was never wired to call _update_system_state.  This left system_state.system_mode permanently None — the async getter always returned None, and ramses_cc's resolve_async_attr cached that None forever.

The ingestion logic existed in pipeline/ingestion.py (StateProjector), but that engine is not started (Phase 2.75 async cutover is paused). The active engine in dispatcher.py only handled hvac, temperature, demand, and faultlog states — system_state was silently dropped.

Symptoms (ramses_cc issue 800):
- Evohome controller entity shows wrong state and system_mode
- preset_mode stuck at 'none' even when changed
- 2E04 RP packets are received and parsed correctly, but never reach the CQRS read-model

Fix (3 changes in dispatcher.py):
1. Add _update_system_state() — ports the 2E04/0100/313F ingestion logic from pipeline/ingestion.py into the active dispatcher engine
2. Update _resolve_logical_targets() — route system-level opcodes (2E04/0100/313F) to the TCS, since they lack domain_id/zone_idx keys that the existing routing steps rely on
3. Call _update_system_state() from _cqrs_ingestion_engine()

Regression test added to test_tcs_hydration.py verifying that a 2E04 packet payload hydrates system_state via the dispatcher path.

Resolves: https://github.com/ramses-rf/ramses_cc/issues/800

### DIFF
--- a/src/ramses_rf/dispatcher.py
+++ b/src/ramses_rf/dispatcher.py
@@ -29,6 +29,7 @@ from .const import (
     SZ_BYPASS_POSITION,
     SZ_BYPASS_STATE,
     SZ_CO2_LEVEL,
+    SZ_DATETIME,
     SZ_DEVICES,
     SZ_EXHAUST_FAN_SPEED,
     SZ_EXHAUST_FLOW,
@@ -41,6 +42,7 @@ from .const import (
     SZ_HAS_FAULT,
     SZ_INDOOR_HUMIDITY,
     SZ_INDOOR_TEMP,
+    SZ_LANGUAGE,
     SZ_MINUTES,
     SZ_OFFER,
     SZ_OUTDOOR_HUMIDITY,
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):

silverailscolo:
This is a core bug for heat, with a fix in line with other 'states' so will merge for peter to check

PR #823, opened 2026-07-15
fix(813): skip zone binding for CTL (01:) from 000A packets

wimpie70:
The CTL sends `000A` with zone config for multiple zones.  The first 2 hex chars are the zone_idx of the zone being configured, NOT the CTL's own zone.  Setting zone_idx on the CTL corrupted its comment and schema entry (zone kept changing: 02, 0B, 08, 05...).

Fix: skip zone binding updates for 01: devices in both the known-device path and the new-device path.

see https://github.com/ramses-rf/ramses_cc/issues/813

AI used: GLM 5.2 high


silverailscolo (approved):

silverailscolo:
@wimpie70 Is this a showstopper for _cc 0.58.0 release?
I figure it isn't, we'll have a follow up soon after but like to hear your take

wimpie70:
Go ahead, but we know the issue is there

Ah, ic, you already released the _rf bump

PR #826, opened 2026-07-15
fix: Gateway.get_state() missing frame key causes cache restore failure

wimpie70:
Gateway.get_state() in gateway.py returns packet dicts with keys {verb, src, dst, addr1, addr2, addr3, code, payload} but NO frame string.  When _restore_cached_packets (lifecycle.py) passes these dicts to Packet.from_dict, it looks for 'frame' or 'raw_packet' — finding neither, the frame body is empty and raises:

  Bad frame: Invalid structure: >>><<<

This is the real root cause of issue 812 (entities not restoring from cache on HA restart).  The earlier fix (52880567) added a fallback from 'frame' to 'raw_packet' in from_dict, but that was addressing the shadowed GatewayLifecycle.get_state() in lifecycle.py — the coordinator calls Gateway.get_state() from gateway.py which never had either key.

Fix: add "frame": msg._pkt._frame to the dict in Gateway.get_state(). The _LegacyPktShim._frame property reconstructs the full frame string from the message's verb/seqn/addrs/code/payload.

Users with existing broken caches (saved without the frame key) should clear their cache once after upgrading — those caches cannot be retroactively restored because the frame string was never persisted.

See: https://github.com/ramses-rf/ramses_cc/issues/812

AI used: GLM 5.2 high

### DIFF
--- a/src/ramses_rf/gateway.py
+++ b/src/ramses_rf/gateway.py
@@ -298,6 +298,11 @@ class Gateway(GatewayLifecycle, GatewayInterface):
                     "addr3": msg._addrs[2].id,  # <-- Exact raw addr3
                     "code": str(msg.code),
                     "payload": msg.payload,
+                    # Frame string is required by _restore_cached_packets /
+                    # Packet.from_dict to reconstruct the Packet on warm restart.
+                    # Without it, from_dict gets an empty frame body and raises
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):
Another good find

PR #827, opened 2026-07-15
fix(821): reject null device type 63: in Passive Device Scan

wimpie70:
_is_valid_address() only filtered the exact string 63:262142, so 63:262143 (0xFFFFFF, the all-ones sentinel) slipped through and was registered as a phantom DEV device.

Type 63 is NUL — a synthetic sentinel type with no entry in the DevType enum; no real device uses it. Both 63:262142 (0xFFFFFE) and 63:262143 (0xFFFFFF) are broadcast/null sentinels, and the HGI80 emits 63:262143 as a self-disguise address (see tests/tests_rf/test_use_regex.py).

Reject the whole 63: prefix instead of the single exact match. Tests
added for 63:262143 and the full 63: prefix.

see: https://github.com/ramses-rf/ramses_cc/issues/821

AI used: GLM 5.2 high

### DIFF
--- a/src/ramses_rf/discovery_scan.py
+++ b/src/ramses_rf/discovery_scan.py
@@ -688,13 +688,20 @@ class DiscoveryScan:
 def _is_valid_address(dev_id: str) -> bool:
     """Quick check if a device ID looks valid (N.N:NNNNNN or N:NNNNNN).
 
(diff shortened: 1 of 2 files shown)


silverailscolo (approved):
Well documented

PR #835, opened 2026-07-16
Refactor: Phase 2.95 (Partial Lobotomy) - Safe purge of legacy OpenTherm data arrays

PWhite-Eng:
### The Problem:
Our previous attempt to execute Phase 2.95 (Issue #639) by completely deleting the legacy `_handle_msg` routing graph caused a critical regression: `1FC9` bind packets were no longer evaluated, which prevented `_make_tcs_controller()` from firing, ultimately breaking TCS Instantiation for `main_tcs`. We cannot fix the `TopologyBuilder` yet because we need to move Commands out of `ramses_tx` first.

### Consequences:
If we continue to use the legacy `_msgs` data arrays (like `_msgs_ot`), they lock us into the legacy state models, preventing the extraction of CQRS read-models and creating memory overhead. If we delete the routing entirely, the Evohome read-model fails to instantiate, bringing down the core feature of the system.

### The Fix:
We executed a "Partial Lobotomy". We successfully purged the legacy data arrays (`_msgs_ot` and `# TODO[F4-PR2]` markers) from `opentherm_bridge.py`. Crucially, we kept the `_handle_msg` class inheritance and dispatching intact to preserve class promotion and TCS Instantiation via `1FC9`.

### Technical Implementation:
- Deleted `self._msgs_ot` tracking and its initialization from `OtbGateway`.
- Neutered legacy read-time calculation getters (`opentherm_schema` and `opentherm_params`) to hardcode return `{}` so downstream API consumers don't crash.
- Updated `oem_code` to return `None` (removing dependency on `_msgs_ot`).
- Removed `mock_msg` injections into `_msgs_ot` from `test_device_heat.py` and updated the test to hydrate `device.opentherm_state` directly.

### Testing Performed:
- Ran `mypy --strict src` locally (100% pass).
- Ran `pytest tests/tests_rf/test_regression_rf.py` to prove `1FC9` interception still triggers `main_tcs` instantiation.
- Ran full test suite (1,268 passing) after fixing `test_otb_gateway_modulation_quarantine_fallback`.
- Ran `prek run -a` (Ruff and hooks passed).

Technical debt stagnation. We would be stuck unable to move forward to Phase 3 and CQRS isolation because of the `_msgs` array tight coupling.

### Risks of Implementing:
Downstream integrations relying specifically on the raw dictionaries from `opentherm_schema` will now receive empty dictionaries until Phase 5 restores the CQRS views. 

### Mitigation Steps:
We intentionally returned `{}` instead of raising exceptions or deleting the methods so that public API signatures remain stable.

### AI Assistance Disclosure:


src/ramses_rf/devices/opentherm_bridge.py
```diff
@@ -54,17 +54,12 @@
...
-    SZ_MSG_TYPE,
-    SZ_VALUE,
     OtDataId,
-    OtMsgType,
 )
 from .heat_actuators import Actuator, HeatDemand
 
 if TYPE_CHECKING:
-    from ..messages import Message
+    pass
```
silverailscolo: If you're not planning to use it soon, remove the `if TYPE_CHECKING:` block

silverailscolo (changes requested):
Hi @PWhite-Eng good to receive your PR.
All methods removed or now returning `{}/None` are not used anywhere in the code?
Your plan is to move a part to `protocol.opentherm` as it says in a comment?

Isn't part of this code representing a device, so state is in `models.opentherm_state`.

silverailscolo (approved):

PWhite-Eng:
Hi @silverailscolo, thanks for the thorough review and excellent questions! You were spot on to call out the state isolation.

To answer your questions directly:
1. **Safety of Neutered Methods:** I ran a full audit and confirmed no core internal routing or zone-matching logic relies on the `opentherm_schema` or `opentherm_params` getters; they are only used for exporting traits. Returning `{}` temporarily ensures no downstream API consumers crash while we transition.
2. **CQRS State Isolation:** You are completely right about the read-models. The state is actually housed in `state_opentherm.py` (`OpenThermState`), and our test suite is successfully hydrating it.
3. **The Missing Link:** The blueprint for Phase 2.95 dictated that `oem_code` should be wired directly to that new state. I stubbed it to `None` in this PR just to ensure the legacy arrays were safely deleted, but that left the job half-finished and caused the disconnect you noticed.

To close this gap, I have just pushed a commit over on PR #838 that properly wires `oem_code` into `self.opentherm_state.oem_code` (casting to float to preserve the legacy API boundary). This formally completes the state isolation for Phase 2.95. Phase 7 will handle the rest of the complex OpenTherm priority logic.

Thanks again for catching that missing wire up of the code

Phil

wimpie70:
## Root cause analysis: HGI80 QosProtocol timeout for W|01FF

Investigation into the persisting `QosProtocol` timeout for `W|01FF` (bind to Itho Spider) on HGI80 gateways revealed **three contributing factors**, two of which are fixable software bugs addressed in PR 967:

### Bug A: `_patch_cmd_if_needed` missing reverse-patch for HGI80

The existing code only patched **evofw3** gateways (swapping the placeholder `18:000730` for the real HGI ID). The code comment already documented the HGI80 requirement:

> Legacy HGI80s (TI 3410) require the default ID (18:000730), or they will silent-fail.

…but no reverse-patching was implemented. When `from_id` was the actual gateway ID (e.g. `18:002965`), the HGI80 silently dropped the frame — no echo, no reply, QoS FSM times out.

**Fix in PR 967**: For HGI80, if `cmd.addr1` is the real HGI ID, swap it back to the placeholder `18:000730`.

### Bug B: `_is_recent_tx` echo mismatch + `_log_tx_packet` strip bug

PR 957 introduced `_is_recent_tx` for echo suppression, but:
1. The HGI80 substitutes `18:000730` with its real ID in the echo, so the key never matched.
2. `_log_tx_packet` used `strip()` which removed the leading space from ` W` / ` I` verbs, making the frame unparsable — so the TX key was never recorded for any I/W command on **any** gateway.

**Fix in PR 967**: Use `rstrip()` and check both `addr1` variants in `_is_recent_tx`.

### Hardware limitation (not fixable in software)

The HGI80 (TI 3410) may not support transmitting `01FF` frames at all — the code may not be in its firmware whitelist. This is tracked in issue 73.

### References

- ramses_cc PR 867 (original AttributeError fix): https://github.com/ramses-rf/ramses_cc/pull/867
- PR 967 (fixes): https://github.com/ramses-rf/ramses_rf/pull/967
- ramses_cc issue 864: https://github.com/ramses-rf/ramses_cc/issues/864
- ramses_rf issue 73 (01FF not understood): https://github.com/ramses-rf/ramses_rf/issues/73
- ramses_rf PR 957 (echo suppression): https://github.com/ramses-rf/ramses_rf/pull/957
- ramses_rf PR 885 (verb normalization): https://github.com/ramses-rf/ramses_rf/pull/885

PR #839, opened 2026-07-17
fix(parser_0004): include `zone_idx` in payload, so zone names propagate

wimpie70:
see https://github.com/ramses-rf/ramses_cc/issues/822

parser_0004 returned only {SZ_NAME: ...} without SZ_ZONE_IDX, which broke three routing paths that all require zone_idx to route 0004 packets to the correct zone:

  - TCS _handle_msg (systems/tcs.py): returns early for 0004 if SZ_ZONE_IDX not in msg.payload, so zone._name was never set.
  - CQRS StateProjector (pipeline/ingestion.py): the Phase 2.95 Strangler Bridge feeds the projector with legacy Message objects whose payload lacked zone_idx, so zone_state.name was never updated.
  - Dispatcher _calculate_targets (dispatcher.py): step 4 checks "zone_idx" in p, which also failed for 0004.

The new pipeline's decode_packet injects zone_idx via _build_idx_dict, so setups with sqlite_index=True could hydrate zone names retroactively via the MessageStore.  But users with sqlite_index=False had no zone names at all (issue #822).

Now parser_0004 returns {SZ_ZONE_IDX: payload[:2], SZ_NAME: ...} so all three routing paths work regardless of sqlite_index setting.

AI used: GLM 5.2 high


peternash:
Hi @wimpie70 and @silverailscolo - I saw the comment here about `sqlite_index`.  I currently have this in my "Gateway configuration" (test system):
```
sqlite_index: false
```
I don't recall changing that so assume it's the default.  Should I be testing with this `true` or `false` (or trying both)?

wimpie70:
Hi @peternash — good question. The sqlite_index setting is deprecated (Q3 2026) and no longer has any effect. The MessageStore is now always created — your own logs confirm it's running:



SQLiteWorker thread started.
Hydrated memory DB from disk: ramses.db

wimpie70:
checking the failed tests as well
ah, those eavesdrop test fails happen all the time....I'll try to fix them

silverailscolo:
> Should I be testing with this true or false (or trying both)?

You can no longer turn it on in Config since 0.57.x. The role of the Sqlite dbase has changed so the setting should be removed from ramses_rf too.

wimpie70:
@silverailscolo, i'm a bit busy, but I try to catch up with the recent posts soon.

silverailscolo (approved):
Thanks for your 😛time

wimpie70:
you too :smile:

PR #843, opened 2026-07-17
Refactor: Cut-over devices and systems to centralized CQRS Command Dispatcher

PWhite-Eng:
### The Problem:

Currently, the codebase executes domain commands directly against the gateway by assembling low-level packets (e.g., `Command.put_outdoor_temp()`). This tightly couples the domain logic directly to the legacy `ramses_tx` payload structure, violating the separation of concerns. As a result, it restricts future expansion (such as using an MQTT gateway) and obscures intent mapping.

### Consequences:

If this pattern is not addressed, migrating away from the legacy packet layer becomes impossible without rewriting the entire domain model. Furthermore, new features and maintenance will become increasingly error-prone because business logic will continue to bleed into the networking transport layers.

### The Fix:

This PR cuts over the initial bulk of the domain objects (`heat_sensors`, `hvac_sensors`, `hvac_ventilators`, and `zones`) to use the new CQRS-style `CommandDispatcher`. Domain actions now emit standardized `Intent` records instead of raw commands.

### Technical Implementation:
- Swapped all `self._gwy.async_send_cmd` calls in the affected models for `self._gwy.dispatcher.send(intent)`.
- Implemented corresponding parser logic within `ramses_rf/commands/builders/` (e.g., `heat.py` and `zones.py`) that strictly translates generic `Action.*` intents into protocol-specific `CommandDTO`s.
- Encapsulated repeating `Intent` generation into type-safe module helpers: `send_fake_intent` for device faking, and `send_system_intent` for controller-bound zone operations.
- Updated the `CommandDispatcher` to resolve strict method signatures instead of passing unstructured `**kwargs`.
- Updated test suites (`test_device_heat.py` and `test_system_zones.py`) to properly bypass or accommodate the dispatcher pipeline.

### Testing Performed:
- Executed the full automated Pytest suite (1270+ passing tests).
- Confirmed total hex parity against legacy network outputs utilizing the automated parity tests.
- Verified mypy type-checking under `--strict` mode across all refactored domain builders and devices.

The codebase remains tightly-coupled and brittle, causing PR 3 and PR 4 of the migration plan to be permanently blocked. Future integrations involving disparate transport layers will be virtually impossible.

### Risks of Implementing:

Because this effectively severs the legacy `Command` interface for basic operations (setting zone modes, faking temperatures, configuring parameters), regressions in byte-packing could occur if an intent builder was written incorrectly.

### Mitigation Steps:
- Maintained exact bit-for-bit hex compatibility as validated by the legacy parity test snapshot framework (`test_tx_parity.py`).
- Kept the dispatch builders completely stateless and isolated to ensure easy integration testing.
- Preserved existing domain object interfaces and unit tests to ensure external library consumers (like Home Assistant) see no difference in behavior.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/commands/builders/__init__.py
+++ b/src/ramses_rf/commands/builders/__init__.py
@@ -9,7 +9,7 @@ from ramses_rf.commands.core import Command
 from ramses_rf.enums import Action
 from ramses_tx.dtos import CommandDTO
 
-from . import dhw, hvac, zones
+from . import dhw, heat, hvac, zones
 
 # Maps an Action intent to the appropriate payload constructor.
 BUILDERS: dict[Action, Callable[[Command], CommandDTO]] = {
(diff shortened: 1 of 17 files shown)


silverailscolo (approved):
Clear separation

PR #849, opened 2026-07-17
Refactor: Purge Legacy HVAC Commands and Reinstate Exhaustive Edge-Case Builder Tests

PWhite-Eng:
### The Problem:
With the successful introduction of the new CQRS `Intent` framework for HVAC (PR 5B #847), the old legacy `Command` methods (like `set_fan_mode`, `set_bypass_position`, etc.) located in `src/ramses_tx/command/hvac.py` were made obsolete. Leaving them in the codebase violates our architectural shift, causes duplicate logic, and increases maintenance overhead. Furthermore, older tests like `test_apis_hvac.py` were tightly coupled to the legacy `CODE_API_MAP` routing rather than the new builder paradigm.

### Consequences:
If not cleaned up, future developers might accidentally use the deprecated `Command` methods instead of the new CQRS `build_dto` logic. Additionally, deleting `test_apis_hvac.py` entirely (without recovering its test cases) would result in a severe loss of test coverage for complex HVAC packet edge cases (e.g. `22F7`, `22F1`, `12A0`, `1298`, and `31DA`), leading to silent failures during packet transmission in boundary conditions.

### The Fix:
- Completely purged `src/ramses_tx/command/hvac.py` and removed the `HvacMixins` class from the base `Command`.
- Removed all HVAC API routing from `CODE_API_MAP` in `src/ramses_tx/command/__init__.py`.
- Reinstated the "spirit" of the deleted legacy edge-case tests by migrating their extensive `kwargs` dictionaries directly into modern parameterized tests in the new `test_hvac_builders.py`.

### Technical Implementation:
- **Deletion**: Removed `src/ramses_tx/command/hvac.py` and `tests/tests/test_apis_hvac.py`.
- **Test Reinstatement**: Iterated over the deleted dictionaries (e.g., `SET_22F1_KWARGS`, `GET_31DA_KWARGS`) and converted them into `pytest.mark.parametrize` datasets in `test_hvac_builders.py` against the `build_dto(Intent(...))` logic. We padded out missing variables (like `_unknown_fan_info_flags`) to ensure strict typing compliance.
- **Test Cleanup**: Extracted and deleted legacy HVAC-specific asserts from `tests/tests_tx/test_command.py` and cleaned up broken `Command.get_fan_param` mocks in `tests/tests_rf/device/test_hvac_ventilator.py`.
- **Consolidation**: Moved `test_tx_hvac_parity.py` directly into `test_tx_parity.py` for uniform parity assertions.

### Testing Performed:
[...]

Issue #840, opened 2026-07-17
Feature: Multi-Gateway Routing & RF Signal Prioritization


PWhite-Eng:
**Context & Motivation**

Proposal to add feature #289

As `ramses_rf` transitions towards a centralized CQRS architecture, we need to explicitly separate transport-level routing decisions from domain state data. Currently, in environments utilizing multiple gateways, there is no isolated routing logic to determine the optimal path for outbound commands, nor is there strict write-path protection for inbound telemetry.

**Proposed Implementation** To ensure reliable performance in multi-gateway setups, we need to implement the following:
1. **RSSI Routing Table:** Implement an internal routing table within the L3 transport layer. This table should actively track the optimal gateway for each discovered device based purely on passive RSSI metrics and signal quality.
2. **Write-Path Protection (Freshness vs. Signal):** Enforce logic within the CQRS pipeline (e.g., the `StateProjector`) that strictly prioritizes data freshness (latest timestamp) over signal quality. This ensures that the domain state remains completely decoupled from the transport routing metrics, preventing older packets with stronger signals from overwriting newer state data.

**Acceptance Criteria**
- The transport layer can independently select the best transmission gateway based on historical RSSI.
- The state projection layer correctly processes and prioritizes the newest packets, regardless of which gateway received them or their respective signal strengths.

PWhite-Eng:
I'm closing this issue to consolidate the conversation, as it covers the same multi-gateway routing concepts already being actively discussed in #289. The proposals made here regarding the RSSI routing table and write-path protection fit perfectly with the ongoing discussion in #289 about tracking RSSI values for multiple gateways. I will copy the main architectural points over to that thread so nothing gets lost.

PR #831, opened 2026-07-16
fix(830): add data_type 80 (bool32) for 2411 param 07

wimpie70:
ClimaRad MiniBox FAN returns data_type 80 for param 07 (Base ventilation enable). The MiniBox does not support this parameter and signals this with sentinel values (0x000000FF for value, 0xFFFFFFFF for min).

Add a bool32 parser: 0=False, 1=True, anything else=None (consistent with hex_to_bool's FF=N/A convention). Register data_type 80 in _2411_DATA_TYPES alongside the existing 4-byte variants (10/11/13/20/90).

Includes regression tests using the actual packet from issue 830 plus bool 0/1 value decoding tests.

see https://github.com/ramses-rf/ramses_rf/issues/830

AI used: GLM 5.2 high

PR #855, opened 2026-07-18
fix(protocol): move foreign-HGI exception before block_list check

wimpie70:
Foreign HGIs (18: devices) communicate with our controller and the controller's responses (e.g. 0004 zone names, 2349 zone modes) are addressed to the foreign HGI.  The active gateway eavesdrops on these responses to populate zone names and other state.

Previously, the foreign-HGI exception in _is_wanted_addrs was placed AFTER the exclude (block_list) check.  If a caller added a foreign HGI to the block_list, the exclude check would return False before the foreign-HGI exception could fire — silently dropping all packets to/from the foreign HGI, including the controller's 0004 zone name responses.

This was the root cause of ramses_cc issue 822: ramses_cc puts foreign devices (_owner != root_owner) in the block_list, which included foreign HGIs.  The 0004 RP packets (zone names) were dropped at the protocol filter, so zone._name stayed None and the device registry showed fallback names like "01:216136_xx".

Now the foreign-HGI exception is checked FIRST, before the exclude check.  This ensures foreign HGIs are never blocked, even if a caller mistakenly adds them to the block_list.  HGI_DEV_ADDR (18:000730, the generic broadcast address) is still subject to the block_list — only specific foreign HGIs (18:XXXXXX where XXXXXX != 000730) are exempt.

This also aligns with multi-gateway support (issue 840): all 18: devices pass the receive filter, the active HGI is identified for sending, and foreign HGIs are listen-only.

see https://github.com/ramses-rf/ramses_cc/issues/822

AI used: GLM 5.2 High

### DIFF
--- a/src/ramses_tx/protocol/base.py
+++ b/src/ramses_tx/protocol/base.py
@@ -570,27 +570,30 @@ class _DeviceIdFilterMixin(_BaseProtocol):
             self._foreign_gwys_lst.append(dev_id)
 
         for dev_id in dict.fromkeys((src_id, dst_id)):  # removes duplicates
+            # HGI devices (18:) are gateways, not sensors/actuators.
+            # Foreign HGIs communicate with our controller and the controller's
+            # responses (e.g. 0004 zone names, 2349 zone modes) are addressed
+            # to them.  Blocking a foreign HGI would prevent the active gateway
+            # from eavesdropping on those responses (issue 822).
+            #
+            # This check is BEFORE the exclude (block_list) check so that
+            # foreign HGIs are never blocked, even if a caller mistakenly
+            # adds them to the block_list.  HGI_DEV_ADDR (18:000730, the
+            # generic broadcast address) is still subject to the block_list.
+            if dev_id[:2] == "18" and dev_id != HGI_DEV_ADDR.id:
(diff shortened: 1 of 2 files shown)


wimpie70:
Conflict check summary
I verified our changes don't conflict with any open work:

ramses_rf PRs #849-854 (all draft, PWhite-Eng): all focused on command builders / CQRS pipeline refactoring (commands/builders/, systems/zones.py, command/hvac.py). None touch base.py
ramses_rf issue #840 (Multi-Gateway Routing): proposes RSSI routing table and CQRS write-path protection — higher-level than the device ID filter. No mention of _is_wanted_addrs, block_list, or dev_filter. Our fix aligns with this issue (all 18: devices pass receive filter)
ramses_rf issue #639 (L1-L7 Decoupling): proposes DTO boundary between ramses_tx and ramses_rf. No mention of protocol filter or block_list
ramses_rf branch refactor_discovery: only touches src/ramses_rf/system/zones.py (1 file)
ramses_cc PR #831 (Phase 3c flagging): touches coordinator.py at lines ~798 and ~2623 — our change is at line ~1794, no conflict

silverailscolo (approved):

PR #857, opened 2026-07-19
fix(schema): allow I/22C9 from THM and accept mode 00 for Itho Spider (cc 835)

wimpie70:
Itho Spider/Autotemp thermostats send I/22C9 setpoint_bounds broadcasts with mode 00 (off), but the schema only allowed W/22C9 from THM and the parser/regex only accepted modes 01 (heat) and 02 (cool). This caused PacketInvalid errors that disrupted communication after upgrading to 0.58.3.

Changes:
- CODES_BY_DEV_SLUG: add I_ to THM/22C9 (both protocol/ramses.py and protocol_schema.py copies)
- CODES_SCHEMA: accept 0[012] instead of 0[12] in 22C9/2209 regexes (both copies)
- parser_22c9: accept mode 00 mapped to "off", add 000003 to 008-len suffix assert
- Add test cases for I/22C9 mode 00 and mode 01 from issue 835

see github.com/ramses-rf/ramses_cc/issues/835

AI used: GLM 5.2 high


silverailscolo (approved):
Thanks

PR #861, opened 2026-07-19
Refactor(System): Migrate System and Faultlog discovery to L7 CQRS pipeline

PWhite-Eng:
### The Problem:
Currently, the system components (`tcs.py`, `zones.py`, `heat_actuators.py`) rely on legacy synchronous `ramses_tx.Command` string builders for their discovery sequences and polling commands. This tightly couples the `ramses_rf` application layer to the `ramses_tx` transport layer's internal logic, preventing the migration of the command generation pipeline to the new L7 CQRS Intent structure (Phase 3 of the refactoring roadmap).

### Consequences:
Failure to migrate these bounds leaves a circular dependency between L7 and L3. We cannot completely strip out the legacy L3 string builders from `ramses_tx` until all domain entities are using the L7 `Intent` -> `CommandDTO` flow. This effectively blocks Phase 3.25 (Transport Layer Decoupling) and Phase 3.5 (TCS Instantiation Fix).

### The Fix:
Migrated all `System` and `Faultlog` domain boundaries (`tcs.py`, `zones.py`, `heat_actuators.py`) to generate outbound requests using the new L7 `Intent` dataclass mapped to `Action` enums.

### Technical Implementation:
- Replaced `.get_tpi_params`, `.get_mix_valve_params`, `.get_system_time`, `.set_system_time`, and `.set_system_mode` calls with pure `Intent` configurations.
- Injected `LegacyCommandShim.from_dto(build_dto(intent))` to wrap the L7 intent into an object that seamlessly satisfies the existing `discovery.add_cmd()` and `self._gwy.async_send_cmd()` transport reactor signatures.
- Re-injected `data={}` payloads into all `Intent` constructors to strictly satisfy the required dataclass fields, resolving `TypeError` bugs that arose during compilation.

### Testing Performed:
- Validated via the full `pytest` suite and 96 parameterized `syrupy` parity snapshots, verifying 100% bit-for-bit equivalence in the generated raw payloads.
- Ensured `mypy --strict` remains clean.

The decoupling refactor roadmap stalls. The transport layer cannot be "dumbed down" and the schema-driven instantiation fix (Phase 3.5) remains blocked due to the persistent coupling to `ramses_tx.Command`.

### Risks of Implementing:
Using a shim to mask L7 intents as L3 commands introduces temporary cognitive overhead and a slight mapping delay in the discovery loop. There is a risk that unhandled edge cases in the shim translation might drop discovery intervals.

### Mitigation Steps:
We relied on strict bit-for-bit parameterized parity tests (`test_tx_parity.py`) generated in Round 4A to guarantee the L3 payload output remains identical prior to this logic cut-over.

### AI Assistance Disclosure:


silverailscolo:
@PWhite-Eng Here too, to ensure we develop the calls in sync, add a test result in this PR/repo that makes clear what needs to be done over there. Thanks

silverailscolo (approved):

PR #867, opened 2026-07-20
fix(dispatcher): skip device creation for foreign HGI src (issue 822)

wimpie70:
PR 855 (fix/822-foreign-hgi-blocklist) moved the foreign-HGI exception before the block_list check in the protocol-level filter (_is_wanted_addrs in ramses_tx/protocol/base.py), so foreign HGI packets are no longer dropped at the protocol layer.

However, there is a SECOND filter layer: the device-registry filter (check_filter_lists in dev_filter.py).  When enforce_known_list is True, this filter rejects any device not in the known_list — including foreign HGIs.  In instantiate_devices (dispatcher.py), the get_device(src) call for the source is NOT suppressed (the src device is needed for payload routing), so a DeviceNotFoundError here drops the entire packet and adds the foreign HGI to the _unwanted list.

This caused the repeating FILTER EXCEPTION warnings reported in issue 822 comment 5017168119:
  FILTER EXCEPTION: Device 18:072981 failed filter checks:
    Can't create 18:072981: it is unwanted or invalid

The fix: when enforce_known_list is True and the src is a foreign HGI (18:XXXXXX where XXXXXX != 000730 and != active_hgi), skip the get_device(src) call entirely.  The rest of the pipeline (validate_slugs, route_payload) handles src_dev=None gracefully, and the payload is still routed to the dst device (typically our CTL) and processed normally.

When enforce_known_list is False, the foreign HGI is created normally (as an HgiGateway) — existing behaviour is preserved.

See:
  - https://github.com/ramses-rf/ramses_cc/issues/822
  - https://github.com/ramses-rf/ramses_cc/issues/822#issuecomment-5017168119

AI used: GLM 5.2 high


src/ramses_rf/dispatcher.py
```diff
@@ -205,16 +206,48 @@ def instantiate_devices(gwy: Gateway, msg: Message) -> bool:
...
 
+        hgi_id = gwy.hgi.id if gwy.hgi else None
+
         if src_dev is None:
-            # may: DeviceNotFoundError, but don't suppress
-            src_dev = gwy.device_registry.get_device(msg.src.id)
-            if msg.dst.id == msg.src.id:
-                return True
+            # Foreign HGIs (18: devices that are not the active gateway and
+            # not the generic HGI_DEV_ADDR 18:000730) communicate with our
```
silverailscolo, line 213: Great! Code explanation

silverailscolo (approved):

PR #868, opened 2026-07-20
fix: re-parent BDR from hotwater_valve to appliance_control (issue 834)

wimpie70:
When the controller's 000C binding table has a BDR in the HTG slot (domain FA = hotwater_valve) but the BDR is actually the appliance_control (domain FC), the 000C HTG binding would win the race and incorrectly bind the BDR as hotwater_valve to a spuriously-created DhwZone.  Subsequent 000C APP bindings or 3B00/3EF0 eavesdrop were blocked by the parent-change guard in _apply_topology_link, and the SystemSchemaInconsistent error was silently swallowed.

This fix allows re-parenting a BDR from DhwZone.hotwater_valve (FA) to System.appliance_control (FC) when the DhwZone has no DHW sensor — if a sensor exists, the system genuinely has DHW and the BDR may truly be the hotwater_valve, so re-parenting is suppressed.

Changes:
- topology_builder.py: add child_id=FC to eavesdrop rule metadata so the binding targets the System's appliance_control slot
- dev_registry.py get_device: re-parent BDR from DhwZone (hotwater_valve) to System (appliance_control) when child_id=FC and DhwZone has no sensor; clean up empty DhwZone after detach
- dev_registry.py _handle_bind_device: fallback child_id=FC when domain_id=FC but child_id missing from metadata
- 3 new tests covering re-parenting, no-reparent-when-sensor-exists, and the normal no-race case

Complements PR #856 (discovery_scan domain_id=FC tagging) which handles the scan/classification layer; this fix handles the topology binding layer.

See https://github.com/ramses-rf/ramses_cc/issues/834
AI used: GLM 5.2 high

### DIFF
--- a/src/ramses_rf/devices/dev_registry.py
+++ b/src/ramses_rf/devices/dev_registry.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING, Any, cast
 
 from ramses_rf.address import Address, is_valid_dev_id
 from ramses_rf.config import GatewayConfig
-from ramses_rf.const import SZ_DEVICES
+from ramses_rf.const import FA, FC, SZ_DEVICES
 from ramses_rf.devices.dev_base import DeviceHeat, DeviceHvac, Fakeable
 from ramses_rf.enums import TopologyAction
 from ramses_rf.exceptions import (
@@ -196,6 +196,11 @@ class DeviceRegistry:
                 # 1. FORWARD BINDING (Delegating down to legacy graph mutation
                 # to allow the Old Brain to hydrate itself)
                 child_id_raw = metadata.get("child_id")
+                # Fallback: if domain_id is FC (appliance_control) but
+                # child_id was not explicitly set in the metadata, use FC
+                # so the binding targets the System's appliance_control slot.
+                if child_id_raw is None and metadata.get("domain_id") == FC:
+                    child_id_raw = FC
                 child_dev = self.get_device(
                     event.child_id,
                     parent=cast("Parent", parent),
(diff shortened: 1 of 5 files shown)


wimpie70:
@PWhite-Eng - you may want to review this

PWhite-Eng:
Hi @wimpie70,

This is my friends thoughts - Thank you for putting this together. The PR successfully resolves the race condition reported in issue #834, and the new tests provide excellent regression coverage for this scenario.

While the logic is functionally correct, I have a few architectural and encapsulation concerns regarding how the graph mutation is handled. I would prefer to see these addressed in this PR if possible, so we can ensure the solution aligns with our architectural standards:
1. **Encapsulation and Private Attribute Mutation:** In `dev_registry.py` (`get_device`), the fix directly manipulates the private properties of the parent objects to clean up the graph (e.g., `old_parent._dhw_valve = None`, `old_parent.childs.remove(dev)`, `old_parent.child_by_id.pop(...)`, and `tcs._dhw = None`). To maintain object-oriented encapsulation, state mutation should remain within the object itself. Could we abstract this graph detachment logic into a dedicated method (e.g., `detach_child()` or `remove_device()`) on the `DhwZone` and `Evohome` classes, so they manage their own referential integrity and cleanup?
2. **Type Checking:** Checking the class type via `type(old_parent).__name__ == "DhwZone"` is generally an anti-pattern. Where possible, please use `isinstance(old_parent, DhwZone)` or rely on a specific interface property to avoid this.
3. **Separation of Concerns:** `get_device()` is fundamentally a factory and retrieval method. Embedding complex graph detachment, re-parenting transitions, and cleanup logic directly into this method overloads its responsibilities. Abstracting the detachment logic, as mentioned above, will help keep this cleaner.

Could you look into refactoring the detachment logic to use dedicated methods on the parent classes?

Thanks again for the great work on identifying and fixing this.

wimpie70:
Would you mind writing a replacement ? I just tried ti fix the issues, but not wanne dive in too deep. Got some other work to attend to

PWhite-Eng:
Hi @wimpie70,

No problem at all! I completely understand that you have other work to attend to and do not want to dive too deep into the registry architecture right now.

This PR cam be merged as-is, since it successfully resolves the issue and the tests are great.

I will take care of writing the replacement for the architectural refactor. I have raised issue #873 to track those encapsulation and graph detachment tweaks, and I will look to implement them soon.

Thanks again for investigating and providing the initial fix!

wimpie70:
If you are happy with this you can close issue https://github.com/ramses-rf/ramses_rf/issues/873

silverailscolo (approved):

PR #869, opened 2026-07-20
fix: hydrate DhwZone CQRS read-models for DHW opcodes (issue 843)

wimpie70:
The Phase 2.95 CQRS cutover redirected DhwZone getters (temperature, setpoint, mode, config) from the legacy SQLite message store to in-memory CQRS read-models (temp_state / dhw_state).  However, the CQRS ingestion engines (dispatcher._cqrs_ingestion_engine and
pipeline.ingestion.StateProjector) never routed DHW opcodes to the DhwZone, because 1260/10A0/1F41 payloads carry no zone_idx/domain_id.

As a result, the DhwZone read-models stayed empty and the ramses_cc water_heater entity reported None for current/target temperature (while the underlying DHW sensor entity still showed its value).

Route 1260/10A0/1F41 to src_dev.tcs.dhw when the source is the DhwSensor (1260) or the Controller (10A0/1F41).  The appliance_control (OTB) also emits 10A0/1260 with different semantics (CH setpoint / null temp) and is excluded to avoid clobbering the DHW read-models.

Also add _update_dhw_state to the legacy dispatcher (mirroring ingestion.py) so dhw_state (setpoint/mode) is hydrated there too, not only via the StateProjector bridge.

Regression tests cover 1260 (temperature), 10A0 (setpoint), 1F41 (mode) and the OTB non-clobbering scenario.

See: https://github.com/ramses-rf/ramses_cc/issues/843

AI used: GLM 5.2 high


wimpie70:
@PWhite-Eng Can you review this one too ?

i'll check on 0008 and 0009 as well

wimpie70:
We found 3 more issues (fixed in this PR):
1. relay_demand mapped to wrong field — dispatcher._update_demand_state mapped relay_demand → heat_demand instead of relay_demand. Result: DhwZone.relay_demand returned None, and heat_demand was wrongly set to the relay value.

2. 0009 key mismatch — parser_0009 produces "failsafe_enabled" but both _update_demand_state functions checked for "relay_failsafe". The key never matched, so 0009 was never ingested and DhwZone.relay_failsafe always returned None.

3. No domain_id routing in ingestion path — StateProjector.process_message_state had no F9/FA → DhwZone routing (only the dispatcher path had it, via _resolve_logical_targets step 5). So even with the field mapping fixed, the ingestion path never reached the DhwZone for 0008/0009/3150 with domain_id.

After the fix (verified against _heat_trv_00 packet log):

DhwZone.relay_demand = 0.1 (from 0008 F914)
DhwZone.relay_failsafe = False (from 0009 array with F9 domain)
DhwZone.heat_demand = None (no longer wrongly set from relay_demand)
Regression test added: test_dhw_zone_relay_demand_and_failsafe_hydrated.

PWhite-Eng:
Hi @wimpie70,

Thank you for the detailed PR. I have reviewed the code, and it definitively fixes the problem reported in Issue #843. By correctly implementing the routing for DHW opcodes and domain-based opcodes, the `DhwZone` read-models are now successfully hydrated. The newly added regression tests are excellent and comprehensively verify the scenarios.

I did note a couple of minor architectural tweaks that would align the implementation closer to industry best practices (specifically around the DRY principle for the duplicated routing logic, and some explicit type checking). Since you don't have the time to tackle them right now, I have documented these improvements in issue #872 to be picked up as a fast follow-up (that I will hope to get to soon).

Thanks again for tracking this down and writing such thorough tests!

silverailscolo (approved):

wimpie70:
checking the changes with ha_sim_test...

silverailscolo:
Rerun Testing action after merging #877. All tests Passed before that point.
Result: Testing Passed

PR #888, opened 2026-07-21
fix: normalize verb in CommandDTO.__str__ and Engine.create_cmd (issu…

wimpie70:
…e 835)

The Command architecture migration (commit 2a86a4833) dropped the verb normalisation that the old Command._from_attrs did:

    verb = I_ if verb == "I" else W_ if verb == "W" else verb

Without this, passing verb="W" (without leading space) — as the ramses_cc send_packet service allows — produces a malformed frame:

    W --- 18:002965 21:057310 --:------ 01FF 026 ...

The HGI80 expects 2-character right-justified verbs (" W"), so it silently drops the frame. No echo comes back, and the QoS FSM times out after 20 seconds:

    Send timed out for 01FF| W|21:057310:
    <ProtocolContext state=WantEcho ...>: Expired global timer after 20.0 sec

The QoS header (tx_header) was correct because Packet._from_cmd uses cmd.verb.strip():>2, but the actual frame sent to the HGI80 used CommandDTO.__str__ which did {self.verb} without normalisation.

Changes:
- CommandDTO.__str__: strip and right-justify the verb to 2 chars (mirrors Packet._from_cmd formatting)
- Engine.create_cmd: convert "I"/"W" to I_/W_ (VerbT) before passing to CommandDTO, restoring the old Command._from_attrs behaviour
- fsm.py: remove leftover debug print() in WantEcho.pkt_rcvd

https://github.com/ramses-rf/ramses_cc/issues/835

AI used: GLM 5.2 high

### DIFF
--- a/src/ramses_tx/dtos.py
+++ b/src/ramses_tx/dtos.py
@@ -81,9 +81,17 @@ class CommandDTO:
     num_repeats: int = 1
 
     def __str__(self) -> str:
-        """Return the string representation of the frame to be transmitted."""
+        """Return the string representation of the frame to be transmitted.
+
+        The verb is stripped and right-justified to 2 characters to match the
+        RF protocol format expected by the HGI80 (e.g. ``" W"`` not ``"W"``).
+        Without this normalisation, verbs passed as plain strings (e.g. from
+        the ramses_cc send_packet service) produce malformed frames that the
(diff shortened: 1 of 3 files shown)


wimpie70:
this is now part of https://github.com/ramses-rf/ramses_rf/pull/885

PR #909, opened 2026-07-22
perf(rf): use cached id_to_address factory in Message.__init__

PWhite-Eng:
### The Problem:
`Message.__init__` in `src/ramses_rf/messages/base.py` currently instantiates `Address` objects directly via `Address(DeviceIdT(addr1))` for positional L2 MAC fields on every incoming packet. Direct `Address(...)` instantiation executes `Address.is_valid` (which triggers regex validation) 3 times per packet, bypassing the `@lru_cache` mechanism in `id_to_address`.

### Consequences:
For large packet logs (e.g. 30,000+ lines), 90,000+ redundant regex matches and object instantiations occur, wasting CPU cycles and slowing down packet processing and test runner performance.

### The Fix:
Updated `Message.__init__` in `src/ramses_rf/messages/base.py` to use the cached factory function `id_to_address(DeviceIdT(...))` instead of instantiating `Address` directly.

### Technical Implementation:
- Imported `id_to_address` from `ramses_rf.address`.
- Replaced direct `Address(DeviceIdT(...))` calls in `Message.__init__` with `id_to_address(DeviceIdT(...))`.

### Testing Performed:
- Benchmark profiling on `tests/tests_rf/test_regression_rf.py` measured a 1.71-second reduction in pytest execution time and **1.993 seconds of raw CPU processing time saved** across 32,325 log lines.
- Verified test suite passes via `.venv/bin/pytest tests/tests_tx/test_packet.py`.

Continued redundant regex evaluation and object instantiation on every packet processed by `Message.__init__`.

### Risks of Implementing:
Low risk. `id_to_address` returns identical `Address` instances from the cache.

### Mitigation Steps:
`id_to_address` delegates directly to `Address(device_id)` when a cache miss occurs, ensuring exact functional equivalence.

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_rf/messages/base.py
+++ b/src/ramses_rf/messages/base.py
@@ -9,7 +9,7 @@ from dataclasses import dataclass
 from datetime import datetime as dt
 from typing import TYPE_CHECKING, Any, TypeAlias, TypeVar
 
-from ramses_rf.address import Address
+from ramses_rf.address import Address, id_to_address
 from ramses_tx import CommandDTO, PacketDTO
 from ramses_tx.models import DeviceId, RawPacket, TransportMessage
 from ramses_tx.typing import DeviceIdT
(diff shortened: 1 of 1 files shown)


silverailscolo (approved):
A lot of description for this change, can you limit this (evident improvements need not spell out what we lose by rejecting, and what it says is often quite general).
Also the big numbers

PR #946, opened 2026-07-29
fix(transport): handle TransportError during closing and validate topic_sub before MQTT unsubscribe

PWhite-Eng:
### The Problem:
1. When a transport is closing or closed during reload/teardown, `_pkt_read()` raises `exc.TransportError("Transport is closing or has closed")`. `_frame_read()` in `_ReadTransport` did not catch `exc.TransportError`, causing unhandled exceptions in Home Assistant's event loop callback wrapper (`Error doing job: Exception in callback _ReadTransport._frame_read()`).
2. `MqttTransport._close()` attempted to call `self.client.unsubscribe(self._topic_sub)` even when `self._topic_sub` was empty `""`, raising `ValueError: Invalid topic.`.

### The Fix:
1. Updated `_frame_read()` in `src/ramses_tx/transport/base.py` to catch `exc.TransportError` when closing and log at `DEBUG` level.
2. Updated `_close()` in `src/ramses_tx/transport/mqtt.py` to check that `self._topic_sub` is non-empty before calling `unsubscribe()`.

### Testing Performed:
- Verified docstring compliance (`ruff check --select D`).
- Passed strict `mypy` type checking (`mypy --strict`).
- Verified `prek` pre-commit checks.
- Validated via `ha_sim_test` integration test suite, eliminating 100% of transport callback errors (reduced unexpected errors from 293 down to 0).

### AI Assistance Disclosure:

### DIFF
--- a/src/ramses_tx/transport/base.py
+++ b/src/ramses_tx/transport/base.py
@@ -169,7 +169,11 @@ class _ReadTransport(_BaseTransport, TransportInterface):
             _LOGGER.warning("%s < PacketInvalid(%s)", frame, err)
             return
 
-        self._pkt_read(pkt)
+        try:
+            self._pkt_read(pkt)
+        except exc.TransportError as err:
+            _LOGGER.debug("%s < Transport Error(%s)", pkt, err)
+            return
 
     def _pkt_read(self, pkt: Packet) -> None:
         """Pass any valid Packets to the protocol's callback."""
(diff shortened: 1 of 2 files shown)


PWhite-Eng:
Hi @silverailscolo and @wimpie70,

I have opened this PR to resolve a couple of transport-layer exceptions identified while running the `ha_sim_test` suite from `ramses_extras`.

Specifically, this patch:
1. Prevents `TransportError` from surfacing as unhandled callback exceptions during transport shutdown or reload.
2. Ensures `_topic_sub` is non-empty prior to unsubscribing during MQTT transport cleanup.

Both changes have been verified against the unit test suite and simulator harness.

silverailscolo (approved):

PR #967, opened 2026-08-01
fix(tx): HGI80 reverse-patch source addr and match echoes with addr substitution

wimpie70:
## Summary

Three fixes for HGI80 (TI 3410) gateways, traced to the `QosProtocol` timeout for `W|01FF` bind to Itho Spider (ramses-rf/ramses_cc#864, ramses-rf/ramses_rf#835).

**1. `_patch_cmd_if_needed`: reverse-patch for HGI80**

The existing code only patched evofw3 gateways (placeholder `18:000730` → real HGI ID). HGI80 firmware requires `18:000730` as the source address — using the real HGI ID causes a silent drop and WantEcho timeout. Now, for HGI80, the real HGI ID is swapped back to the placeholder.

**2. `_is_recent_tx`: match HGI80 echoes with addr substitution**

PR 957's `_is_recent_tx` didn't match HGI80 echoes where `addr1` is substituted (real ID instead of placeholder). Now both variants are checked.

**3. `_log_tx_packet`: use `rstrip` instead of `strip`**

`strip()` removed the leading space from ` W` / ` I` verbs, making the frame unparsable — TX key was never recorded for any I/W command on any gateway. Now `rstrip()` is used.

## References

- ramses-rf/ramses_cc#867 (original AttributeError fix)
- ramses-rf/ramses_cc#864
- ramses-rf/ramses_rf#835
- ramses-rf/ramses_rf#73 (01FF not understood — HW limitation)
- ramses-rf/ramses_rf#957 (echo suppression)

## Test plan

- [x] `test_patch_cmd_if_needed_hgi80_reverse` — HGI80 swaps real ID back to placeholder
- [x] `test_patch_cmd_if_needed_hgi80_no_change_when_placeholder` — no-op when already placeholder
- [x] `test_patch_cmd_if_needed_hgi80_no_change_when_impersonating` — no-op for non-gateway source
- [x] `test_is_recent_tx_matches_hgi80_echo_with_addr_substitution` — echo with real HGI ID matches TX with placeholder
- [x] `test_is_recent_tx_matches_evofw3_echo_directly` — evofw3 echo still matches directly
- [x] `test_is_recent_tx_rejects_unrelated_frame` — unrelated frame rejected
- [x] `test_hardware_echo_logging_suppressed_hgi80_addr_substitution` — end-to-end echo suppression
- [x] Full test suite: 1445 passed, 4 skipped
- [x] pre-commit: all hooks pass

### DIFF
--- a/src/ramses_tx/protocol/base.py
+++ b/src/ramses_tx/protocol/base.py
@@ -9,6 +9,7 @@ logging and device ID filtering mechanisms.
 from __future__ import annotations
 
 import asyncio
+import dataclasses
 import logging
 import re
 from collections import deque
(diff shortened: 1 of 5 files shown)


wimpie70:
@PWhite-Eng you may want to check this.

silverailscolo (approved):

PR #969, opened 2026-08-02
fix: guard MQTT `_on_message` against _closing race, add PowerState to `apply_state_update`

wimpie70:
## Summary

Two fixes for bugs found via ha_sim_test:

1. **MQTT transport `_closing` race**: the paho-mqtt network thread may still deliver messages to `_on_message` after `_close()` has set `_closing=True` (race during gateway reload). Without a guard, these stale callbacks log "Transport Error" for every inbound packet until the thread fully exits. Add an early return in `_on_message` when `_closing` is True.

2. **Missing `PowerState` in `apply_state_update`**: the CQRS entity method handled `TemperatureState`, `DemandState`, `ScheduleState`, `FaultLogState`, `OpenThermState`, `HvacState`, and `ZoneState`, but was missing `PowerState`. The 1060 battery packet restore path sets `power_state` directly and emits a `StateUpdatedEvent`, but the event was silently ignored because no branch matched `PowerState`. Add the missing branch so battery state is properly ingested via the CQRS pipeline.

#### Test plan
- [x] `ruff check` and `mypy --strict` pass
- [x] `pytest .` — 1446 passed, 4 skipped
- [ ] ha_sim_test R27/R47 (transport) and R32 (battery restore) pass with these fixes


wimpie70:
AI used: GLM 5.2 high

silverailscolo (approved):

wimpie70:
## Additional context for review

This PR now has two commits — the second commit (`befaf3d2`) adds three fixes found during ha_sim_test parallel runs:

### 1. Deferred `/rx` subscribe on LWT online
When the broker connects and establishes the protocol connection **before** the ramses_esp sends its "online" LWT message, the `/rx` subscription was never made (race introduced by the issue 871 fix in the first commit). Now `_on_lwt_online` subscribes to `/rx` if `_topic_sub` is not yet set, and unsubscribes the data wildcard if present.

### 2. Non-blocking `loop_stop()` via daemon thread
`loop_stop()` joins the paho-mqtt network thread, which can block for up to 1s (the thread's `select()` timeout). Under parallel test load with frequent ramses_cc reloads, this blocked the HA event loop and caused cascading timeouts. Running it in a daemon thread avoids the block. The `_closing` guard in `_on_message` (from the first commit) ensures the old transport won't process messages even if the thread hasn't fully exited when the new transport starts.

Also: `disconnect()` is now always called, even when `_connected` is False. The paho-mqtt network thread may still be alive and auto-reconnecting with the same client ID, which would kick the new transport off the broker.

### 3. Drain pending handler tasks after packet restore
`GatewayLifecycle.restore` now `await`s `asyncio.gather()` on `tmp_protocol._handler_tasks` after restoring cached packets. Without this, entity properties may query `power_state` before the async handler tasks finish processing 1060 battery packets, causing flaky `'unknown'` states (R32 battery restore test).

### Test results
- Full parallel run (4 containers, 62 recipes): **381 passed, 3 failed**
- The 3 remaining failures are: R34 (pre-existing 000C HTG issue), R31/R47 (transient timeouts now fixed in ramses_extras PR 125)

silverailscolo (approved):

PR #970, opened 2026-08-02
Bump version to 0.59.2

silverailscolo:


silverailscolo:
@wimpie70 I am planning a new release. Can you run the sim test suite and report here whether it runs OK (or not)? Thanks

wimpie70:
nope, sorry, can't do.
It still has fails that I've been trying to solve all day. Not reliable atm.
And...probably won't have time to check it anymore until next weekend.

silverailscolo:
> nope, sorry, can't do. It still has fails that I've been trying to solve all day. Not reliable atm. And...probably won't have time to check it anymore until next weekend.

Will wait np
Take your time, enjoy the weather. We'll get there

wimpie70:
I have ha_sim_test stable again on my local, but I notice we have some PR's on the way...
I'll release the current _extras master as a pre-release...0.21.0 (@PWhite-Eng )
check for the .md file to find the extra parameters (time-factors, simultaneous containers...)

wimpie70:
Ok, I should make a note here. I checked how things were fixed on the ha_sim_test. and i'm not completely happy yet. Because the traffic on MQTT can be a bit too much some msg may be dropped and we have retries to cover this. But this may camouflage actual timing/race conditions.
R06, R24, R28, R38 are easy to convert to wait_for, but R10, R34, R40 need more investigation.
I'll try to fix this...

PR #973, opened 2026-08-04
fix(zones): check `known_list` class instead of address prefix for sensor/actuator binding

wimpie70:
## Summary
- PR 965 (eavesdrop decouple) added an address-prefix filter in
  `Zone._update_schema` that skips binding sensors/actuators whose
  address starts with `01:`/`02:`/`18:` to prevent CTLs/UFCs/HGIs
  from being bound as zone sensors.
- This filter is too broad: a faked device may have a
  controller-class address prefix (e.g. `01:150003`) while being
  explicitly classed as a sensor (e.g. THM) in the known_list.
  The address-based filter prevents these faked sensors from being
  bound to their parent zone, causing `zone_idx` to fall back to
  `"00"` in `set_temperature`.
- Replace with a two-tier check: if the device is in the known_list,
  check its class (only skip if CTL/UFC/HGI); otherwise fall back to
  the address prefix check so eavesdropped controllers are still
  filtered out.

#### Test plan
- [x] `prek run -a`: passed
- [x] `mypy --strict`: passed (no new issues)
- [x] `pytest`: 657 passed, 4 skipped
- [x] ha_sim_test parallel (4 containers, 62 recipes):
      386 passed, 1 failed (R34 transient 000C HTG race)
- [x] R38 now passes: 30C9 payload starts with zone_idx `03`


wimpie70:
R34 still fails...checking that now

silverailscolo (approved):

PR #994, opened 2026-08-07
refactor(const): Decouple protocol vs domain constants and enums (PR 1A)

PWhite-Eng:
### The Problem
Protocol payload constants and domain enums were mixed between `ramses_tx.const` and `ramses_rf.const`, and `ramses_tx.address` imported `DEV_TYPE_MAP` from `ramses_rf`, violating strict architectural layer separation.

### The Fix
- Relocated low-level protocol payload constants to `ramses_tx.const` and domain enums/constants to `ramses_rf.enums` and `ramses_rf.const`.
- Re-pointed imports across `src/ramses_rf`, `src/ramses_cli`, and `tests/`.
- Decoupled `ramses_tx.address` from higher-level `DEV_TYPE_MAP` domain lookup.
- Coordinated via ramses-rf/ramses_rf#992 (ref: ramses-rf/ramses_rf#639, ramses-rf/ramses_rf#930, ramses-rf/ramses_cc#884, ramses-rf/ramses_cc#868).

### Testing Performed
- `mypy --strict`: Success (0 issues in 138 source files).
- `prek run -a`: Passed all checks.
- `pytest`: 1457 passed, 4 skipped.

### AI Assistance Disclosure


src/ramses_rf/devices/dev_base.py
```diff
@@ -36,18 +32,17 @@
 )
 from ramses_rf.topology import Child
 from ramses_tx import CommandDTO, Packet, Priority, QosParams
+from ramses_tx.const import (
+    SZ_BATTERY_LEVEL,
```
silverailscolo, line 36: Why is this type of const in _tx, as it carries meaning of contents ("domain"), not just frame properties?
Same question for similar terms (SZ_TEMPERATURE, ...).

I understand these consts were always here, you just import directly and not via _rf/const. But as we are cleaning up? None of these examples is used in _tx

@PWhite-Eng see this remark

PWhite-Eng: `@silverailscolo` You are completely right. A code audit confirmed that none of the payload schema keys (`SZ_TEMPERATURE`, `SZ_SETPOINT`, `SZ_HEAT_DEMAND`, etc.) are used inside `ramses_tx`—all payload parsers reside in `ramses_rf/parsers/`.

We have reviewed the location of these constants and relocated all domain payload string keys (`SZ_*`) from `ramses_tx.const` to `ramses_rf.const`, leaving `ramses_tx.const` with strictly low-level wire-format frame constants (`Verbs`, `Code`, `Addr`, `Header`, `SZ_ACTIVE_HGI`, `SZ_IS_EVOFW3`, `SZ_SIGNATURE`). All imports across `ramses_rf` parsers, read-models, devices, systems, CLI, and test modules have been updated accordingly, preserving all legacy comment markers verbatim.

Rather than forcing this change through the existing stacked Phase 5 PR series, this refactoring has been submitted as an additional stacked draft PR: ramses-rf/ramses_rf#999 (stacked on ramses-rf/ramses_rf#998).

silverailscolo (approved):

PR #1028, opened 2026-08-11
fix: 30C9 decoder acceptance for non-controllers + sensor-sourced zone temp hydration (issues 927, 929)

wimpie70:
## Summary

This PR fixes two issues:

### 1. Faking Zone THM broken (ramses-rf/ramses_cc#929)

The 0xAB guard in `_pkt_idx` rejected any non-zero idx from non-controller devices (03:/04:/12:). This broke faked THM sensors after commit 5b9abbe4 (0.59.2), which started sending 30C9 with the parent zone's idx instead of the hard-coded `"00"`. The decoder raised `PacketPayloadInvalid` and the fake command failed entirely, leaving all faked zones in "waiting for sync".

`CODE_IDX_ARE_SIMPLE` codes (e.g. 30C9) have schemas that explicitly permit a non-zero idx. Move the SIMPLE check before the 0xAB guard so these packets are accepted. The idx is not injected into the result dict (that remains gated by `_has_ctl` in `_build_idx_dict`), but the packet is no longer rejected.

This also fixes 3 real 30C9 packets from 03: sensors in the regression fixture that were being wrongly rejected.

### 2. Climate temperature not displayed for DT4R thermostat (ramses-rf/ramses_cc#927)

The CQRS ingestion pipeline and state_projector only routed controller-sourced 30C9 (with `zone_idx`) to zones. Sensor-sourced 30C9 (e.g. from a 22: DT4R thermostat) has no `zone_idx` in the decoded payload because the sensor is not a controller, so `_build_idx_dict` injects no `zone_idx`. Without routing, the zone's `current_temperature` stayed None when the controller didn't broadcast 30C9 for that zone.

- Added `Code._30C9` to the zone routing path's `_update_temperature_state` call (for controller-sourced 30C9 with `zone_idx`)
- Added a new path to route 30C9 from a sensor device to its parent zone via `src_dev._parent` (for sensor-sourced 30C9 without `zone_idx`)

In 0.55.3, `Zone.temperature` queried the msg_db directly for 30C9 packets from either the controller or the sensor. The CQRS cutover in 0.59.x broke this by reading from `temp_state`, which was only hydrated from controller-sourced 30C9.

#### Test plan

- [x] `prek run -a` passes (ruff, codespell, mypy --strict)
- [x] `pytest .` — 1668 passed, 4 skipped
- [x] New unit tests: `test_30c9_non_controller_with_zone_idx_accepted` (3 parametrized cases), `test_30c9_controller_still_injects_zone_idx`, `test_zone_temperature_hydrated_from_sensor_30c9`
- [x] Regression baseline updated: `EXPECTED_SCHEMA_SKIPS` 15 to 11, snapshot updated (03: thermostat sensors now discovered)
- [x] ha_sim_test R69: faked 03: THM 30C9 accepted with correct zone_idx
- [x] ha_sim_test R70: 22: sensor 30C9 accepted by decoder with correct temperature
- [x] ha_sim_test R38: 01: faked THM 30C9 still works


silverailscolo (approved):
Great fix, thanks

PR #1086, opened 2026-08-18
fix(payloads): restore data-type scaling for 2411 fan parameters

PWhite-Eng:
### The Problem:
Opcode `2411` (`HvacFanParamPayload`) decoded raw multi-byte integers without applying `data_type` divisor scaling, causing fan parameter values to display with incorrect magnitudes in downstream integrations (e.g. `20000%` instead of `100%`, and `2000 °C` instead of `20.0 °C`, as reported in ramses-rf/ramses_cc#962).

### The Fix:
- Added data-type scaling in `HvacFanParamPayload._scale_value`, `_scale_precision`, `_unscale_value`, and `_unscale_precision` for `0x0F` (fan percentage $\rightarrow /200.0$), `0x92` (temperature $\rightarrow /100.0$), `0x01` (centile $\rightarrow /10.0$), and unscaled integer counters.
- Applied scaling conversions in `HvacFanParamPayload.from_bytes()` and `to_bytes()`.
- Added unit test cases covering all data-type variants and round-trip re-encoding in `tests/tests_rf/test_payload_dataclasses.py`.

### Testing Performed:
- Added unit tests in `test_payload_dataclasses.py` validating scaling for temperature (`75`), fan rate (`41`), and sensitivity (`52`).
- Ran `.venv/bin/prek run -a` and `.venv/bin/ruff check --select D`.
- Verified strict type compliance with `.venv/bin/mypy --strict .`.
- Executed full test suite (`pytest -n auto`): 2,585 passed.

### AI Assistance Disclosure:
This contribution was developed with the assistance of Google Gemini 3.7 Flash for code generation and documentation. All logic and implementations were reviewed and verified by the author.

### DIFF
--- a/src/ramses_rf/payloads/hvac.py
+++ b/src/ramses_rf/payloads/hvac.py
@@ -2077,14 +2077,14 @@ class HvacFanParamPayload(PayloadBase):
     :type parameter_id: int
     :param data_type: Parameter data type integer.
     :type data_type: int
-    :param value_scaled: Current scaled parameter value integer, or None if sentinel/N/A.
-    :type value_scaled: int | None
-    :param min_value_scaled: Minimum allowed scaled parameter value integer.
-    :type min_value_scaled: int
-    :param max_value_scaled: Maximum allowed scaled parameter value integer.
-    :type max_value_scaled: int
-    :param precision_scaled: Parameter scaling precision integer.
-    :type precision_scaled: int
+    :param value_scaled: Scaled parameter value, or None if sentinel/N/A.
+    :type value_scaled: float | int | None
+    :param min_value_scaled: Minimum allowed scaled parameter value.
+    :type min_value_scaled: float | int | None
+    :param max_value_scaled: Maximum allowed scaled parameter value.
+    :type max_value_scaled: float | int | None
+    :param precision_scaled: Parameter scaling precision.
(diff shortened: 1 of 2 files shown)


src/ramses_rf/payloads/hvac.py
```diff
@@ -2093,13 +2093,87 @@ class HvacFanParamPayload(PayloadBase):
...
-    max_value_scaled: int | None
-    precision_scaled: int
+    value_scaled: float | int | None
+    min_value_scaled: float | int | None
+    max_value_scaled: float | int | None
+    precision_scaled: float | int
     trailer_bytes: bytes
     _raw_3b: bytes | None = None
 
+    @staticmethod
```
silverailscolo, line 2103: Looks as something that belongs in `payload/helpers.py`.

silverailscolo (changes requested):

src/ramses_rf/payloads/hvac.py
PWhite-Eng: It is specific to code 2411, I can't forsee any other code needing their hex value 0F having its value divided by 200, or hex value 92 having its value divided by 100.
So I think they are better to be placed with the code 2411 in hvac.py, do you agree?

silverailscolo (approved):

PR #1096, opened 2026-08-20
fix: eavesdropping fails to discover TRV as zone sensor (issue 1010)

wimpie70:
## Summary

Two regressions in the eavesdropping engine prevented it from correctly identifying which TRV is the designated zone sensor in multi-TRV zones. This was reported in https://github.com/ramses-rf/ramses_cc/issues/984#issuecomment-5356693367 — the old `working_schema` correctly identified the representative TRV, but the new discovery logic did not.

### Root causes

**1. `_eavesdrop_from_trv_broadcast` skipped TRVs already bound as actuators**

The refactored code added a check that skipped any TRV with a `_parent` (i.e. already bound as an actuator to a zone). But a TRV can be **both** an actuator and the zone sensor — Evohome lets the user pick which TRV senses temperature for the zone. This check was not present in the pre-refactor code.

**Fix:** Skip only devices that are already a sensor for another zone (check `_is_sensor`, not `_parent`).

**2. `_eavesdrop_from_controller_broadcast` lost the temporal filter**

The refactored code considered every device with any temperature, ignoring when that temperature was last broadcast. The pre-refactor code only considered devices that had broadcast 30C9 **after** the previous controller array — ensuring temporal correlation within the same polling cycle. Without this filter, stale temperatures can produce false matches or miss the correct sensor.

**Fix:** Restore the temporal filter — only consider devices whose most recent 30C9 was broadcast after `prev.dtm`.

### Test evidence

The existing test data (`zone_sensors_003`) already had this bug: zone 05 had a single TRV (`04:230315`) that broadcast 30C9 but was never discovered as the zone sensor. Every other single-TRV zone in the same test correctly had its TRV as both actuator and sensor — zone 05 was the only one missed because its TRV broadcast was processed after it was already bound as an actuator.

The expected schema has been updated to reflect the correct behaviour (zone 05 now has `sensor: 04:230315`).

#### Test plan

- [x] `ruff check` + `ruff format` pass
- [x] All 2602 ramses_rf tests pass (3 skipped)
- [x] All 14 eavesdrop schema tests pass
- [x] All 4 multi-TRV issue 976 tests pass
- [x] `zone_sensors_003` test now correctly expects zone 05 sensor = `04:230315`

### DIFF
--- a/src/ramses_rf/eavesdropper.py
+++ b/src/ramses_rf/eavesdropper.py
@@ -612,10 +612,19 @@ class EavesdropEngine:
                 None,
             ):
                 d_temp = await device.temperature()
-                if d_temp is not None:
-                    if d_temp not in testable_sensors_map:
-                        testable_sensors_map[d_temp] = []
-                    testable_sensors_map[d_temp].append(device)
+                if d_temp is None:
+                    continue
+                # Only consider devices that broadcast 30C9 after the
+                # previous controller array — this ensures temporal
+                # correlation within the same polling cycle (issue 1010).
+                d_msgs = await device.entity_state.get_message_log_flat()
+                if Code._30C9 not in d_msgs:
+                    continue
+                if d_msgs[Code._30C9].dtm <= prev.dtm:
+                    continue
+                if d_temp not in testable_sensors_map:
+                    testable_sensors_map[d_temp] = []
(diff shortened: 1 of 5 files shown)


silverailscolo (approved):
Nice, incremental fix

PR #1110, opened 2026-08-24
fix: suppress TX echo duplicates in CallbackTransport and `Packet._from_cmd` (issue 1041)

wimpie70:
## Summary

- Remove `CallbackTransport.write_frame` override so it inherits `_FullTransport.write_frame`, which calls `_log_tx_packet` with `is_tx=True` and enables echo detection via `_is_recent_tx`. Move the `io_writer` delegation to `_write_frame` (where it belongs in the template method pattern).
- Set `is_echo=True` on `Packet._from_cmd` so the internal helper used by `CommandDTO.tx_header` no longer emits spurious `PKT_LOGGER` entries every time the FSM's `WantEcho` state accesses `tx_header` while waiting for an echo.

These two fixes together eliminate the duplicate `... RQ` entries reported in issue 1041 for MQTT/callback transports. The `Packet._from_cmd` fix applies to all transport types (serial, MQTT, callback, Zigbee).

## Context

- Issue: https://github.com/ramses-rf/ramses_cc/issues/1041
- The existing fix (PR 957) for serial transport did not apply to MQTT because `CallbackTransport` overrode `write_frame` and bypassed `_log_tx_packet`.
- The `Packet._from_cmd` spurious logging was the primary source of `...` RQ duplicates: each incoming packet during the echo wait window triggered a `tx_header` access, which created a new `Packet` via `_from_cmd` and logged it.

## Test plan

- [x] ramses_rf unit tests: 2630 passed, 3 skipped
- [x] ha_sim_test R88 (TX echo suppression): 6/6 PASS
- [x] Full ha_sim_test single-container: 383/395 passed (12 pre-existing failures in R16/R81)
- [x] Full ha_sim_test 3-parallel: 384/388 passed (4 pre-existing failures in R07/R15 under parallel load)
- [x] Verified on `hass` instance (MQTT transport): zero `...` RQ duplicates after restart

### DIFF
--- a/src/ramses_tx/packet.py
+++ b/src/ramses_tx/packet.py
@@ -693,7 +693,13 @@ class Packet:
             f"{command.verb.strip():>2} --- {command.addr1} {command.addr2} {command.addr3} "
             f"{command.code} {int(len(command.payload) / 2):03d} {command.payload}"
         )
-        return cls.from_raw_line(dtm, f"... {raw_line}")
+        # is_echo=True suppresses PKT_LOGGER output in _validate — this
+        # Packet is an internal helper for computing tx_header, not an
+        # actual packet reception, so it must not appear in the packet log.
+        # Without this, every access to CommandDTO.tx_header (which happens
+        # per-incoming-packet in WantEcho.packet_rcvd) logs a spurious
+        # "... RQ" entry (issue 1041).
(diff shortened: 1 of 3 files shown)


silverailscolo (approved):
This doesn't skip the `log_all_mqtt` option switch?

PR #1183, opened 2026-09-01
fix(strategy): normalise 31DA fan_info to canonical mode names

wimpie70:
## Summary

The 31DA parser produces descriptive `fan_info` strings like `"speed 1, low"` / `"speed 2, medium"` / `"speed 3, high"` that don't match the canonical mode names (`"low"` / `"medium"` / `"high"`) used in 22F1 mode maps and HA dropdowns. This caused the `fan_mode` dropdown in ramses_cc to revert to blank after selecting a mode — the real 31DA broadcast overwrote the optimistic value with a string the dropdown couldn't display (ramses_cc issue 1116).

Add `normalize_fan_info()` to `HvacStrategyBase` and the `HvacStrategy` protocol. Call it in `apply_quirk()` for 31DA packets, mapping descriptive names back to canonical names. Unknown strings pass through unchanged.

This is the ramses_rf half of the fix for ramses-rf/ramses_cc#1116. The ramses_cc half (optimistic state update + fan_modes filtering) is in ramses-rf/ramses_cc#1117.

#### Test plan

- [x] 26 unit tests added (15 for `normalize_fan_info` across all vendor strategies, 6 for `apply_quirk` integration, 5 passthrough)
- [x] `ruff check` — passed
- [x] `mypy` — passed (0 issues)
- [x] `pytest` (full suite) — 2922 passed, 8 skipped, 0 failures
- [x] Manual test on hass: select "low" from dropdown, verify it sticks after 31DA broadcast arrives

### DIFF
--- a/src/ramses_rf/strategies/base.py
+++ b/src/ramses_rf/strategies/base.py
@@ -49,6 +49,23 @@ class HvacStrategy(Protocol):
         """
         ...
 
+    def normalize_fan_info(self, fan_info: str) -> str:
+        """Normalise a 31DA fan_info string to a canonical mode name.
+
+        The 31DA parser produces descriptive strings like
+        ``"speed 1, low"`` that don't match the canonical mode names
+        (``"low"``) used in 22F1 mode maps and HA dropdowns.  This
+        method maps descriptive names back to canonical names so
+        consumers can compare fan_info against fan_modes.
+
+        :param fan_info: The descriptive fan_info string from 31DA.
+        :type fan_info: str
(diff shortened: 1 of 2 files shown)


silverailscolo (changes requested):
As I already wrote in the comment, the form the value is parsed should be passed to the front end as-is.
You can normalise there to use it as a selector, but the fan_info value should stay as parsed.
In 2024 I had the same idea, but we rejected it to stick to the manufacturer's names, and that still applies.

PR #1188, opened 2026-09-02
Use indexed binding domains for Orcon CO2 sensors

maleadt:
- [x] AI Coding Assistance disclosure: used for 75%

Follow-up to #1186, based on the fully working version I have now:

- allow binding offers to retain an explicit domain index, including duplicate command codes with different indices
- use the observed `00/31E0`, `01/31E0`, `00/1298` offer for Orcon CO2 sensors
- replace the earlier generic offer fixture with the final live exchange between a faked CO2 identity and a physical HRC-350
- make the virtual-RF binding test retain offer indices so it exercises the full faked-device path

The HRC-350 accepted the indexed offer and completed the accept, confirm, and device-information sequence. The earlier fixture from #1186 was also accepted by the HRC, but it represented the library's generic CO2 offer rather than the
indexed Orcon offer used after further testing.


silverailscolo (approved):
Consulted co-reviewer

PR #1227, opened 2026-09-20
Discovery: classify DIS via VMI-only request evidence, ignore echo packets

wimpie70:
## Summary

A `29:`/`37:` display (DIS) polls the FAN with `RQ` on 'VMI only'
request codes (2411, 31DA, 10D0, 10E0, 1470, 22F7), but the passive
discovery scan fell through to the prefix fallback — `29:` → FAN — so
a display was reported as `schema=REM, discovery=FAN`.  A prefix
fallback then counted each request as *confirming* FAN and reset the
contradiction counter, so the misclassification was stable.

- `_classify`: a source `RQ` on a VMI request code from an ambiguous
  HVAC prefix returns `DIS`, unless the device has already sent
  `W`/`RP` as source (a writer/responder is not a pure display).
  Unambiguous prefixes (`32:`) and VC pairs still win.
- `DiscoveredDevice.verbs_seen`: deduplicated source verbs, persisted
  via `export_json`, gate the DIS signature on verb history rather
  than `codes_seen` alone.
- `_is_evidence_based`: `RQ` on VMI request codes is evidence, so the
  contradiction counter can re-classify a wrongly declared device.
- `PacketDTO.is_echo`: the echo flag (issue 1185) now propagates
  through `Packet.to_dto()` — including post-construction marking by
  the pooled transport — and `DiscoveryScan` ignores echo packets, so
  service traffic sent with a spoofed `from_id` cannot feed passive
  discovery.

NOTE: `313F` was deliberately left out of the VMI set — a plain time
request is too generic (TRVs send `RQ 313F` as well).

## Test plan

- [x] `RQ 2411`/`31DA`/`10D0` from `29:` and `37:` → DIS; `32:` → FAN;
      non-HVAC prefix → unchanged
- [x] Device with `W`/`RP` history is not forced to DIS
- [x] Declared-REM `29:` re-classifies to DIS after RQ threshold; no
      flapping on interleaved `I 22F1`
- [x] Echo packets are ignored entirely (no entry, no counters)
- [x] `verbs_seen` round-trips through export/import; old exports
      without the field import cleanly
- [x] `Packet.to_dto()` propagates `is_echo`, including
      post-construction marking
- [x] Full suite: 3224 passed, 9 skipped; prek/ruff/mypy --strict clean


wimpie70:
I thought we had fixed this, but it's not working correctly yet...checking...

wimpie70:
since it touches some transport files, i'll run a full ha_sim_test on this against regressions

silverailscolo (approved):

PR #1228, opened 2026-09-21
feat(devices): public strategy accessors on DeviceBase

wimpie70:
## Summary
- add public `scheme`/`strategy` properties and `get_strategy(model=None)`/`get_configured_strategy(model=None)` methods on `DeviceBase`, so cross-device strategy resolution no longer reads private members
- `_get_strategy`/`_get_configured_strategy` remain as delegating wrappers for backward compatibility (ramses_cc getattr fallbacks, test mocks)
- migrate internal callers: `HvacCarbonDioxideSensor._resolve_fan_strategy` (`fan._strategy`/`fan._scheme` → `fan.strategy`/`fan.scheme`), the state projector and ingestion quirk paths, and ventilator/remote self-calls
- `EntityState.get_cached_value(code, key)` — synchronous read of the in-memory message state (no sqlite/event loop), backing the new `DeviceBase.model` property (latest 10E0 `description`)
- both accessors default `model` to the cached model, so model-aware scheme selection (e.g. `OrconHrc350Strategy` for VMD-15RMS64) now also applies to sync consumers such as ramses_cc `climate.fan_modes`; `_resolve_fan_strategy` uses the same read path instead of awaiting `get_value`
- `HvacStrategyBase` gains public `aliases`/`boost_aliases` properties, added to the `HvacStrategy` Protocol — the private maps remain for compatibility

Ref: https://github.com/ramses-rf/ramses_cc/issues/1137

#### Test plan
- [x] `pytest` — 3241 passed, 9 skipped
- [x] `ruff check`, `mypy --strict`, `prek run -a`
- [x] new unit tests: public accessor semantics, `model=` → OrconHrc350Strategy, cached-model default, sync `get_cached_value` equivalence, public alias accessors
- [x] ha_sim_test on ha-sim (restarted, new code loaded): R18 R43 R65-68 R79 R90 R91 R93-97 R101 R125 — all pass, no unexpected log errors
- [x] hass dev instance restart — clean startup, no errors


wimpie70:
while working on the diagnostic PR, I found some other private lookups, so considering to extend this PR with public gateway accessors for these in this PR:

Private touchpoints (the fragile bits, both guarded)

Gateway._config() — private; the only source of known_list/block_list/enforce_known_list. Noted in the docstring. Each gateway section is captured independently so a rename → {"error": ...} marker, not a failed download.
Gateway._engine._transport — private chain to the transport object; there is currently no public Gateway.engine/transport accessor (just checked). getattr-guarded.

wimpie70:
I'm moving the commits that belong to issue https://github.com/ramses-rf/ramses_cc/issues/1216 to a separate PR since that's clearer separation, and it needs some other adaptions as well.

silverailscolo (approved):

